PR #15, opened 2020-03-17
Add Magnolia Support

adamgfraser:
Adds Magnolia support for automatically deriving instances for `Equal`. Right now I set it up as a separate project so it is optional. We could potentially use an annotation to do this instead of having to import the derivation.


joroKr21:
Deriving `Equal` would definitely run into #113 so `-Xsource:3.0` would be necessary.

zio-prelude-deriving/src/main/scala/zio/prelude/deriving/DeriveEqual.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import magnolia._
+
+import zio.prelude.Equal
+
+object DeriveEqual {
+
+  type Typeclass[A] = Equal[A]
+
+  def combine[A](caseClass: CaseClass[Equal, A]): Equal[A] =
```
joroKr21, line 11: You can use `ReadOnlyCaseClass`

jdegoes:
Closed, we can re-open later and scale down and unify with Scala 3.0.

PR #164, opened 2020-07-23
Build Out Effect Typeclasses

adamgfraser:
Work in process on building out type classes in `zio.prelude.fx` package. Running into some issues regarding `Stateful`. In particular I think we need `R <:< Has[_]` to be able to implement the `Contextual` combinators since we need to locally modify or eliminate that part of the environment, but that interferes with using `Any` to express the lack of environmental dependencies. Everything else seems to be working well.

### DIFF
--- a/src/main/scala/zio/prelude/fx/Contextual.scala
+++ b/src/main/scala/zio/prelude/fx/Contextual.scala
@@ -1,5 +1,6 @@
 package zio.prelude.fx
 
+import zio.ZIO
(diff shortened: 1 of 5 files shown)


jdegoes:
There's a lot of good work here, but I propose we close for now and scale back the 1.0 release to not include the `fx` type classes.

adamgfraser:
Agreed. I will remove the existing ones and we can add back later when we are ready.

PR #180, opened 2020-08-02
Remove controversies from our readme

mijicd:
I don't think we need to compare to any existing solution in a way that minimizes their impact on the ecosystem. Instead, we should focus on the qualities we bring to it.

That said, I propose a small update to our project's readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,12 +6,9 @@
 
 # Summary
 
-ZIO Prelude is a radically new approach to functional abstractions in Scala, which throws out the 
+ZIO Prelude is an alternative approach to functional abstractions in Scala, which throws out the 
 classic functor hierarchy in favor of a modular algebraic approach that is smaller, easier 
 to understand and teach, and more expressive.
-
-ZIO Prelude is an alternative to libraries like Scalaz and Cats, which imported the Haskell type 
-class hierarchy into Scala without making significant changes.
 
 ZIO Prelude has three key areas of focus:
 


raulraja:
Thank you @mijicd ! 🙌

fommil:
As a former contributor and maintainer to Scalaz, and author of the book, I'm 100% fine with this comparison between ZIO and Scalaz and I do not see it as negative in any way; it's a statement of fact.

Scalaz taking the Haskell typeclass hierarchy (almost) verbatim was **by design** and, if anything, it was a mistake to try and add abstraction (e.g. Invariant, Apply, Divide, Bind, etc... all needless abstractions).

jdegoes:
It is indeed not a negative statement, but a statement of fact that Scalaz, Cats, the Haskell 'standard library', and the Purescript 'standard library' share an almost identical functor hierarchy. But it's also an incomplete statement, because it doesn't mention the ways that, e.g. Scalaz improved over Haskell (e.g. inserting Applicative before Monad, before Haskell did the same; deleting `fail` from Monad) or the ways that, e.g. Cats improved over Scalaz (documentation). Rather than include any necessarily incomplete comparison in the README, it's less disruptive to just introduce the product and third-party sources can do more detailed comparisons for purposes of helping people make adoption decisions.

fommil:
Applicative is still considered an abstration too far in many circles (and I tend to agree on the most part...)

srid:
This is not a bad change. However, I don't think it should be controversial to mention a comparison with other libraries in the README. In fact, it is only very useful to users.

PR #184, opened 2020-08-04
Fix double equality check

mijicd:
cc @alexknvl

### DIFF
--- a/src/main/scala/zio/prelude/Equal.scala
+++ b/src/main/scala/zio/prelude/Equal.scala
@@ -258,10 +258,7 @@ object Equal extends Lawful[Equal] {
    * return `true`, which is different from the behavior of `Double#equals`.
    */
   implicit val DoubleEqual: Equal[Double] =
-    make { (n1, n2) =>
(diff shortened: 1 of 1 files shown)


alexknvl:
Personally, I think this is the correct solution. But I feel like it is important to list the pros / cons:

- The proposed version can distinguish between multiple `NaN`s (~2^53 different ones!).
- It distinguishes between `+0.0` and `-0.0`. This might be problematic because `Ord[Double]` currently doesn't. However, `def f(x: Double) = (1 / x) > 0` has different behavior on `+0.0` and `-0.0`, so they are not really "equal".
- It has the same exact behavior as `==` on every other `Double` value.
- IIRC it was very fast in my experiments. In fact, slightly faster than `==`.

mijicd:
If anything, I think we'll need to accommodate the change in `Ord` rather than dropping this one. To the best of my knowledge, these are the recommended way of comparing doubles (and floats). As far as I know, even `Double.compare` has a flaw as it doesn't really handle NaN properly, but I might be mistaken here.

PR #196, opened 2020-08-09
Added distinct to NonEmptyList

Michaelt293:
Feel free to suggest a more performant implementation :)

### DIFF
--- a/src/main/scala/zio/prelude/NonEmptyList.scala
+++ b/src/main/scala/zio/prelude/NonEmptyList.scala
@@ -55,6 +55,12 @@ sealed trait NonEmptyList[+A] { self =>
     foldLeft(0)((n, a) => if (f(a)) n + 1 else n)
 
   /**
+    * Removes duplicate elements from this `NonEmptyList`.
+    */
+  final def distinct(implicit A: Equal[A]): NonEmptyList[A] =
+    reduceMapLeft(single)((as, a) => if (as.contains(a)) as else cons(a, as)).reverse
+
+  /**
    * Returns whether this `NonEmptyList` and the specified `NonEmptyList` are
    * equal to each other.
    */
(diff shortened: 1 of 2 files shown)


adamgfraser:
@Michaelt293 Thanks!

The problem with this is that it is going to have N squared complexity because for each element in the collection we have to traverse the list of accumulated values. To avoid this we need a way to determine whether an element is in the set of value we have seen before that doesn't require potentially comparing every element, which requires a way of hashing the values. How about something like this?

```scala
final def distinct(implicit A: Hash[A]): NonEmptyList[A] =
  reduceMapLeft(a => (single(a), Set(a.hash))) { case ((as, seen), a) =>
    val hash = a.hash
    if (seen(hash)) (as, seen) else (cons(a, as), seen + hash)
  }._1.reverse
```

Michaelt293:
Thanks @adamgfraser, I was considering passing around a set to keep track of the seen elements. I will go ahead and make the change.

src/main/scala/zio/prelude/NonEmptyList.scala
```diff
@@ -54,6 +54,16 @@ sealed trait NonEmptyList[+A] { self =>
...
     foldLeft(0)((n, a) => if (f(a)) n + 1 else n)
 
+  /**
+   * Removes duplicate elements from this `NonEmptyList`.
+   */
+  final def distinct(implicit A: Hash[A]): NonEmptyList[A] =
+    reduceMapLeft(a => (single(a), Set(a.hash))) {
+      case ((as, seen), a) =>
+        val hash = a.hash
+        if (seen(hash)) (as, seen) else (cons(a, as), seen + hash)
```
joroKr21, line 64: Hmm what if there is a hash collision? 🤔

Michaelt293: If hash collisions are a concern, a `Equal` constraint could be used instead. @adamgfraser let me know if you want me to raise a PR.

PR #199, opened 2020-08-10
Drop/take methods added to NonEmptyList

Michaelt293:


src/main/scala/zio/prelude/NonEmptyList.scala
```diff
@@ -64,6 +64,35 @@ sealed trait NonEmptyList[+A] { self =>
...
+      self match {
+        case Cons(_, t) => t.drop(n - 1)
+        case Single(_)  => Nil
+      }
+
+  /**
+   * Drops the last `n` elements from this `NonEmptyList` returning a `List`.
+   */
+  final def dropRight(n: Int): List[A] =
+    reverse.drop(n).reverse
```
adamgfraser, line 83: I think `take(length - n)` is equivalent and would be slightly more efficient (one traversal for `length` and another for `take` versus two travels for `reverse` and one for `drop`).

src/main/scala/zio/prelude/NonEmptyList.scala
```diff
@@ -309,6 +338,42 @@ sealed trait NonEmptyList[+A] { self =>
   final def tails: NonEmptyList[NonEmptyList[A]] =
     unfold(self)(identity)(_.tailOption)
 
+  @tailrec
+  private def takeHelper[A](n1: Int, nel: NonEmptyList[A], taken: List[A]): List[A] =
```
adamgfraser, line 342: We usually list private methods after all public ones.

src/main/scala/zio/prelude/NonEmptyList.scala
```diff
@@ -481,7 +546,16 @@ object NonEmptyList extends LowPriorityNonEmptyListImplicits {
     }
 
   /**
-   * Constructs a `NonEmptyList` from an `Iterable`.
+   * Constructs a `NonEmptyList` from an `Iterable` or `None` otherwise.
+   */
+  def fromIterable[A](iterable: Iterable[A]): Option[NonEmptyList[A]] =
```
adamgfraser, line 551: Maybe `fromIterableOption` to avoid overloading?

src/main/scala/zio/prelude/NonEmptyList.scala
```diff
@@ -309,6 +338,42 @@ sealed trait NonEmptyList[+A] { self =>
...
+   * Takes the first `n` elements from this `NonEmptyList` returning a `List`.
+   */
+  final def take(n: Int): List[A] =
+    takeHelper(n, self, Nil).reverse
+
+  /**
+   * Takes the last `n` elements from this `NonEmptyList` returning a `List`.
+   */
+  final def takeRight(n: Int): List[A] =
+    takeHelper(n, self.reverse, Nil)
```
adamgfraser, line 360: Could also be `drop(length - n)`. Then we could inline `takeHelper` into `take` like you did with `loop` below.

src/main/scala/zio/prelude/NonEmptyList.scala
Michaelt293: Done.

src/main/scala/zio/prelude/NonEmptyList.scala
Michaelt293: This method has now been deleted.

src/main/scala/zio/prelude/NonEmptyList.scala
Michaelt293: Done.

src/main/scala/zio/prelude/NonEmptyList.scala
Michaelt293: Done.

Michaelt293:
Thanks for your comments. I have made the suggested changes. 

> One question here is how many combinators we want to implement that don't preserve the `NonEmptyList`, since they are already available by an implicit conversion to `List`. But I think having them on `NonEmptyList` itself probably improves discoverability.

Wouldn't we also get better performance since conversion to an intermediate `List` is not required? For example, `take(0)` simply returns an empty list whereas I assume `take(0)` on `List` would first require converting the `NonEmptyList` to a `List` before taking zero elements?

jdegoes:
@adamgfraser Will let you take this one

PR #201, opened 2020-08-11
Fix NonEmptyList#distinct

sideeffffect:
Use Ord and SortedSet instead of Hash, because of possible collisions

fixes https://github.com/zio/zio-prelude/pull/196/files#r467821788

### DIFF
--- a/src/main/scala/zio/prelude/NonEmptyList.scala
+++ b/src/main/scala/zio/prelude/NonEmptyList.scala
@@ -5,10 +5,8 @@ import zio.prelude.NonEmptyList._
 import zio.prelude.newtypes.{ Max, Min, Prod, Sum }
 
 import scala.annotation.tailrec
-import scala.collection.immutable.SortedSet
 import scala.language.implicitConversions
 import scala.util.hashing.MurmurHash3
-import scala.{ math => sm }
 
 /**
  * A `NonEmptyList[A]` is a list of one or more values of type A. Unlike a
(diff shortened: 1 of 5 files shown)


adamgfraser:
I think the issue with this is that it has worse algorithmic complexity than the previous implementation (logarithmic time with a sorted set versus effectively constant time with a hash set). Given this is what is used in Scala's collection library I think it may make sense to continue to implement it in terms of a hash set but open to other opinions.

sideeffffect:
@adamgfraser makes sense, I've made the adjustment

joroKr21:
Yeah unlike `SortedSet` though, `HashSet` cannot be parameterised by `Hash` + `Equal`.

adamgfraser:
@sideeffffect If we are going with using a hash set versus a sorted set then isn't it just the existing implementation?

@joroKr21 That's true, though `HashSet` also cannot be parameterized on a `Hash` instance.

src/main/scala/zio/prelude/NonEmptyList.scala
```diff
@@ -57,11 +57,10 @@ sealed trait NonEmptyList[+A] { self =>
   /**
    * Removes duplicate elements from this `NonEmptyList`.
    */
-  final def distinct(implicit A: Hash[A]): NonEmptyList[A] =
-    reduceMapLeft(a => (single(a), Set(a.hash))) {
+  final def distinct: NonEmptyList[A] =
+    reduceMapLeft(a => (single(a), Set(a))) {
```
sideeffffect, line 61: @adamgfraser look here, there's the crucial bit. The old implementation stores only hashes in the Set, so it can't deal with hash collisions.

jdegoes:
@sideeffffect Nice!

PR #202, opened 2020-08-12
Delete ZPure#eventually

adamgfraser:
Given that the underlying type of `ZPure` is essentially `R => S1 => Either[E, (S2, A)]` I don't think there is a way that `eventually` can produce a different result on subsequent evaluations since it is otherwise pure and is not passing through the updated state on the error case.

Note that this is a little bit of judgment here on the ordering of effects. We could have instead said that the computation always produces an updated state and then also fails with an error or produces an `A` (e.g. `R => S1 => (S2, Either[E, A])`). I thought allowing the computation to fail and not even produce an updated state made sense, but I think you could do it either way.


jdegoes:
This does make me want a way to do `a orElse b` where the output state of the left hand side is fed into the input state of the right hand side.

PR #204, opened 2020-08-12
Float's Equal and Ord consistency

sideeffffect:
Make `Equal[Float]` use floatToIntBits (not floatToRawIntBits) since that is what `Ord[Float]`'s `Float.compare` uses
https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Float.java#L940
(`floatToIntBits` coalesces NaNs into one canonical NaN)

Side note: Why do we even define separate `Equal[Float]` instance anyway? Isn't having just `Ord[Float]` enough, since it's a subtype of `Equal`?

/cc @alexknvl and https://gist.github.com/alexknvl/28dd6a0693c2d92d45528a62f7df9e0e

### DIFF
--- a/src/main/scala/zio/prelude/Equal.scala
+++ b/src/main/scala/zio/prelude/Equal.scala
@@ -258,7 +258,7 @@ object Equal extends Lawful[Equal] {
    * return `true`, which is different from the behavior of `Double#equals`.
    */
   implicit val DoubleEqual: Equal[Double] =
-    make((n1, n2) => java.lang.Double.doubleToLongBits(n1) == java.lang.Double.doubleToLongBits(n2))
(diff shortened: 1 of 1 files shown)


jdegoes:
Thank you and @alexknvl!

sideeffffect:
@jdegoes shouldn't we simply remove the `Equal[Float]` instance, since we have `Ord[Float]` (and analogically for other types)?

adamgfraser:
@sideeffffect We need the instances in `Equal` for implicit resolution since instances in the companion object of a type are automatically in scope. We tried just creating an instance that referred to the instance in the subtype companion object but it seemed to create brittleness with cyclic dependency issues, I think because like `Ord` depends on `Equal` but then `Equal` depends on the `Ord` instance.

PR #205, opened 2020-08-12
Debug improvements

sideeffffect:
Precisely match on tuple in Scala
Render Char properly in Simple
Render Tuples with only `(` `)` in Simple


jdegoes:
Awesome. I was hoping someone would do this!

PR #211, opened 2020-08-13
Add IdentityBoth instances for Either, Try and Future

### LINKED ISSUE #208: Add IdentityBoth instances for Either, Try and Future
We need IdentityBoth instances for Either, Try and Future

jorge-vasquez-2301:
Resolves #208

### DIFF
--- a/src/main/scala/zio/prelude/Derive.scala
+++ b/src/main/scala/zio/prelude/Derive.scala
@@ -2,6 +2,8 @@ package zio.prelude
 
 import zio.{ Cause, Chunk, Exit, NonEmptyChunk }
 
+import scala.util.Try
+
 /**
  * `Derive[F, Typeclass]` represents a universally quantified function from
  * `Typeclass[A]` to `Typeclass[F[A]]` for some `F[_]`. You can think of
@@ -89,6 +91,15 @@ object Derive {
     new DeriveEqual[({ type lambda[x] = Set[A] })#lambda] {
       def derive[B: Equal]: Equal[Set[A]] =
         Equal.SetEqual
+    }
+
+  /**
+   * The `DeriveEqual` instance for `Try`.
+   */
+  implicit val TryDeriveEqual: DeriveEqual[Try] =
+    new DeriveEqual[Try] {
+      def derive[A: Equal]: Equal[Try[A]] =
+        Equal.TryEqual
     }
 
   /**
--- a/src/main/scala/zio/prelude/Equal.scala
+++ b/src/main/scala/zio/prelude/Equal.scala
@@ -6,6 +6,7 @@ import zio.test.laws.{ Lawful, Laws }
 import zio.{ Cause, Chunk, Exit, NonEmptyChunk }
 
 import scala.annotation.implicitNotFound
+import scala.util.Try
 import scala.{ math => sm }
 
 /**
@@ -366,6 +367,16 @@ object Equal extends Lawful[Equal] {
    */
   implicit val StringEqual: Equal[String] =
     default
+
+  /**
+   * Derives an `Equal[Try[A]]` given an `Equal[A]`.
+   */
+  implicit def TryEqual[A: Equal]: Equal[Try[A]] =
+    make {
+      case (scala.util.Success(a1), scala.util.Success(a2)) => a1 === a2
+      case (scala.util.Failure(e1), scala.util.Failure(e2)) => e1 === e2
+      case _                                                => false
+    }
 
   /**
    * Derives an `Equal` for a product type given an `Equal` for each element of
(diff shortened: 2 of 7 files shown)


src/main/scala/zio/prelude/GenFs.scala
```diff
@@ -24,15 +24,6 @@ object GenFs {
         Gen.causes(gen, Gen.throwable)
     }
 
-  /**
```
jorge-vasquez-2301, line 27: I added this on my previous PR, but just realized there's already a `GenF` for `Option` in zio-test

src/main/scala/zio/prelude/GenFs.scala
```diff
@@ -24,15 +24,6 @@ object GenFs {
...
-   * A generator of `Option` values.
-   */
-  val option: GenF[Random with Sized, Option] =
-    new GenF[Random with Sized, Option] {
-      def apply[R1 <: Random with Sized, A](gen: Gen[R1, A]): Gen[R1, Option[A]] =
-        Gen.option(gen)
-    }
-
```
jorge-vasquez-2301, line 35: I added this on my previous PR, but just realized there's already a `GenF` for `Option` in zio-test

src/main/scala/zio/prelude/Derive.scala
```diff
@@ -56,6 +59,15 @@ object Derive {
         Equal.EitherEqual
     }
 
+  /**
+   * The `DeriveEqual` instance for `Future`.
+   */
+  implicit val FutureDeriveEqual: DeriveEqual[Future] =
```
adamgfraser, line 65: I think I would leave out the `Equal` instance for `Future` here. I understand wanting to test the laws for it but defining an `Equal` instance implies there is a sensible way to compare `Future` values for equality. If all `Future` did was asynchronously return a pure value that could make sense. But I think `Future` is normally used as a quasi-effect type where within `Future` you could be doing anything so the fact that two `Future` values return the same result doesn't mean they are the same (maybe one also launched a web query and the other also send some information to a database). For the same reason we haven't defined an `Equal` instance for ZIO and I think the fact that you are having to call `unsafeRun` here reflects the same problem. We have had some discussion of having a concept of `TestEqual` purely for testing but I think we want to keep that separate from actual equality.

src/main/scala/zio/prelude/Equal.scala
```diff
@@ -803,6 +828,15 @@ object Equal extends Lawful[Equal] {
         a1 === a2 && b1 === b2 && c1 === c2 && d1 === d2 && e1 === e2 && f1 === f2 && g1 === g2 && h1 === h2 && i1 === i2 && j1 === j2 && k1 === k2 && l1 === l2 && m1 === m2 && n1 === n2 && o1 === o2 && p1 === p2 && q1 === q2 && r1 === r2 && s1 === s2 && t1 === t2 && u1 === u2 && v1 === v2
     }
 
+  /**
+   * Equality for `Throwable` values.
+   * Comparison is based on: Class, message and cause (stack trace is ignored).
+   */
+  implicit val ThrowableEqual: Equal[Throwable] =
```
adamgfraser, line 835: Not quite sure what the right answer is here. We probably need some instance for derived types that use `Throwable` but the exact definition is not as clear cut as I would like.

src/main/scala/zio/prelude/GenFs.scala
adamgfraser: Yes, I think ideally all of the instances other than the ones for ZIO Prelude data types would like in ZIO Test, but sometimes we need them here temporarily until we can get a ZIO release that has them.

src/main/scala/zio/prelude/Invariant.scala
```diff
@@ -49,6 +52,15 @@ object Invariant {
         Equivalence(_.map(f.to), _.map(f.from))
     }
 
+  implicit val FutureInvariant: Invariant[Future] =
+    new Invariant[Future] {
+      def invmap[A, B](f: A <=> B): Future[A] <=> Future[B] =
+        Equivalence(
+          fa => zio.Runtime.default.unsafeRunToFuture(ZIO.fromFuture(_ => fa).map(f.to)),
```
adamgfraser, line 59: Can't we just `map` here instead of unsafely running? If we need to we can make it an `implicit def` and require an `ExecutionContext` to be in scope.

src/main/scala/zio/prelude/Derive.scala
jorge-vasquez-2301: Yeah, makes sense, I'll just remove this for now

src/main/scala/zio/prelude/Invariant.scala
jorge-vasquez-2301: Yeah, that was my initial approach, but wasn't sure about it

src/main/scala/zio/prelude/Equal.scala
jorge-vasquez-2301: Yeah, this is something I've been discussing with @jdegoes, so John what do you think about this definition?

src/main/scala/zio/prelude/IdentityBoth.scala
```diff
@@ -65,6 +68,43 @@ object IdentityBoth extends LawfulF.Invariant[DeriveEqualIdentityBothInvariant,
...
+            .unwrap(fa)
+            .left
+            .flatMap(a => Failure.unwrap(fb).left.map(b => (a, b)))
+        }
+    }
+
+  /**
+   * The `IdentityBoth` instance for `Future`.
+   */
+  implicit def FutureIdentityBoth: IdentityBoth[Future] =
```
adamgfraser, line 101: Can be a `val` here since we don't need an implicit `ExecutionContext` for `zip`.

src/main/scala/zio/prelude/IdentityBoth.scala
```diff
@@ -99,6 +139,16 @@ object IdentityBoth extends LawfulF.Invariant[DeriveEqualIdentityBothInvariant,
       def both[A, B](fa: => Id[A], fb: => Id[B]): Id[(A, B)] =
         Id(Id.unwrap(fa) -> Id.unwrap(fb))
     }
+
+  /**
+   * The `IdentityBoth` instance for `Try`.
+   */
+  implicit def TryIdentityBoth[R, E]: IdentityBoth[Try] =
```
adamgfraser, line 146: `implicit val` here since the type parameters `R` and `E` are extraneous here.

src/main/scala/zio/prelude/IdentityBoth.scala
jorge-vasquez-2301: Ohhh yeah, right, thanks!

src/main/scala/zio/prelude/IdentityBoth.scala
jorge-vasquez-2301: Yeah, I don't know why I included R and E, thanks!

PR #212, opened 2020-08-13
Coherent Ord and Hash instance for Float (and Double)

sideeffffect:
The instances weren't coherent before. `==` in Hash behaves differently from `java.lang.Float.compare`.

### DIFF
--- a/src/main/scala/zio/prelude/Equal.scala
+++ b/src/main/scala/zio/prelude/Equal.scala
@@ -259,12 +259,23 @@ object Equal extends Lawful[Equal] {
     Derive[F, Equal].derive(Equal[A])
 
   /**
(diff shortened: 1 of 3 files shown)


sideeffffect:
> We need to keep the Hash and Ord instances in their respective companion objects so they are automatically in the implicit scope.

Are you sure? According to [this post](https://meta.plasm.us/posts/2019/09/30/implicit-scope-and-cats/), Scala's implicit scope also includes the companion objects of the super-types. That means that, for example, the instances of `Hash` will be successfully found in `Equals`.
As the post says, this mechanism is already deployed and works.

It is always better to have just one instance, then there could be no incoherency :wink:

adamgfraser:
Wow. Very nice!

PR #221, opened 2020-08-18
equals and hashCode improvements

sideeffffect:


src/main/scala/zio/prelude/Validation.scala
```diff
@@ -34,17 +34,6 @@ sealed trait Validation[+E, +A] { self =>
   final def <&>[E1 >: E, B](that: Validation[E1, B]): Validation[E1, (A, B)] =
     zipPar(that)
 
-  /**
-   * Returns whether this `Validation` and the specified `Validation` are equal
-   * to each other.
-   */
-  override final def equals(that: Any): Boolean =
```
adamgfraser, line 41: We definitely need to keep this because we need to maintain the invariant that equality for `Validation` values is independent of the other of errors.

src/main/scala/zio/prelude/NonEmptyList.scala
```diff
@@ -174,17 +163,6 @@ sealed trait NonEmptyList[+A] { self =>
   final def foreach[F[+_]: AssociativeBoth: Covariant, B](f: A => F[B]): F[NonEmptyList[B]] =
     reduceMapRight(f(_).map(single))((a, fas) => f(a).zipWith(fas)(cons))
 
-  /**
-   * Returns the hashCode of this `NonEmptyList`.
-   */
-  override final def hashCode: Int = {
```
adamgfraser, line 180: Not sure about the performance implications of repeatedly calling the `hashCode` definition for product types here instead of using the one for collection types. Would be inclined to keep these.

src/main/scala/zio/prelude/Equal.scala
```diff
@@ -838,11 +838,17 @@ object Equal extends Lawful[Equal] {
 
   /**
    * Equality for `Throwable` values.
-   * Comparison is based on: Class, message and cause (stack trace is ignored).
+   * Comparison is based on: Class, message, cause and stack trace.
    */
   implicit val ThrowableEqual: Equal[Throwable] =
```
adamgfraser, line 843: We had a previous discussion in #211 about not requiring the same stack trace here. Equality for `Throwable` values is not very well defined but not obvious to me that this is an improvement.

adamgfraser: Right. I think the other side of that is that it can make it very difficult to state any properties regarding values that contain `Throwable`. For example we might like to say something like:

```scala
Future(throw new Exception("boom")).value === Some(Failure(new Exception("boom")))
```

But this isn't necessarily true because these exceptions may contain different stack traces. Essentially the `Throwable` carries along this metadata about the history of the computation in addition to just the value itself. The stack trace is definitely useful for debugging purposes but I think it is questionable whether that should be part of the definition of equality or be regarded as more an implementation detail like the order in which parallel failures occurred (which again could be very useful for debugging but would not consider in the definition of equality).

It is probably less of a problem for ZIO itself because we are polymorphic in our error type but I think for anyone who uses `Throwable` as part of their data types it would be pretty painful.

sideeffffect: I've removed the check of stack traces

src/main/scala/zio/prelude/Validation.scala
adamgfraser: We could potentially do that. At the time we implemented this we didn't have that data structure and it isn't necessary here. We can just use a `NonEmptyChunk` and ignore its ordering. This can be more efficient if we don't need the other properties of `Set` like efficient lookup but just basically need a collection that can be iterated over with an efficient append and concatenation operation.

sideeffffect:
@adamgfraser I've reworked this. Could you please have a look?

sideeffffect:
@adamgfraser can you please have a look at this? :pray: The changes here are small and shouldn't be controversial.

adamgfraser (approved):
Looks good!

PR #222, opened 2020-08-18
Embrace HashSet in ZSet implementation

sideeffffect:
The implementation of `ZSet` stands on `HashSet`, it's better to admit and embrace it than to hide it behind "just" `Map`.

### DIFF
--- a/src/main/scala/zio/prelude/Hash.scala
+++ b/src/main/scala/zio/prelude/Hash.scala
@@ -1,10 +1,11 @@
 package zio.prelude
 
-import scala.annotation.implicitNotFound
-
-import zio.{ Chunk, NonEmptyChunk }
 import zio.test.TestResult
 import zio.test.laws.{ Lawful, Laws }
+import zio.{ Chunk, NonEmptyChunk }
+
+import scala.annotation.implicitNotFound
+import scala.collection.immutable.SortedMap
 
 /**
  * `Hash[A]` provides implicit evidence that a value of type `A` can be hashed.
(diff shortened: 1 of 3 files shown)


src/main/scala/zio/prelude/Hash.scala
```diff
@@ -231,6 +232,17 @@ object Hash extends Lawful[Hash] {
   implicit def SetHash[A]: Hash[Set[A]] =
     default
 
+  /**
+   * Derives a `Hash[SortedMap[A, B]]` given a `Hash[A]` and `Hash[B]`.
+   */
+  implicit def SortedMapHash[A: Hash, B: Hash]: Hash[SortedMap[A, B]] =
```
adamgfraser, line 238: Does relying on the implicit `Hash[A]` make sense? The `SortedMap` is going to be constructed with an arbitrary ordering and Scala's definition of equality, so do we have a guarantee that those are going to be consistent with the definition of equality embodied in the `Hash` instance?

src/test/scala/zio/prelude/ZSetSpec.scala
```diff
@@ -16,7 +16,7 @@ object ZSetSpec extends DefaultRunnableSpec {
     }
 
   def genZSet[R <: Random with Sized, A, B](a: Gen[R, A], b: Gen[R, B]): Gen[R, ZSet[A, B]] =
-    Gen.mapOf(a, b).map(ZSet.fromMap)
+    Gen.mapOf(a, b).map(ZSet.fromMap(_: Map[A, B]))
```
adamgfraser, line 19: The need to use type annotations in unfortunate here. It seems like what this PR buys us is not having to copy the map if it is a `HashMap` versus a normal map, but most of the time that information is not captured in the time signature even if the underlying representation is a `HashMap`. Does it help if we call the methods `fromMap` and `fromHashMap`? Does feel like we are exposing an implementation detail a bit there.

PR #225, opened 2020-08-20
Consolidate Associative Instances

### LINKED ISSUE #219: Consolidate the implementation of Type class instances: Associative
Consolidate the `Associative` type class hierarchy

parent issue: https://github.com/zio/zio-prelude/issues/215

adamgfraser:
Resolves #219.

Consolidates all instances in `Associative` hierarchy in their common super types since instances defined in a companion object of a parent are also in implicit scope.

### DIFF
--- a/src/main/scala/zio/prelude/Associative.scala
+++ b/src/main/scala/zio/prelude/Associative.scala
@@ -10,6 +10,24 @@ import zio.test.laws.{ Lawful, Laws }
  * The `Associative[A]` type class describes an associative binary operator
  * for a type `A`. For example, addition for integers, and string
  * concatenation for strings.
+ *
+ * `Associative` is at the top of the hierarchy for abstracting over operations
+ * to combine types because while there are some operations that are not
+ * associative but do obey other laws, it is generally difficult to combine
+ * more than two values in interesting ways with these operators, and thus to
+ * build solutions to more complicated problems out of solutions to simpler
+ * ones.
+ *
+ * For example, the mean of two numbers is an operation that is commutative but
+ * not associative. However, the lack of associativity is an indication that we
+ * can't combine the means of multiple values in an interesting way with this
+ * definition. If we attempt to take the mean of three values we always place
+ * twice as much weight on one number as the others, which is rarely what we
+ * want.
+ *
+ * If we instead define this operation using a `StatsCounter` object then means
+ * can be combined in ways that are associative, commutative, and have an
+ * identity element, supporting much more interesting modes of composition.
  */
 trait Associative[A] {
   def combine(l: => A, r: => A): A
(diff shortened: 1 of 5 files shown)


src/main/scala/zio/prelude/Associative.scala
```diff
@@ -76,102 +158,228 @@ object Associative extends Lawful[AssociativeEqual] {
...
+      def combine(l: => Prod[Int], r: => Prod[Int]): Prod[Int] = Prod(l * r)
+      val identity: Prod[Int]                                  = Prod(1)
+    }
 
-  implicit val IntSumAssociative: Associative[Sum[Int]] =
-    make[Sum[Int]]((l, r) => Sum(l + r))
+  /**
+   * The `Commutative` and `Identity` instance for the sum of `Int` values.
+   */
+  implicit val IntSumCommutativeIdentity: Commutative[Sum[Int]] with Identity[Sum[Int]] =
```
sideeffffect, line 219: shouldn't this be `Inverse[Sum[Int]]`? (analogously for other `Sum`s of numbers)

src/main/scala/zio/prelude/Associative.scala
```diff
@@ -76,102 +158,228 @@ object Associative extends Lawful[AssociativeEqual] {
   ): Associative[F[A]] =
     derive.derive(associative)
 
-  implicit val DoubleProdAssociative: Associative[Prod[Double]] =
-    make[Prod[Double]]((l, r) => Prod(l * r))
+  /**
+   * The `Commutative` and `Identity` instance for the product of `Double`
+   * values.
+   */
+  implicit val DoubleProdCommutativeIdentity: Commutative[Prod[Double]] with Identity[Prod[Double]] =
```
sideeffffect, line 165: will we be as bold as to declare `Inverse[Prod[Double]]` (and for other `Prod`s for numbers)? :laughing: :pirate_flag:

sideeffffect (changes requested):
consider `Inverse` for (at least some :upside_down_face: ) numbers

src/main/scala/zio/prelude/Associative.scala
```diff
@@ -76,102 +158,228 @@ object Associative extends Lawful[AssociativeEqual] {
...
-        case (Some(l), None)    => Some(l)
-        case (None, Some(r))    => Some(r)
-        case _                  => None
-      }
-    )
+  /**
+   * Derives an `Identity[Option[A]]` given an `Associative[A]`.
+   */
+  implicit def OptionIdentity[A: Associative]: Identity[Option[A]] =
+    new Identity[Option[A]] {
```
sideeffffect, line 294: is there a reason not to use `Identity.make`?

src/main/scala/zio/prelude/Associative.scala
adamgfraser: Great point! Will change.

src/main/scala/zio/prelude/Associative.scala
adamgfraser: I think I just copied this from the previous implementation. It looks like we weren't using `make` there. But we definitely can.

src/main/scala/zio/prelude/Associative.scala
adamgfraser: I think so. We have already said we are going to provide instances for floating point values despite potential rounding errors so I don't see why this would be any different.

sideeffffect: Just for the record, what we did here: the inverse for multiplication is division and that can mean _division by zero_.
the `Inverse[Prod[_]]` instances for `Float` and `Double` are fine, because division by both `+0.0` and `-0.0` are defines (it's one of the `NaN`s, but that is still valid _IEEE 754_ `Float`/`Double`).
Instances for `Byte`, `Short`, `Int` and `Long`, maybe not, there can happen division by `0` and that will throw. But maybe that's fine and pragmatic solution. Or maybe not and we're being to wild :pirate_flag: :smile_cat: 

I just want to make sure we're both on the same page :wink:

adamgfraser: That is a good point. I think floating point errors are okay but division by zero is not. I will revert these.

PR #226, opened 2020-08-23
Function composition helpers

sideeffffect:


src/main/scala/zio/prelude/package.scala
```diff
@@ -112,4 +114,21 @@ package object prelude
...
   }
+
+  implicit class AnySyntax[A](private val a: A) extends AnyVal {
+
+    @silent("side-effecting nullary methods are discouraged")
+    /* Ignores the value, if you explicitly want to do so and avoids "Unused value" compiler warnings. */
+    def ignore: Unit = ()
+
+    /** Applies function `f` to a value `a`, like `f(a)`, but in flipped order and doesn't need parentheses. Can be chained, like `x |> f |> g`. */
+    def |>[B](f: A => B): B = f(a)
```
adamgfraser, line 125: Does this conflict with the new `|>` operator in Scala 2.13?

src/main/scala/zio/prelude/package.scala
```diff
@@ -112,4 +114,21 @@ package object prelude
...
+
+    /** Applies the function `f` to the value `a`, ignores the result, and returns the original value `a`. Practical for debugging, like `x.someMethod.tee(println(_)).someOtherMethod...` . Similar to the `tee` UNIX command. */
+    def tee[B](f: A => B): A = {
```
adamgfraser, line 128: I think this would be called `tap`. Given that this function is only meaningful if `f` is side effecting not sure it is something we want to be adding to every method in a functional programming library. Also could eliminate the `B` type parameter and use `Any` for the result type.

src/main/scala/zio/prelude/AssociativeCompose.scala
```diff
@@ -21,3 +21,16 @@ object AssociativeCompose {
         bc.compose(ab)
     }
 }
+
+trait AssociativeComposeSyntax {
+  implicit class AssociativeComposeOps[:=>[-_, +_], A, B](private val ab: A :=> B) {
+
+    /** Composes `A -> B` with `B -> C` to form `A -> C`. */
+    def >>>[C](bc: B :=> C)(implicit ev: AssociativeCompose[:=>]): A :=> C =
```
adamgfraser, line 29: Great!

src/main/scala/zio/prelude/package.scala
sideeffffect: AFAIK there is no `|>` operator in Scala 2.13

src/main/scala/zio/prelude/AssociativeCompose.scala
sideeffffect: :blush:

src/main/scala/zio/prelude/package.scala
sideeffffect: > `tap`

OK, no problem

> `B` -> `Any`

Good point!

> only meaningful if `f` is side effecting

Yeah, it's a pickle. It seemed like a practical thing, especially for debugging purposes. But you're right, ZIO prelude is supposed to be about pure functional programming, so I definitely wouldn't be insisting on including it.

So what do you say? `tap` goes out?

src/main/scala/zio/prelude/package.scala
adamgfraser: I think these both got added here: https://www.scala-lang.org/api/current/scala/util/ChainingOps.html.

sideeffffect: Ah, I forgot about `tap` being there. Yes, `pipe` got added, but
 * it's only in 2.13
 * needs an explicit import
 * it's not `|>` which is an "operator"
 * `anArgument |> someMethod |> someOtherFunction` is much easier to read than `anArgument.pipe(someMethod).pipe(someOtherFunction)`, especially if the names are more involved

adamgfraser: I hear you. I'm just a little worried we're adding generally nice syntax but not syntax that is specifically related to this library. Like I like both the `|>` and `tap` methods but I'm not sure I would expect them to automatically get added to any type by using ZIO Prelude.

sideeffffect: Noted. I can split it into a new PR and we can move forward with `>>>` . In the meantime, we can think of +/- of `|>` & co.

adamgfraser: Sounds good. You want to just include the `AssociativeComposeSyntax` and we can open an issue for the `AnySyntax`?

sideeffffect: OK, removed
new PR: https://github.com/zio/zio-prelude/pull/227

PR #229, opened 2020-08-26
Implement Constructors And Combinators For These

adamgfraser:
Builds out `These` data type


adamgfraser:
One thing that is a little ambiguous is what the semantics of `orElse` should be. There are two alternative implementations of `orElseEither` that both satisfy the properties of `AssociativeEither`.

The first is:

```scala
  final def orElseEither[A1 >: A, C](that: => These[A1, C]): These[A1, Either[B, C]] =
    self match {
      case Left(_) => that.map(scala.util.Right(_))
      case Right(b) => right(scala.util.Left(b))
      case Both(a, b) => both(a, scala.util.Left(b))
    }
```

This embodies semantics of "recovering" from the case where the left hand side has no success value, so if there is no success on the left hand side we just discard whatever prior error information existed there and replace it with the right hand side. This is consistent with the semantics of `orElse` on other data types like ZIO, but arguably that is not an appropriate comparison because they embody either success or failure rather than both success or failure. It also discards information.

The second is:

```scala
  final def orElseEither[A1 >: A: Associative, C](that: => These[A1, C]): These[A1, Either[B, C]] =
    self match {
      case Left(a) =>
        that match {
          case Left(a1) => left(a <> a1)
          case Right(c) => both(a, scala.util.Right(c))
          case Both(a1, c) => both(a <> a1, scala.util.Right(c))
        }
      case Right(b) => right(scala.util.Left(b))
      case Both(a, b) => both(a, scala.util.Left(b))
    }
```

This embodies adding in the success value from the alternative if there wasn't one before but keeping both error values if the left side and right side both contain errors. This preserves information but arguably goes against the spirit of "recovering" from errors (you recovered from the left side errors so they are no longer relevant).

We could also potentially have variants for both though I feel like that goes a little against the spirit of providing one way to do things.

Copying @calvinlfer.

jdegoes:
I think I'd err on the side of having too many errors, though it strikes me in this case some ADT could help, e.g.:

```scala
sealed trait Cause[+E]
final case class Both[E](left: Cause[E], right: Cause[E])
final case class Irrelevant[E](value: Cause[E]) extends Cause[E]
```

etc., where `left.irrelevant && right == right`, `left && right.irrelevant == left`, for `==` that ignores what is essentially logging metadata--failures that may be interesting but which do not actually contribute to the failure cause (e.g. `orElse`).

adamgfraser:
@jdegoes Yes I was leaning that way as well. Will add that. And also working on something like that, which I think we will want to have for parallel and sequential cases for `Validation` too.

calvinlfer:
That's a great way to do this (introducing a `Cause`), we could also provide derived combinators that could throw away the cause for end-users that don't care but I really love the idea of `Irrelevant` for the purpose of observability.

core/shared/src/main/scala/zio/prelude/Associative.scala
```diff
@@ -1315,30 +1315,30 @@ trait AssociativeSyntax extends PlatformSpecificAssociativeSyntax {
    * Provides infix syntax for combining two values with an associative
    * operation.
    */
-  implicit class AssociativeOps[A](l: A) {
+  implicit class AssociativeOps[+A](l: A) {
```
sideeffffect, line 1318: I thought this isn't necessary, no other Ops classes have this :thinking: Is there an example  where this is needed?

Should we adjust the other Ops classes to also have covariant type parameters?

core/shared/src/main/scala/zio/prelude/These.scala
```diff
@@ -1,77 +1,512 @@
...
+ * both a value and an error. For example, transferring money between two bank
+ * accounts might either succeed, fail completely if there are not sufficient
+ * funds, or succeed with some warnings if the sender's account balance would
+ * be very low after the transfer or the receiver's account has not been
+ * verified. `These` allows modeling these types of computations and preserving
+ * information regarding all errors while still potentially returning a
+ * successful computation.
+ */
 sealed trait These[+A, +B] { self =>
-  def combine[A1 >: A, B1 >: B](that: These[A1, B1])(implicit A1: Associative[A1], B1: Associative[B1]): These[A1, B1] =
```
sideeffffect, line 6: We might consider keeping the `combine` method, it would be useful for defining many type class instances.

core/shared/src/main/scala/zio/prelude/These.scala
```diff
@@ -1,77 +1,512 @@
...
+      case (Both(a, b), Left(a1))     => both(a <> a1, b)
+      case (Both(a, b), Right(b1))    => both(a, b <> b1)
+      case (Both(a, b), Both(a1, b1)) => both(a <> a1, b <> b1)
     }
 
   /**
-   * The `Bicovariant` instance for `These`.
+   * The `AssociativeEither` instance for `These`.
    */
-  implicit val TheseBicovariant: Bicovariant[These] = new Bicovariant[These] {
```
sideeffffect, line 34: We seem to have lost `Bicovariant`

core/shared/src/main/scala/zio/prelude/These.scala
```diff
@@ -1,77 +1,512 @@
...
+  implicit def TheseIdentityBoth[A: Associative]: IdentityBoth[({ type lambda[+b] = These[A, b] })#lambda] =
+    new IdentityBoth[({ type lambda[+b] = These[A, b] })#lambda] {
+      val any: These[Nothing, Any]                                             =
+        These.unit
+      def both[B, C](fa: => These[A, B], fb: => These[A, C]): These[A, (B, C)] =
+        fa.zip(fb)
     }
 
-  /** The `Idempotent` instance for `These`. */
-  implicit def TheseIdempotent[A: Idempotent, B: Idempotent]: Idempotent[These[A, B]] =
```
sideeffffect, line 72: We seem to have lost `Idempotent`

core/shared/src/main/scala/zio/prelude/These.scala
```diff
@@ -1,77 +1,512 @@
...
   final case class Both[+A, +B](left: A, right: B) extends These[A, B]
 
-  /** The `Associative` instance for `These`. */
+  /**
+   * Derives an `Associative[These[A, B]]` given an `Associative[A]` and an `Associative[B]`.
+   */
   implicit def TheseAssociative[A: Associative, B: Associative]: Associative[These[A, B]] =
-    new Associative[These[A, B]] {
-      override def combine(l: => These[A, B], r: => These[A, B]): These[A, B] = l.combine(r)
+    Associative.make {
```
sideeffffect, line 324: We could use `combine` here.

core/shared/src/main/scala/zio/prelude/These.scala
```diff
@@ -1,77 +1,512 @@
...
-  }
 
-  /** The `Commutative` instance for `These`. */
+  /**
+   * Derives a `Commutative[These[A, B]]` given a `Commutative[A]` and a `Commutative[B]`.
+   */
   implicit def TheseCommutative[A: Commutative, B: Commutative]: Commutative[These[A, B]] =
-    new Commutative[These[A, B]] {
-      override def combine(l: => These[A, B], r: => These[A, B]): These[A, B] = l.combine(r)
+    Commutative.make {
```
sideeffffect, line 349: We could use `combine` here.

sideeffffect (commented):
Looks nice, I have only few suggestions.

jdegoes:
@adamgfraser We can close this because of `ZPure`, right?

adamgfraser:
@jdegoes My vision is that with changes to `ZPure` it can encompass the use case for a data type to mode computations that can have non-fatal warnings through the log type parameter. That would allow `These` to be more of an unbiased data type to make it easier to model in a flat way a data type that can be either or both of two values, for example pulling values from two streams. So I think there is still work to do to build out the data type a bit since right now it is quite minimal but it would look substantially different than what is currently here.

core/shared/src/main/scala/zio/prelude/These.scala
```diff
@@ -1,77 +1,512 @@
...
+  final def toValidation: Validation[A, B] =
+    fold(a => Validation.fail(a), b => Validation.succeed(b), (_, b) => Validation.succeed(b))
+
+  /**
+   * Converts this value to a validation success if this value contains a
+   * success or a validation failure otherwise, submerging multiple errors into
+   * the structure of the validation and discarding information about any
+   * errors in the case of success.
+   */
+  final def toValidationNonEmptyChunk[A1](implicit ev: A <:< NonEmptyChunk[A1]): Validation[A1, B] =
```
jdegoes, line 228: Missing implementation.

core/shared/src/main/scala/zio/prelude/These.scala
```diff
@@ -1,77 +1,512 @@
...
+  /**
+   * Transforms the `A` value with the specified function.
+   */
+  final def mapLeft[A2](f: A => A2): These[A2, B] =
+    self match {
+      case Left(a)    => left(f(a))
+      case Right(b)   => right(b)
+      case Both(a, b) => both(f(a), b)
+    }
+
```
jdegoes, line 182: Maybe good to have `merge` that would merge them into the same type.

jdegoes: Assuming `Associative` of course.

jdegoes (dismissed):
Looks good to merge when conflicts resolved!

sideeffffect:
A little `fix` and you should be green :wink:

sideeffffect:
Everything looks green, do you want to merge this @adamgfraser ?

adamgfraser:
@sideeffffect Let's do it!

PR #230, opened 2020-08-26
More and coherent/recycled TupleX instances

sideeffffect:


src/main/scala/zio/prelude/Inverse.scala
```diff
@@ -62,10 +62,801 @@ object Inverse extends Lawful[EqualInverse] {
...
+  /**
+   * Provides infix syntax for combining two values with an inverse
+   * operation.
+   */
+  implicit class InverseOps[A](l: A) {
+
+    /**
+     * A symbolic alias for `inverse`.
+     */
+    def ~~(r: => A)(implicit inverse: Inverse[A]): A =
```
sideeffffect, line 853: I've picked `~~` as the symbolic alias for `inverse`. But feel free to suggest a better name, I was empty on ideas :smile:

adamgfraser: Yeah, it is definitely a decent one. Another idea would be `-` since it isn't overloaded by `anystring2add` but it probably has more risks of being shadowed by another definition and is not really consistent with `<>`. `~<>` is perhaps the most accurate but a little long. I don't really have a better suggestion either which is why I merged as is but always think it is good to be careful with new symbolic operators, especially when it is not an associative operation. 😃

PR #231, opened 2020-08-26
Remove Some Inverse Instances

adamgfraser:
Division as an inverse operation to multiplication is not defined for division by zero.

### DIFF
--- a/src/main/scala/zio/prelude/Associative.scala
+++ b/src/main/scala/zio/prelude/Associative.scala
@@ -89,14 +89,13 @@ object Associative extends Lawful[AssociativeEqual] {
     }
 
   /**
-   * The `Commutative` and `Inverse` instance for the product of `Boolean`
+   * The `Commutative` and `Identity` instance for the product of `Boolean`
    * values.
    */
-  implicit val BooleanProdCommutativeInverse: Commutative[Prod[Boolean]] with Identity[Prod[Boolean]] =
-    new Commutative[Prod[Boolean]] with Inverse[Prod[Boolean]] {
+  implicit val BooleanProdCommutativeIdentity: Commutative[Prod[Boolean]] with Identity[Prod[Boolean]] =
+    new Commutative[Prod[Boolean]] with Identity[Prod[Boolean]] {
       def combine(l: => Prod[Boolean], r: => Prod[Boolean]): Prod[Boolean] = Prod(l && r)
       val identity: Prod[Boolean]                                          = Prod(true)
-      def inverse(l: => Prod[Boolean], r: => Prod[Boolean]): Prod[Boolean] = Prod(l || !r)
     }
 
   /**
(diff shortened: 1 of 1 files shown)


sideeffffect (approved):

src/main/scala/zio/prelude/Associative.scala
```diff
@@ -89,14 +89,13 @@ object Associative extends Lawful[AssociativeEqual] {
...
+   * The `Commutative` and `Identity` instance for the product of `Boolean`
    * values.
    */
-  implicit val BooleanProdCommutativeInverse: Commutative[Prod[Boolean]] with Identity[Prod[Boolean]] =
-    new Commutative[Prod[Boolean]] with Inverse[Prod[Boolean]] {
+  implicit val BooleanProdCommutativeIdentity: Commutative[Prod[Boolean]] with Identity[Prod[Boolean]] =
+    new Commutative[Prod[Boolean]] with Identity[Prod[Boolean]] {
       def combine(l: => Prod[Boolean], r: => Prod[Boolean]): Prod[Boolean] = Prod(l && r)
       val identity: Prod[Boolean]                                          = Prod(true)
-      def inverse(l: => Prod[Boolean], r: => Prod[Boolean]): Prod[Boolean] = Prod(l || !r)
```
sideeffffect, line 99: why are we getting rid of this? this doesn't divide by zero? Or did it break the laws?

sideeffffect (changes requested):
can we keep `Inverse[Prod[Boolean]]`?

src/main/scala/zio/prelude/Associative.scala
```diff
@@ -89,14 +89,13 @@ object Associative extends Lawful[AssociativeEqual] {
     }
 
   /**
-   * The `Commutative` and `Inverse` instance for the product of `Boolean`
+   * The `Commutative` and `Identity` instance for the product of `Boolean`
    * values.
    */
-  implicit val BooleanProdCommutativeInverse: Commutative[Prod[Boolean]] with Identity[Prod[Boolean]] =
```
sideeffffect, line 95: this would then need to be `Inverse[Prod[Boolean]]`

sideeffffect (changes requested):

src/main/scala/zio/prelude/Associative.scala
adamgfraser: I just thought for consistency we should remove the `Inverse` instances for `Prod`. This behavior is already described by boolean conjunction and since we don't have any other instances of `Inverse[Prod[_]]` anymore it doesn't really make sense to try to abstract over that type.

src/main/scala/zio/prelude/Associative.scala
adamgfraser: See comment above.

src/main/scala/zio/prelude/Associative.scala
sideeffffect: I think if it doesn't violate any laws, then it should be perfectly fine to keep it. I'm sure some users will find it useful for something.
(And we're keeping `Inverse[Prod[_]]` for floats anyway, right?)

adamgfraser: No, I got rid of those. I don't think those are lawful because` 0.0 / 0.0 != 0.0`.

sideeffffect (approved):

sideeffffect:
sorry for the constant nudging @adamgfraser, but I think this PR is fine and can be merged now :smile:

adamgfraser:
@sideeffffect No you're great. Thanks for the reminder!

PR #232, opened 2020-08-26
Add NonEmptySet

sideeffffect:
Add `NonEmptySet`.
As the name suggests, it is a set which is guaranteed to contain at least one element. Based on `Set`.
The documentation is still to do, but the code is ready to be reviewed.

### DIFF
--- a/src/main/scala/zio/prelude/NonEmptySet.scala
+++ b/src/main/scala/zio/prelude/NonEmptySet.scala
@@ -70,8 +70,7 @@ final class NonEmptySet[A] private (private val set: Set[A]) { self =>
   def flatten[B](implicit ev: A <:< NonEmptySet[B]): NonEmptySet[B] =
     new NonEmptySet[B](set.foldLeft[Set[B]](Set.empty)((b, a) => b union ev(a)))
 
-  private lazy val hashCodeLazy: Int = set.hashCode ^ getClass.hashCode
-  override def hashCode: Int         = hashCodeLazy
+  override def hashCode: Int = set.hashCode ^ NonEmptySet.NonEmptySetSeed
 
   override def equals(that: Any): Boolean = that match {
     case that: AnyRef if self.eq(that) => true
@@ -168,6 +167,7 @@ object NonEmptySet {
   implicit def toSet[A](nonEmptySet: NonEmptySet[A]): Set[A] =
     nonEmptySet.toSet
 
+  private val NonEmptySetSeed: Int = 1247820194
 }
 
 trait NonEmptySetSyntax {
--- a/src/main/scala/zio/prelude/NonEmptySet.scala
+++ b/src/main/scala/zio/prelude/NonEmptySet.scala
@@ -1,16 +1,82 @@
 package zio.prelude
 
+import zio.NonEmptyChunk
+
+import scala.collection.{ GenSet, GenTraversableOnce }
 import scala.language.implicitConversions
 
-final class NonEmptySet[A] private (val toSet: Set[A]) {
-  def destruct: (A, Set[A])                            = (toSet.head, toSet.tail)
-  def toNonEmptyList: NonEmptyList[A]                  = destruct match { case (head, tail) => NonEmptyList.fromIterable(head, tail) }
-  def add(elem: A): NonEmptySet[A]                     = new NonEmptySet(toSet + elem)
-  def +(elem: A): NonEmptySet[A]                       = new NonEmptySet(toSet + elem)
+final class NonEmptySet[A] private (val toSet: Set[A]) { self =>
+
+  /**
+   * Returns an element of this `NonEmptySet` and the remainder, which is a (possibly empty) `Set`.
+   */
+  def destruct: (A, Set[A]) = (toSet.head, toSet.tail)
+
+  /**
+   * Converts this `NonEmptySet` to a `NonEmptyChunk`.
+   */
+  def toNonEmptyChunk: NonEmptyChunk[A] = destruct match { case (head, tail) => NonEmptyChunk.fromIterable(head, tail) }
+
+  /**
+   * Converts this `NonEmptySet` to a `NonEmptyList`.
+   */
+  def toNonEmptyList: NonEmptyList[A] = destruct match { case (head, tail) => NonEmptyList.fromIterable(head, tail) }
+
+  /** Creates a new `NonEmptySet` with an additional element, unless the element is
+   *  already present.
+   *
+   *  @param elem the element to be added
+   *  @return a new set that contains all elements of this set and that also
+   *          contains `elem`.
+   */
+  def +(elem: A): NonEmptySet[A] = new NonEmptySet(toSet + elem)
+
+  /** Creates a new `NonEmptySet` with additional elements, omitting duplicates.
+   *
+   *  This method takes two or more elements to be added. Elements that already exist in the `NonEmptySet` will
+   *  not be added. Another overloaded variant of this method handles the case where a single element is added.
+   *
+   *  Example:
+   *   {{{
+   *    scala> val a = NonEmptySet(1, 3) + 2 + 3
+   *    a: zio.prelude.NonEmptySet[Int] = NonEmptySet(1, 3, 2)
+   *   }}}
+   *
+   *  @param elem1 the first element to add.
+   *  @param elem2 the second element to add.
+   *  @param elems the remaining elements to add.
+   *  @return   a new `NonEmptySet` with the given elements added, omitting duplicates.
+   */
   def +(elem1: A, elem2: A, elems: A*): NonEmptySet[A] = new NonEmptySet(toSet + elem1 + elem2 ++ elems)
-  def union(that: Set[A]): NonEmptySet[A]              = new NonEmptySet(toSet.union(that))
-  def ++(that: Set[A]): NonEmptySet[A]                 = new NonEmptySet(toSet ++ that)
-  def remove(elem: A): Set[A]                          = toSet - elem
+
+  /** Computes the union between of `NonEmptySet` and another set.
+   *
+   *  @param   that  the set to form the union with.
+   *  @return  a new `NonEmptySet` consisting of all elements that are in this
+   *  set or in the given set `that`.
+   */
+  def union(that: GenSet[A]): NonEmptySet[A] = new NonEmptySet(toSet.union(that))
+
+  /** Creates a new `NonEmptySet` by adding all elements contained in another collection to this `NonEmptySet`, omitting duplicates.
+   *
+   * This method takes a collection of elements and adds all elements, omitting duplicates, into `NonEmptySet`.
+   *
+   * Example:
+   *  {{{
+   *    scala> val a = NonEmptySet(1, 2) ++ NonEmptySet(2, "a")
+   *    a: zio.prelude.NonEmptySet[Any] = NonEmptySet(1, 2, a)
+   *  }}}
+   *
+   *  @param elems     the collection containing the elements to add.
+   *  @return a new `NonEmptySet` with the given elements added, omitting duplicates.
+   */
+  def ++(elems: GenTraversableOnce[A]): NonEmptySet[A] = new NonEmptySet(toSet ++ elems)
+
+  /** Adds the `elem` to this `NonEmptySet`. Alias for `+`. */
+  def add(elem: A): NonEmptySet[A] = self + elem
+
+  /** Removes the `elem` from this `NonEmptySet`. Alias for `-`. */
+  def remove(elem: A): Set[A] = toSet - elem
 
   /**
    * Flattens a `NonEmptySet` of `NonEmptySet` values into a single
@@ -27,8 +93,8 @@ final class NonEmptySet[A] private (val toSet: Set[A]) {
   override def hashCode: Int = toSet.hashCode()
 
   override def equals(that: Any): Boolean = that match {
-    case that: AnyRef if this.eq(that) => true
-    case that: NonEmptySet[A]          => this.toSet == that.toSet
+    case that: AnyRef if self.eq(that) => true
+    case that: NonEmptySet[A]          => self.toSet == that.toSet
     case _                             => false
   }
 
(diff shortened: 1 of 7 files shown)


src/main/scala/zio/prelude/coherent/coherent.scala
```diff
@@ -32,11 +32,11 @@ object AssociativeEitherDeriveEqualInvariant {
     }
 }
 
-trait AssociativeEqual[A] extends Associative[A] with Equal[A]
```
adamgfraser, line 35: The rule I had been trying to maintain here was that we listed the type classes in alphabetical order in the coherent instances.

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -0,0 +1,129 @@
+package zio.prelude
+
+import scala.language.implicitConversions
+
+sealed abstract case class NonEmptySet[A] private (elem: A, others: Set[A]) extends (A => Boolean) {
```
adamgfraser, line 5: Since we are so reliant on the underlying representation of `Set` here I wonder if it makes more sense to model this more like we did `NonEmptyChunk` where you just have `final class NonEmptySet[A] private (private val set: Set[A])` and you just enforce the invariant that all of the constructors only return underlying sets that are not empty. That way we don't have to keep adding this one additional element to the set each time we want to do something.

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * Provides an implicit conversion from `NonEmptySet` to the `Set`
+   * for interoperability with Scala's collection library.
+   */
+  implicit def toSet[A](nonEmptySet: NonEmptySet[A]): Set[A] =
+    nonEmptySet.toSet
+
+}
+
+trait NonEmptySetSyntax {
```
adamgfraser, line 115: Can't these just go on `NonEmptySet` directly since we control the class? May not need the ones that don't preserve the nonempty property since they should already be available through implicit conversion to `Set`.

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -0,0 +1,129 @@
...
+  def fromIterableOption[A](iterable: Iterable[A]): Option[NonEmptySet[A]] =
+    iterable.headOption.fold(None: Option[NonEmptySet[A]])(h => Some(fromIterable(h, iterable.tail)))
+
+  /**
+   * Constructs a `NonEmptySet` with the specified single value.
+   */
+  def single[A](head: A): NonEmptySet[A] =
+    NonEmptySet(head)
+
+  def union[A](l: NonEmptySet[A], r: Set[A]): NonEmptySet[A] =
```
adamgfraser, line 60: I think these methods should all work with `set union nonEmptySet` with the implicit conversion. I think maybe the only thing we need here is implicit syntax on `Set` for when the right hand side is a `NonEmptySet`, because then `union` is guaranteed to return a `NonEmptySet`.

src/main/scala/zio/prelude/coherent/coherent.scala
sideeffffect: :heavy_check_mark:

src/main/scala/zio/prelude/coherent/coherent.scala
```diff
@@ -211,7 +191,27 @@ object DeriveEqualTraversable {
     }
 }
 
-trait EqualIdentity[A] extends AssociativeEqual[A] with Identity[A]
+trait EqualAssociative[A] extends Equal[A] with Associative[A]
```
adamgfraser, line 194: Should be `AssociativeEqual`, right?

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -0,0 +1,207 @@
+package zio.prelude
+
+import zio.NonEmptyChunk
+
+import scala.language.implicitConversions
+
+final class NonEmptySet[A] private (val toSet: Set[A]) { self =>
```
adamgfraser, line 7:
```suggestion
final class NonEmptySet[A] private (private val set: Set[A]) { self =>
```
Would make this `private` so we are hiding the implementation. Maybe just `set` since we aren't converting it to a set, it just is the underlying set?

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -0,0 +1,207 @@
...
+
+  /**
+   * Returns an element of this `NonEmptySet` and the remainder, which is a (possibly empty) `Set`.
+   */
+  def destruct: (A, Set[A]) = (toSet.head, toSet.tail)
```
adamgfraser, line 12: Maybe make this private or inline? Exposing it kind of implies that this is an efficient way to process the elements of the set which I don't think it is.

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -0,0 +1,207 @@
...
+   *    scala> val a = NonEmptySet(1, 3) + 2 + 3
+   *    a: zio.prelude.NonEmptySet[Int] = NonEmptySet(1, 3, 2)
+   *   }}}
+   *
+   *  @param elem1 the first element to add.
+   *  @param elem2 the second element to add.
+   *  @param elems the remaining elements to add.
+   *  @return   a new `NonEmptySet` with the given elements added, omitting duplicates.
+   */
+  def +(elem1: A, elem2: A, elems: A*): NonEmptySet[A] = new NonEmptySet(toSet + elem1 + elem2 ++ elems)
```
adamgfraser, line 49: I think this `+` for adding multiple elements syntax has been removed in the Scala 2.13 collections rework so we probably should drop as well since we already have `+` and `++`.

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -0,0 +1,207 @@
...
+  def flatten[B](implicit ev: A <:< NonEmptySet[B]): NonEmptySet[B] =
+    destruct match {
+      case (head, tail) =>
+        tail.foldLeft(ev(head)) { (b, a) =>
+          b ++ ev(a)
+        }
+    }
+
+  override def hashCode: Int = toSet.hashCode()
+
```
adamgfraser, line 93:
```suggestion
set.foldLeft[Set[B]](Set.empty)((b, a) => b union ev(a))
```
I think slightly more efficient because avoids destructuring the set.

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -0,0 +1,207 @@
...
+  def add(elem: A): NonEmptySet[A] = self + elem
+
+  /** Removes the `elem` from this `NonEmptySet`. Alias for `-`. */
+  def remove(elem: A): Set[A] = toSet - elem
+
+  /**
+   * Flattens a `NonEmptySet` of `NonEmptySet` values into a single
+   * `NonEmptySet`.
+   */
+  def flatten[B](implicit ev: A <:< NonEmptySet[B]): NonEmptySet[B] =
```
adamgfraser, line 84:
```scala
set.foldLeft[Set[B]](Set.empty)((b, a) => b union ev(a))
```
I think slightly more efficient because it avoids destructuring the set.

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -0,0 +1,207 @@
...
+    destruct match {
+      case (head, tail) =>
+        tail.foldLeft(ev(head)) { (b, a) =>
+          b ++ ev(a)
+        }
+    }
+
+  override def hashCode: Int = toSet.hashCode()
+
+  override def equals(that: Any): Boolean = that match {
```
adamgfraser, line 94: I think there is a slight inconsistency between the definitions of `hashCode` and `equals` here because a `NonEmptySet` has the same hash code as its underlying set representation but `equals` is saying that a nonempty set is only equal to another nonempty set.

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -0,0 +1,207 @@
...
+   * Provides an implicit conversion from `NonEmptySet` to the `Set`
+   * for interoperability with Scala's collection library.
+   */
+  implicit def toSet[A](nonEmptySet: NonEmptySet[A]): Set[A] =
+    nonEmptySet.toSet
+
+}
+
+trait NonEmptySetSyntax {
+  implicit class SetOps[A](private val set: Set[A]) {
```
adamgfraser, line 182: Have you checked whether these implicit syntax methods work or get shadowed by the normal methods with the same name?

src/main/scala/zio/prelude/coherent/coherent.scala
sideeffffect: I wanted to unify the naming scheme. So you're saying that you would prefer `XxxEqual`? Should I rename `EqualIdentity` to `IdentityEqual`?

src/main/scala/zio/prelude/NonEmptySet.scala
sideeffffect: I don't think there's any harm in the underlying `Set` being available -- it's immutable after all. `toSet` is just a customary name for a member which gives you the thing converted to a `Set` that's what the NonEmptySet's users will be looking for.

src/main/scala/zio/prelude/coherent/coherent.scala
adamgfraser: At least the way I had tried to unify the naming scheme was it always listed the type classes in alphabetical order. So it would be AssociativeEqual and EqualIdentity.

src/main/scala/zio/prelude/NonEmptySet.scala
sideeffffect: OK, I'll make it inline. The performance will be mainly impacted by `.tail`, but that should be ["effectively constant"](https://docs.scala-lang.org/overviews/collections-2.13/performance-characteristics.html), so it should be acceptable.

src/main/scala/zio/prelude/NonEmptySet.scala
adamgfraser: Okay, definitely agree about being able to access the set, just from a binary compatibility point of view this seems to commit us to the underlying representation also being a set which I think is the right representation but could also be different like in your original implementation.

src/main/scala/zio/prelude/NonEmptySet.scala
sideeffffect: That should be fine, shouldn't it? The law that requires equal instances to return equal hashcodes still holds.

adamgfraser: Yes if values are equal they have the same hash code but a lot of values with the same hash code are not equal which isn't a law violation but is less than ideal. Why don't we add a salt to the hash specific to NonEmptySet so the hash codes of NonEmptySet values are distinct from the hash codes of their underlying representations?

src/main/scala/zio/prelude/NonEmptySet.scala
sideeffffect: they don't :disappointed: 
I've removed them

adamgfraser: If you want we could add a different name for them. Like I think we used concat for Chunk for that reason. Though it is implicit syntax on every Set and I don't know if it is the most important use case.

src/main/scala/zio/prelude/coherent/coherent.scala
sideeffffect: Oh, like that, OK

src/main/scala/zio/prelude/NonEmptySet.scala
sideeffffect: I'd be weary of making the `hashCode` method (or `equals`, for that matter) any more complicated than necessary. The more code there is, the more places for a bug to hide.
I like that now it's quite foolproof.

What would be the practical concern, when a `NonEmptySet` and a `Set` with the same content collide on hashCode? In the vast majority of cases, they won't even be in the same hash-based collection (like Map or Set), because they have different type.

Are you sure we want to do that?

src/main/scala/zio/prelude/NonEmptySet.scala
sideeffffect: I think we can try living without them and then we'll see

src/main/scala/zio/prelude/NonEmptySet.scala
adamgfraser: I think in general wrapper types have different hash codes than the values they wrap. Like `Option(1)` has a different hash code than `1`, `Right(1)`, or `Left(1)`. Most of the time we just get this for free from using case classes because this is how hash code for case classes work but since we aren't using a case class to hide the implementation we have to do it ourselves. I think if we just pick one arbitrary hash, maybe the hash code of "NonEmptySet" and combine it with the hash code of the set itself we should generate different hash codes with minimal work. This is basically what case classes do.

sideeffffect: fair enough

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -0,0 +1,180 @@
...
+  def remove(elem: A): Set[A] = set - elem
+
+  /**
+   * Flattens a `NonEmptySet` of `NonEmptySet` values into a single
+   * `NonEmptySet`.
+   */
+  def flatten[B](implicit ev: A <:< NonEmptySet[B]): NonEmptySet[B] =
+    new NonEmptySet[B](set.foldLeft[Set[B]](Set.empty)((b, a) => b union ev(a)))
+
+  override def hashCode: Int = set.hashCode ^ getClass.hashCode
```
adamgfraser, line 73: Shall we extract the NonEmptySet hash code into a `val` so we don't have to recompute it each time we hash something?

sideeffffect: :heavy_check_mark:

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -0,0 +1,181 @@
...
+  def remove(elem: A): Set[A] = set - elem
+
+  /**
+   * Flattens a `NonEmptySet` of `NonEmptySet` values into a single
+   * `NonEmptySet`.
+   */
+  def flatten[B](implicit ev: A <:< NonEmptySet[B]): NonEmptySet[B] =
+    new NonEmptySet[B](set.foldLeft[Set[B]](Set.empty)((b, a) => b union ev(a)))
+
+  private lazy val hashCodeLazy: Int = set.hashCode ^ getClass.hashCode
```
adamgfraser, line 73: Sorry, didn't mean we had to cache the whole hash code but just the seed for the non-empty set. Something like:

```scala
class NonEmptySet {
  override def hashCode: Int = set.hashCode ^ nonEmptySetSeed
}

object NonEmptySet {
  private val NonEmptySetSeed: Int = 1247820194
```

sideeffffect: But isn't that already better? The whole hash is cached and we don't need to come up with any seed ourselves.

adamgfraser: Well I think using a specific number versus `getClass.hashCode` is just a small optimization so we don't have to do that additional operation. Then a second thing is that portion of the hashCode only needs to be computed once for the entire application. Even if we cache the hash code of each individual entry we end up computing that portion once for each instance instead of only once for the entire application. So those seem like clear wins.

Caching the whole hash code for each individual instance seems like more of a mixed bag since there is also a cost to storing these things and using lazy values. So you definitely see caching hash codes sometimes for data types where it is expensive to hash or is going to be used often but most of the time I don't think you cache the whole hash code for generic data types like this.

PR #233, opened 2020-08-26
Implement ZValidation

### LINKED ISSUE #193: Turn Validation into ZValidation with extra type parameter
We should add a `Log` type parameter (covariant) that appears on successful values (and failures).

Then an operator like `??` (`log`) that allows adding something to the log.

```scala
sealed trait ZValidation[+W, +E, +A] { self =>
  def ?? [W1 >: W](w: W1): ZValidation[W1, E, A] = ???
  def log[W1 >: W](w: W1): ZValidation[W1, E, A] = self ?? w
}
object ZValidation {
  def log(w: W): ZValidation[W, Nothing, Unit] = ???
}

type Validation[+E, +A] = ZValidation[Nothing, E, A]
```

Internally the log can be stored in `Chunk[W]` in both the success and failure cases.

adamgfraser:
Resolves #193.

I think we need some type of `Cause` data type here for the log. Right now I am just using the value or error for equality and hash code. If we want to include the log I think we need some concept of parallel versus sequential errors because for `zipPar` we need combining the log to be commutative but in general I think we want the log to have a concept of ordering. So we could replace the `Chunk[W]` by a `Cause[W]` to represent this.

We also I think are going to want to use `Cause` in `These`. One complication here is that for `Validation` we want the `Cause` to include an `Empty` case because the log could be empty, but for `These` the `Cause` should always contain at least one error.

Thinking a little more broadly, I wonder if we add this writer effect to `ZPure` and then if we wrap its error type in `Cause` I think it could potentially subsume `ZValidation`. At that point I would also wonder if we need `These` for the concept of "non-fatal errors" since these could also be represented by the log. Perhaps at that point `These` becomes a more unbiased data type that is more useful as just an algebraic data type for things like streaming applications and zipping where you want to be able to represent the both case in addition to left and right?


jdegoes:
We can have a `Cause`-like structure (basically free semi-ring) with a type parameter for the "empty", so you can plug in `Nothing` if you don't want an empty.  I think you are right that the natural evolution is pushing writer into `ZPure` and having `ZValidation` emerge as a case of that. Talk about type parameters! 😆 

What do you want to do with this one?

adamgfraser:
Just a few more and we will be up to seven!

I'll work on incorporating into ZPure and once we verify we can do everything with that we can close this.

jdegoes:
It looks like you got everything in the ZPure PR, so will close this one for now. We can re-open later if necessary.

adamgfraser:
Yes I think we can delete the whole data type.

PR #238, opened 2020-09-03
Add ZNonEmptySet

sideeffffect:


adamgfraser (changes requested):
There is a lot of code duplication here. We're basically reimplementing all of the `ZSet` which is going to create a maintenance issue in the future. Why don't we wrap a `ZSet` and just maintain the invariant that it is non-empty like we did for set?

sideeffffect:
good point! I'll make the changes

adamgfraser (approved):

adamgfraser:
Good to merge when conflicts are resolved!

sideeffffect:
Conflicts resolved, build green, ready to be merged :rocket:

adamgfraser (approved):

PR #239, opened 2020-09-03
Add PartialOrd

sideeffffect:
Not everything is totally ordered.
Usable for example for Sets and analogously behaving things.


adamgfraser (commented):
The concept of a partial ordering definitely makes sense.

I think there is a little bit of a conceptual question of how far we want to go in implementing "all the abstractions". I think when we initially implemented `Ord` there was some discussion about also implementing `PartialOrd` and at least at the time the thought was that most things had a total ordering and we would be streamlined and really focus on the key use cases.

That was a while ago and things have evolved so not at all opposed to adding this but may also be good to think a bit more about our philosophy for adding additional abstractions. I think there is some risk with a library like this that you end up with a lot of less used abstractions that impact accessibility and usability, not that we are there yet by any means.

sideeffffect:
For what it's worth, Rust, which also aims to be very pragmatic (yet takes much inspiration from functional programming languages) also has a [`PartialOrd`](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html) Type Class.

sideeffffect:
Dotty fails on
```
[info]       java.lang.AbstractMethodError: Method zio/prelude/Ord$$$Lambda$10282.checkCompare(Ljava/lang/Object;Ljava/lang/Object;)Lzio/prelude/PartialOrdering; is abstract
[info]       	at zio.prelude.Ord$$$Lambda$10282/1407254277.checkCompare(Unknown Source)
[info]       	at zio.prelude.PartialOrd.compare(PartialOrd.scala:21)
```
interesting :thinking:

sideeffffect:
The hierarchy after this change can be seen at the [diagrame here](https://github.com/sideeffffect/zio-prelude/blob/PartialOrd/docs/overview/diagrams.md#equal)

src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -162,10 +163,10 @@ object NonEmptySet {
     chunk => Debug.Repr.VConstructor(List("zio", "prelude"), "NonEmptySet", chunk.toNonEmptyList.map(_.debug).toCons)
 
   /**
-   * Derives an `Equal[NonEmptyList[A]]` given an `Equal[A]`.
+   * Derives an `Hash[NonEmptySet[A]]` and `PartialOrd[NonEmptySet[A]]` (and thus `Equal[NonEmptyList[A]]`) instance.
```
AesaKamar, line 166:
```suggestion
   * Derives a `Hash[NonEmptySet[A]]` and `PartialOrd[NonEmptySet[A]]` (and thus `Equal[NonEmptyList[A]]`) instance.
```

src/main/scala/zio/prelude/NonEmptyList.scala
```diff
@@ -610,4 +610,10 @@ trait LowPriorityNonEmptyListImplicits {
 
     Ord.make((l, r) => loop(l, r))
   }
+
+  /**
+   * Derives an `PartialOrd[NonEmptyList[A]]` given an `PartialOrd[A]`.
```
AesaKamar, line 615:
```suggestion
   * Derives a `PartialOrd[NonEmptyList[A]]` given a `PartialOrd[A]`.
```

src/test/scala/zio/prelude/IdempotentSpec.scala
```diff
@@ -9,6 +11,23 @@ object IdempotentSpec extends DefaultRunnableSpec {
...
+  }
+
+  val anyOrdering: Gen[Random, Ordering] = Gen.anyInt.map { n =>
+    abs(n) % 3 match {
+      case 0 => Ordering.LessThan
+      case 1 => Ordering.Equals
+      case 2 => Ordering.GreaterThan
+    }
+  }
+
```
AesaKamar, line 30: I haven't personally encountered this pattern before for property testing, but I like it.

AesaKamar (commented):
Looks cool

sideeffffect:
Thanks for your review @AesaKamar ! :pray:

core/shared/src/main/scala/zio/prelude/Ord.scala
```diff
@@ -246,17 +192,18 @@ object Ord extends Lawful[Ord] {
...
+  def make[A](ord: (A, A) => Ordering): Ord[A] = new Ord[A] {
+    override protected def checkCompare(l: A, r: A): Ordering = ord(l, r)
+  }
 
   /**
    * Constructs an instance from an `ord` function and a `equal0` function.
    * Since this takes a separate `equal0`, short-circuiting the equality check (failing fast) is possible.
    */
-  def makeFrom[A](ord: (A, A) => Ordering, equal0: Equal[A]): Ord[A] =
+  def makeFrom[A](ord: (A, A) => Ordering, partialOrd: PartialOrd[A]): Ord[A] =
```
jdegoes, line 203: This is weird. What happens if the partial ordering disagrees with the total ordering?

core/shared/src/main/scala/zio/prelude/Ord.scala
```diff
@@ -961,17 +878,76 @@ sealed trait Ordering { self =>
...
+
+object Comparison {
+
+  sealed trait NotEqual extends Comparison
+
+}
+
+object PartialOrdering {
+
+  case object NoOrder extends PartialOrdering with Comparison.NotEqual
```
jdegoes, line 909: `Incomparable`, maybe?

core/shared/src/main/scala/zio/prelude/Ord.scala
```diff
@@ -961,17 +878,76 @@ sealed trait Ordering { self =>
...
+
+sealed trait Comparison extends Product with Serializable
+
+object Comparison {
+
+  sealed trait NotEqual extends Comparison
+
+}
+
+object PartialOrdering {
```
jdegoes, line 907: This "overlapping" ADT is interesting but I wonder what problems it will cause with type inference. In particular, there are some cases where we would like a partial ordering, but due to specific return types, we will get an ordering.

I understand it is convenient to do it like this and I don't want to minimize that. But we should be open to changing our mind one way or the other based on tests.

core/shared/src/main/scala/zio/prelude/Ord.scala
jdegoes: Probably it makes more sense to accept `(A, A) => Option[Ordering]` and convert that to a partial ordering.

core/shared/src/main/scala/zio/prelude/Ord.scala
jdegoes: Another alternative would be to have a separate method for a `PartialOrd`:

```scala
trait PartialOrd[A] {
  def compareOption(l: A, r: R): Option[Ordering]
}
```

Then we wouldn't need a separate ADT and we wouldn't have any overlaps.

@sideeffffect What do you think?

jdegoes (changes requested):
A few comments / suggestions, but overall looking good!

core/shared/src/main/scala/zio/prelude/PartialOrd.scala
```diff
@@ -0,0 +1,955 @@
...
+      }
+
+    makeFrom((l, r) => loop(l, r), Equal.ListEqual)
+  }
+
+  /**
+   * Derives an `PartialOrd[Map[A, B]]` given an `PartialOrd[B]`.
+   * Due to the limitations of Scala's `Map`, this uses object equality on the keys.
+   */
+  implicit def MapPartialOrd[A, B: PartialOrd]: PartialOrd[Map[A, B]] =
```
adamgfraser, line 287: I'm not sure this definition makes sense in general.

It seems to be saying that one map is smaller than another if the keys are a subset and the values are less than or equal. This makes a lot of sense when the values represent some notion of "how many" of each key exist in the map as in `ZSet`. For example:

```scala
val shoppingList1 = Map("Apple" -> 2)
val shoppingList2 = Map("Apple" -> 3, "Orange" -> 1)
shoppingList1 < shoppingList2
```

But consider something like:

```scala
val nickNames1 = Map("William" -> "Will")
val nickNames2 = Map("William" -> "Bill")
```

Here it doesn't really seem like it makes sense to say one of these mappings is "smaller" than the other, they are just different.

I would be inclined to use a definition of subset for maps in general more analogous to sets (tuples of keys and values in one map are a subset of tuples in the other) and use a definition like the current one in `ZSet` where we have a more specific structure regarding what the values represent.

adamgfraser (commented):
I'm worried about the impact on performance here.`PartialOrd` is going to be a lot slower than `Ord` because we have to wrap all the results in an `Option` even if in most cases the ordering is total so we know we don't need to do that. And these operations can get called a lot. In ZIO Cache we saw a big performance improvement when we went from a more complex comparison result type to a simpler one.

One possibility we could explore is using the new typing functionality to help. Alternatively, we could expose variants that allowed working with a partial ordering in cases where we thought that made sense and otherwise require a total ordering.

core/shared/src/main/scala/zio/prelude/PartialOrd.scala
sideeffffect: > Here it doesn't really seem like it makes sense to say one of these mappings is "smaller" than the other, they are just different.

But that's precisely the use case for `PartialOrd`, isn't it? Neither `Map("William" -> "Will")` nor `Map("William" -> "Bill")` is less nor greater than each other, let alone being equal. That's why the comparison for them will return `None` (aka `NoOrder`, or `Incomparable` as suggested above).

I think that's an intuitive and practically useful way for `compareOption` to work.

sideeffffect:
@adamgfraser That's a very good point about performance. Sadly, the overhead of `Option` is a sad reality on Scala on JVM :crying_cat_face:

The more I think of it, the more I find the cost of `Option` too high for our purposes here. I have reverted it back to the `NoOrder` way, now renamed to `Incomparable`. Since you @jdegoes have originally suggested `Option`, please let me know, if you thing the `Incomparable` way is worth the improved resource efficiency.

core/shared/src/main/scala/zio/prelude/PartialOrd.scala
adamgfraser: Right. There is definitely a certain logic embodied in the current approach, I'm just not sure it makes the most sense in general. As that example shows the ordering on the values intuitively doesn't seem to be related to the ordering of the maps. Of course if you take the current logic you can work backwards from it but your initial intuition was that these are things that are just different but not "bigger" or "smaller".

An alternative would be to implement a more limited notion of partial ordering where one map is only "smaller" than another if it contains a subset of the bindings in the other map. So `Map("William" -> "Will")` would be smaller than `Map("William" -> "Will", "Richard" -> "Dick")` but `Map("William" -> "Bill")` would not be comparable to `Map("William" -> "Bill")`. We could still use the current definition for `ZSet` where I think your implementation makes a ton of sense.

sideeffffect: I've change the implementation in Map's PartialOrd :+1: please have another look

sideeffffect:
This should be ready for merging :tada:

sideeffffect:
The build is now green :+1:

jdegoes (approved):

jdegoes:
Thank you! 🙏

PR #240, opened 2020-09-03
Add Idempotent

sideeffffect:
Some operations, like unions of sets don't make sense to perform repeatedly on the same elements


adamgfraser (commented):
Should `Idempotent` really extend `Commutative` and `Identity`? Are we really confident that for at least practical purposes everything that is idempotent is commutative and has an identity element? Idempotency is definitely an interesting concept but does it make the most sense as a binary operator? I think some of the most interesting applications of idempotency are in the effectual context and not clear how we describe them here.

sideeffffect:
Oh, good point! `Idempotent` actually doesn't need `Identity`, only `Commutative`.

The actual name of the mathematical structure that I'm going for here is [Semilattice](https://en.wikipedia.org/wiki/Semilattice). When we mix in `Identity`, like with `IntMaxIdempotentIdentity: Idempotent[Max[Int]] with Identity[Max[Int]]` or `SetIdempotentInverse[A]: Idempotent[Set[A]] with Inverse[Set[A]]`, we get a Bounded Semilattice.
I used the name `Idempotent` to follow the established convention of this library to not be conventional with names of algebraic structures, but I'm open to different name :wink: 

I think (Semi)lattices (their structure and behavior) are common in practice and thus are useful. One example that comes to my mind could be synchronization of shopping carts across different devices/clients and maybe other examples from distributed computing, where (Semi)lattices come up very often AFAIK.

jdegoes:
I like this. @adamgfraser Thoughts?

adamgfraser:
I think this looks good. One question I have is whether it should extend `Commutative`.

It seems like idempotency and commutativity are potentially orthogonal properties. Set union is both idempotent and commutative because sets have no duplicates (idempotency) and are unordered (commutativity). But we could imagine a "mailbox" that filters out any duplicative messages but does maintain an order of messages like a list. Combining two of these mailboxes would be idempotent because we just ignore attempts to add duplicate values but would not in general be commutative because we would be appending messages in one case and prepending them in the other.

This seems like it could be relatively common. For example if we think about idempotent effects such as perhaps writing to a database it would often be the case that writing X and then writing X again is the same as just writing X, but writing X and then writing Y is different from writing Y and then writing X.

sideeffffect:
@adamgfraser Great insight! And we would get Semilattice as `Commutative[A] with Idempotent[A]`. Set could then have `Commutative[Set[A]] with Idempotent[Set[A]] with Inverse[Set[A]]`.
However, mere `Idempotent[A]` can't do `Iterable[A] => Set[A]`, only `Commutative[A] with Idempotent[A]` could do that. But what it could do is `Iterable[A] => Iterable[A]` as in dropping subsequent duplicates.

adamgfraser:
Yes. I think there is also a connection to `Equal` that we haven't fully surfaced yet. Because the guarantee is essentially that doing the "same" thing twice is the same as doing it once, but that also implicitly requires some definition of what it means for  two things to be the "same" which we are kind of implicitly assuming to be default equality or memory equality since in the law we just use the same value twice, but that isn't necessarily the same.

sideeffffect:
Sadly we don't (yet) have rich collections which would take into account ZIO Prelude's `Equal`/`Hash`/`Ord`. But for `Idempotent[A]`, we can definitely do `def optimize(Iterable[A])(implicit A: Equal[A]): Iterable[A]`

adamgfraser:
Yes. I think it is really just `distinct`. `Equal` does the work of saying that these are the distinct elements of a collection. `Idempotent` guarantees that removing the duplicate elements is value preserving with respect to some operation.

sideeffffect:
Reworked

src/main/scala/zio/prelude/Idempotent.scala
```diff
@@ -0,0 +1,498 @@
...
+ * operation results in the same value regardless of the number of values
+ * are combined, allowing us to optimize out unnecessary combinations of the same values.
+ */
+trait Idempotent[A] extends Associative[A] { self =>
+
+  /**
+   * If there are many Equal elements in  sequence, it will remove them all except for the first one.
+   * `a a b c c c a` will result in `a b c a`.
+   */
+  def optimize[T[+_]: Traversable](elems: T[A])(implicit A: Equal[A]): List[A] = {
```
adamgfraser, line 23: I wonder if this method really belongs here. It requires `Equal` and `Traversable` but doesn't actually use the functionality of `Associative` or `Idempotent` at all. I think this might make more since as a method on `Traversable`, `distinct` and then the fact that something is `Idempotent` just tells us that calling distinct on it and then reducing it with an idempotent operation is identical to reducing the original collection, just like we can fold left or fold right with an associative operation.

sideeffffect: Good point. Moved.
I named it `dropRedundant` though, because `distinct` has the connotation that it creates as-if a Set, but that's not the case here.

adamgfraser: Sorry can you say more about what you mean? `distinct` normally returns the same collection type so like called `distinct` on a `List` would return a `List` and maintain the ordering. We can't quite honor the returns the same collection type guarantee because `Traversable` only allows us to rebuild a structure with the shape as the original structure but I think returning a `List` or `Chunk` is the closest thing.

sideeffffect: `List(1,2,2,1).distinct` == `List(1,2)`
`List(1,2,2,1).dropRedundant` == `List(1,2,1)`

sideeffffect:
@adamgfraser @jdegoes Hi, is there something more I can do to move this forward?

src/main/scala/zio/prelude/Idempotent.scala
```diff
@@ -0,0 +1,498 @@
...
+ * are combined, allowing us to optimize out unnecessary combinations of the same values.
+ */
+trait Idempotent[A] extends Associative[A] { self =>
+
+  protected def combineNormal(l: => A, r: => A): A = combine(l, r)
+
+  final def combineIdempotent(l: => A, r: => A)(implicit A: Equal[A]): A =
+    if (l === r) l else combineNormal(l, r)
+
+  def idempotent(implicit A: Equal[A]): Idempotent[A] = new Idempotent[A] {
```
adamgfraser, line 24: Isn't this more a constructor that should be in the companion object rather than a method on the `Idempotent` trait. Like `fromEqual`?

src/main/scala/zio/prelude/Idempotent.scala
```diff
@@ -0,0 +1,498 @@
...
+ * This means that `a <> a` is equal to `a` for all values `a`.
+ * Example of idempotent operations is union of sets, but not addition of integers.
+ *
+ * Idempotent operators are useful because combining  the values with an idempotent
+ * operation results in the same value regardless of the number of values
+ * are combined, allowing us to optimize out unnecessary combinations of the same values.
+ */
+trait Idempotent[A] extends Associative[A] { self =>
+
+  protected def combineNormal(l: => A, r: => A): A = combine(l, r)
```
adamgfraser, line 19: If it is really idempotent do we need `combineNormal` and `combineIdempotent`? Can't we just know that the `combine` operation is idempotent if the type extends `Idempotent`?

src/main/scala/zio/prelude/NonEmptyTraversable.scala
```diff
@@ -85,6 +85,12 @@ trait NonEmptyTraversable[F[+_]] extends Traversable[F] {
     reduceMap(fa)(identity)
   }
 
+  /**
+   * Reduces the non-empty collection of associative elements.
+   */
+  def reduce1[A: Associative](fa: F[A]): A =
```
adamgfraser, line 91: I would use `reduce` for `NonEmptyTraversable` and `reduceOption` for `Traversable`.

src/main/scala/zio/prelude/Traversable.scala
```diff
@@ -189,6 +189,26 @@ trait Traversable[F[+_]] extends Covariant[F] {
...
+   * Reduces the collection to a summary value using the idempotent operation,
+   * returning `None` if the collection is empty.
+   */
+  def reduceIdempotent[A: Idempotent: Equal](fa: F[A]): Option[A] =
+    reduceAssociative(fa)(Idempotent[A].idempotent)
+
+  /**
+   * Reduces the collection to a summary value using the associative operation.
+   */
+  def reduceIdentity[A: Identity](fa: F[A]): A =
```
adamgfraser, line 209: I don't think this is any point in having this operator since an `Identity` is an associative and we're not actually using the power of having an `Identity` as opposed to an `Associative` so you might as well just call `reduceOption`.

src/main/scala/zio/prelude/Traversable.scala
```diff
@@ -189,6 +189,26 @@ trait Traversable[F[+_]] extends Covariant[F] {
...
+   * returning `None` if the collection is empty.
+   */
+  def reduceAssociative[A: Associative](fa: F[A]): Option[A] =
+    foldMap(fa)(a => Option(a))
+
+  /**
+   * Reduces the collection to a summary value using the idempotent operation,
+   * returning `None` if the collection is empty.
+   */
+  def reduceIdempotent[A: Idempotent: Equal](fa: F[A]): Option[A] =
```
adamgfraser, line 203: Shouldn't we actually use the fact that it is idempotent to optimize this?

adamgfraser (changes requested):
@sideeffffect Looks good! Left some specific comments and then I think this will be ready to go.

src/main/scala/zio/prelude/NonEmptyTraversable.scala
sideeffffect: I can't, the name `reduce` is already taken by the method above.

src/main/scala/zio/prelude/Traversable.scala
sideeffffect: > use the fact that it is idempotent

And we do! :) Just have a look at the line below:
```scala
Idempotent[A].idempotent
```

src/main/scala/zio/prelude/Traversable.scala
sideeffffect: > we're not actually using the power of having an `Identity`

But we do. We return `A` and use this on the line below:
```scala
identity[A]
```

src/main/scala/zio/prelude/Idempotent.scala
sideeffffect: `combineNormal` is a detail of the internal implementation that allows `def idempotent(...` to work properly.

We need `combineIdempotent` to have a method that omits unnecessary combining (but to do that, it needs `Equal[A]`).

src/main/scala/zio/prelude/Idempotent.scala
sideeffffect: I took an inspiration from `Commutative`, which has `commute`.

src/main/scala/zio/prelude/Idempotent.scala
adamgfraser: Right. Since it isn't commutative we can't rearrange the terms to further simplify. Goes back to that question of whether it should extend commutative or not. If it was commutative then we could remove all the duplicates.

adamgfraser (approved):
Looks great!

PR #243, opened 2020-09-05
Start using Scalafix

sideeffffect:


adamgfraser (commented):
I think it is great to use Scalafix but would suggest going for a lighter touch here. In ZIO we use it primarily to sort imports with https://github.com/NeQuissimus/sort-imports.

sideeffffect:
OK, I'll get rid of the Disable* rules.
I reached out for [Organize Imports](https://github.com/liancheng/scalafix-organize-imports), because it's a semantic Scalafix rule (unlike sort-imports, which is a syntactic one). That means it's much less likely to produce a broken code.
It's even [recommended by Olafur](https://old.reddit.com/r/scala/comments/g6l83k/organizeimports_a_scalafix_rule_that_helps_you/focswka/)

sideeffffect:
I've disabled all the rules. Only one exception is `noUniversalEquality`, I kept it, because I found it genuinely useful for spotting mistakes. It's very easy to just write `==` even though `===` should have been used. It made me fix many of such places.

src/main/scala-2.11/zio/prelude/TraversableVersionSpecific.scala
```diff
@@ -20,6 +20,7 @@ trait TraversableVersionSpecific {
   object DeriveCanBuildFrom {
     implicit def default[F[+_]](implicit bf: CanBuildFrom[F[Any], Any, F[Any]]): DeriveCanBuildFrom[F] =
       new DeriveCanBuildFrom[F] {
+        @SuppressWarnings(Array("scalafix:DisableSyntax.asInstanceOf"))
```
adamgfraser, line 23: Personally I'm not sure how much value there is in using settings like this within this library.

I can see this being useful when you are trying to introduce more functional paradigms into an organization that doesn't currently follow them and want to just disallow a lot of things that people do by default. But I think we are pretty far away from that.

We already disfavor things like `asInstanceOf` but they can be useful in certain cases for allowing us to express things we know to be true but the compiler can't prove or for efficiency to for example avoid mapping implicit evidence over a deeply nested data structure.

Currently if you use `asInstanceOf` you can do it but it is definitely going to raise an eyebrow in a code review and with this you can suppress the warning and that is going to raise an eyebrow in a code review. I think my suggestion would be to focus these rules on things that are for at least practical purposes "always wrong" (e.g. automatic tupling).

src/main/scala/zio/prelude/CommutativeEither.scala
```diff
@@ -1,14 +1,13 @@
...
-import scala.concurrent.{ ExecutionContext, Future, Promise }
-
 import zio.ZIO
 import zio.prelude.coherent.CommutativeEitherDeriveEqualInvariant
 import zio.stream.{ ZSink, ZStream }
 import zio.test.TestResult
 import zio.test.laws._
 
 import scala.annotation.implicitNotFound
+import scala.concurrent.{ ExecutionContext, Future, Promise }
```
adamgfraser, line 10: Would suggest we follow zio style of `java` first, then `scala`, then `zio` and others.

src/main/scala/zio/prelude/Ord.scala
```diff
@@ -283,9 +283,9 @@ object Ord extends Lawful[Ord] {
 
         @tailrec
         def loop(i: Int): Ordering =
-          if (i == j && i == k) Ordering.Equals
```
adamgfraser, line 286: I don't think this makes sense. We're comparing integer indices in a tight loop. We aren't interested in supporting custom notions of equality we just just want to compare these two indices. This seems to just add unnecessary overhead.

src/main/scala/zio/prelude/Validation.scala
```diff
@@ -183,9 +184,9 @@ object Validation extends LowPriorityValidationImplicits {
    * Derives an `Equal[Validation[E, A]]` given an `Equal[E]` and an
    * `Equal[A]`.
    */
-  implicit def ValidationEqual[E, A: Equal]: Equal[Validation[E, A]] =
+  implicit def ValidationEqual[E: Equal, A: Equal]: Equal[Validation[E, A]] =
     Equal.make {
-      case (Failure(es), Failure(e1s)) => es.groupBy(identity) == e1s.groupBy(identity)
+      case (Failure(es), Failure(e1s)) => es.groupBy(identity) === e1s.groupBy(identity)
```
adamgfraser, line 189: We aren't actually able to use a custom definition of equality for `E` here because we are comparing two `Map` values that use Scala's default notion of equality and hash code for the keys. So let's not pretend otherwise.

src/test/scala/zio/prelude/AssociativeBothSpec.scala
```diff
@@ -5,7 +5,7 @@ import zio.test.laws._
 
 object AssociativeBothSpec extends DefaultRunnableSpec {
 
-  def spec =
+  def spec: ZSpec[Environment, Failure] =
```
adamgfraser, line 8: Understand the principle of including return types for public members but normally we would not include the return type for `spec` here.

src/main/scala-2.11/zio/prelude/TraversableVersionSpecific.scala
sideeffffect: oh, sure! these two were actually leftovers, because the belong to Scala != 2.12, so Scalafix didn't notify me about these unused suppressions and so I haven't removed them :wink:

src/main/scala/zio/prelude/CommutativeEither.scala
sideeffffect: We can definitely do that. But please note, that this order is different from IntelliJ IDEA's default. That means they will fight each other when ordering imports. And IDEA is the most prevalent IDE, so it may be confusing/difficult for some (many?) people.

src/main/scala/zio/prelude/Ord.scala
sideeffffect: yep, performance is worth `==`

src/main/scala/zio/prelude/Validation.scala
sideeffffect: Sure, that's what's currently going to take place, but isn't it better to delegate this behaviour to `Equal[Map[A, B]] where B: Equal` anyway?
The type of `es.groupBy(identity)` is `MapEqual[E, Chunk[E]]`, so the `Equal` instance is used for `Chunk[E]`.
Maybe that's something worth it? It would at least be consistent (in the sense that we're using `===` in the definitions of `Equal` as much as possible).

Let me know, if you would prefer `==` here, no big deal, I'm just trying to figure out what's better by thinking out loud.

src/test/scala/zio/prelude/AssociativeBothSpec.scala
sideeffffect: That's a reasonable request. Sadly, I don't know how to configure Scalafix in such a way, so that it, for example, omits the `ExplicitResultTypes` rule in `Test` configuration :disappointed:

src/main/scala/zio/prelude/CommutativeEither.scala
adamgfraser: That's the order we have used for ZIO and I think it has worked okay. One think that helps was that we added a command alias `prepare` that ran both Scalafmt and Scalfix so we could tell contributors to do this one thing before pushing to make sure that their code worked.

src/test/scala/zio/prelude/AssociativeBothSpec.scala
adamgfraser: Okay.

src/main/scala/zio/prelude/Validation.scala
sideeffffect: :heavy_check_mark:

sideeffffect:
Build is green, I think we're ready for another round of review :smile:

adamgfraser (changes requested):
I would leave out the universal equality and final class rules. Otherwise looks good! Thanks for working on this!

sideeffffect:
fair enough :wink: they're gone

adamgfraser (approved):

sideeffffect:
@adamgfraser could you please restart that one failed job? :pray:

adamgfraser:
@sideeffffect Just did!

sideeffffect:
@adamgfraser great! It's all green now :fireworks:

PR #246, opened 2020-09-08
Add Dotty to the CI

sideeffffect:


adamgfraser:
Nice!

adamgfraser:
Maybe we can copy the latest versions of these files from ZIO since it works on Dotty?

sideeffffect:
Dotty now fails the test
```
[info] - NewtypeSpec
[info]   + NewtypeSmart
[info]     + valid values
[info]     + invalid values
[info]   - examples from documentation
[info]     - meter
[info]       Fiber failed.
[info]       An unchecked error was produced.
[info]       java.lang.ClassCastException: java.lang.Double cannot be cast to zio.prelude.NewtypeExports$Newtype$Tag
[info]       	at zio.prelude.NewtypeSpec$.spec$$anonfun$3(NewtypeSpec.scala:33)
```

adamgfraser:
I can take a look tomorrow. Seems like great progress!

adamgfraser:
I think the issue with the new types in particular will require more work from us. A relatively minimized reproducer of the issue is below:

```scala
sealed trait NewtypeModule {

  def newtype[A]: Newtype[A]

  sealed trait Newtype[A] {
    type Type
    def apply(value: A): Type
  }
}

object NewtypeModule {
  val instance: NewtypeModule =
    new NewtypeModule {
      def newtype[A]: Newtype[A] =
        new Newtype[A] {
          type Type = A
          def apply(value: A): Type = value
        }
    }
}

trait NewtypeExports {
  import NewtypeModule._

  abstract class Newtype[A] extends instance.Newtype[A] {
    val newtype: instance.Newtype[A] = instance.newtype[A]
    trait Tag
    type Type = newtype.Type with Tag
    def apply(value: A): Type = newtype(value).asInstanceOf[Type]
  }
}

object Example extends App with NewtypeExports {
  
  object Meter extends Newtype[Double]
  type Meter = Meter.type
  
  Meter(3.4) // class cast exception
}
```

The problem is that in `NewTypeExports` we define `trait Tag` and then `type Type = new type.Type with Tag`. Then we cast a value of type `A` to a value of type `Type`. `A` is indeed identical to `newtype.Type` so this part of the cast is valid but `Tag` is a completely arbitrary type that we just made up so I don't think it is actually sound to cast `A` to be an instance of this type. It ends up not mattering because `Tag` has no methods defined on it and we never match against `Tag` so we never do anything with it that would be invalid, but I can understand the cast being invalid at this point.

It seems like to some extent we were relying on an idiosyncracy of the Scala 2 compiler so I don't think we really have a basis for going to the Dotty team and saying there is something wrong here, if anything I think they fixed an incorrect behavior we were relying on for our trick here.

We probably need to spend some more time on this. We originally added so we could support defining type class instances for new type. Why don't we mark this test as `scala2Only` until we resolve this?

Copying @jdegoes.

src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -326,14 +326,14 @@ sealed trait ZPure[-S1, +S2, -R, +E, +A] { self =>
...
+              val zPure2 = nested.asInstanceOf[Modify[Any, Any, Any]]
 
-              val updated = zPure2.run(s0)
+              val updated = zPure2.run0(s0)
               s0 = updated._1
               a = updated._2
               curZPure = continuation(a)
 
-            case _ =>
+            case Tags.FlatMap | Tags.Fail | Tags.Fold | Tags.Access | Tags.Provide =>
```
adamgfraser, line 336: Opened a bug report for this with the Dotty team at lampepfl/dotty#9776.

src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -575,26 +575,26 @@ object ZPure {
...
 
-  private final case class Succeed[+A](value: A)                        extends ZPure[Any, Nothing, Any, Nothing, A] {
+  private final case class Succeed[+A](value: A)                     extends ZPure[Any, Nothing, Any, Nothing, A] {
     override def tag: Int = Tags.Succeed
   }
-  private final case class Fail[+E](error: E)                           extends ZPure[Any, Nothing, Any, E, Nothing] {
+  private final case class Fail[+E](error: E)                        extends ZPure[Any, Nothing, Any, E, Nothing] {
     override def tag: Int = Tags.Fail
   }
-  private final case class Modify[-S1, +S2, +E, +A](run: S1 => (S2, A)) extends ZPure[S1, S2, Any, E, A]             {
```
adamgfraser, line 584: Good catch removing the unnecessary `E` type parameter here.

src/main/scala/zio/prelude/Contravariant.scala
```diff
@@ -75,12 +75,16 @@ object Contravariant       extends LawfulF.Contravariant[ContravariantDeriveEqua
    */
   val compositionLaw: LawsF.Contravariant[ContravariantDeriveEqual, Equal] =
     new LawsF.Contravariant.ComposeLaw[ContravariantDeriveEqual, Equal]("compositionLaw") {
-      def apply[F[-_]: ContravariantDeriveEqual, A: Equal, B: Equal, C: Equal](
```
adamgfraser, line 78: Opened an issue for this at lampepfl/dotty#9778.

joroKr21:
It sounds like Dotty has a different erasure for `newtype.Type with Tag` than Scala 2.
In Scala 2 it's just `Object`, but it's reasonable to take the GLB (which is `Tag`).

src/main/scala/zio/prelude/fx/ZPure.scala
sideeffffect: Dotty basically force me to :)

sideeffffect:
The build is green :tada: 
I've created tickets for the uncovered issues, so that we can solve them in other PRs

adamgfraser (dismissed):
Looks good! Let's get this in and then we can follow up on the other items.

adamgfraser (approved):

PR #249, opened 2020-09-08
Compiler fail 2

sideeffffect:


sideeffffect:
fixed by upgrading the versions of Scala and compiler plugins https://github.com/zio/zio-prelude/pull/254

PR #251, opened 2020-09-08
cats 2.2.0

sideeffffect:


adamgfraser (dismissed):
Maybe adding Scala Steward is the more sustainable solution here?

adamgfraser (approved):

PR #253, opened 2020-09-08
Start using Mergify to automate merging Scala Steward's PRs

sideeffffect:
Mergify then has to be enabled here https://doc.mergify.io/getting-started.html

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,24 @@
+pull_request_rules:
+  - name: assign and label scala-steward's PRs
+    conditions:
+      - author=scala-steward
+    actions:
+      assign:
+        users: [adamgfraser]
+      label:
+        add: [dependency-update]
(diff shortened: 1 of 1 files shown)


adamgfraser:
@softinio @mijicd Did you look into this? I thought we had considered doing something to automate the Scala Steward PRs and decided not to for some reason.

mijicd:
@adamgfraser yes, we considered mergify, I even set it up for monix interop back in the days. The reason we decided to drop it was waiting for approval, as it was a must have for us. However, if it is possible to make it play better nowadays, let's give it a try.

adamgfraser:
Should be enabled now. Let's see how this works.

adamgfraser (approved):

adamgfraser:
@sideeffffect Awesome! Thank you! 🙏

PR #266, opened 2020-09-10
Mergify: don't merge major version upgrades

sideeffffect:


adamgfraser:
Good idea. Is there a way to make it work with our setting to squash merge commits? Looks like that is what is preventing the other ones from being merged automatically.

adamgfraser (approved):

PR #279, opened 2020-09-15
add Bicovariant and RightCovariant

lemastero:
* add Bicovariant (in other libraries it is Bifunctor)
```scala
trait Bicovariant[P[+_, +_]] {
  def bimap[AA,A,B,BB](f: A => AA, g: B => BB): P[A,B] => P[AA,BB]
}
```
* because both Divariant and Bicovariant both have the ability to map over second type parameter they have a method that does the same but has a different name in Divariant and Bicovariant (this is in Scalaz, Cats, Haskell)

In this PR I unified this into a single abstraction:
```scala
trait RightCovariant[:=>[_,+_]] {
  def rightMap[A, B, C](f: B => C): (A :=> B) => (A :=> C)
}
```

* renamed `leftMap` => `leftContramap` name is longer but it is consistent with `contramap` from `Contravariant`

* common instances: Either, Tuple2 and some instances for zio: `Exit` 

* ad type alias `Bifunctor`

@jdegoes @adamgfraser WDYT about adding this guy into `zio-prelude`? It is handy when one builds more complex abstractions, moreover, ZIO itself is a bicovariant.


src/main/scala/zio/prelude/Bicovariant.scala
```diff
@@ -0,0 +1,56 @@
+package zio.prelude
+
+trait RightMap[:=>[_,+_]] {
```
sideeffffect, line 3: Wouldn't a better name be `RightCovariant`? To be consistent with the other names...

src/main/scala/zio/prelude/Divariant.scala
```diff
@@ -1,30 +1,24 @@
...
     }
 
   def dimap[A, B, C, D](f: C => A, g: B => D): (A :=> B) => (C :=> D) =
-    (ab: A :=> B) => rightMap(g)(leftMap(f)(ab))
-
-  def leftMap[A, B, C](f: C => A): (A :=> B) => (C :=> B)
+    (ab: A :=> B) => rightMap(g)(leftContramap(f)(ab))
 
-  def rightMap[A, B, C](f: B => C): (A :=> B) => (A :=> C)
+  def leftContramap[A, B, C](f: C => A): (A :=> B) => (C :=> B)
```
sideeffffect, line 13: Nice! `leftContramap` is definitely a better name.

src/main/scala/zio/prelude/Bicovariant.scala
sideeffffect, line 3: Even just the name suggests that there should be a complementary `LeftContravariant`. WDYT?

lemastero: Nice name 👍  thanks.
Renamed: https://github.com/zio/zio-prelude/pull/279/commits/bb2059936e9f205d571a760ab846c020ad230eb9

sideeffffect:
Lint makes the build red. Fix just by running `sbt fix` :wink:

src/main/scala/zio/prelude/Bicovariant.scala
lemastero: Indeed it is possible to have LeftContravariant and LeftCovariant but there are not so many instances for those and only RightCovariant is reusable. OTOH using those 3 abstractions gives a nice split of laws.

I explored this idea and pushed it a lot further in my library, if you want to take a 👀  : https://github.com/lemastero/Triglav/blob/master/src/main/scala/Triglav/face2/Bifunctor.scala.

But lets start with checking if Bifunctor is of interest to zio-prelude :)

sideeffffect: 🃏 :clown_face:
:smile:

adamgfraser:
@lemastero The question for me is just whether we should standardize everything to have three type parameters. These are all logical abstractions but the combinations add up quickly. It would be interesting to see if we can make all of the type classes have three parameters and then make the less polymorphic ones aliases of that.

lemastero:
@adamgfraser WDYT about the following approach:
1) remove instances from `Covariant` for: Either, `Tuple2`
1b) optionally I would remove `Covariant` instances for `Tuple3`, ... as there is no canonical way to say how do you want to map over tuple (only when you have the same type?, over the first slot, over the last slot, over 7th because it is lucky ;) ) In practice, you never need to map over `Tuple3`+
2) add only `Tuple2` and `Either`, `Exit` instances for `Bicovariant`
3) keep only `Function1` instance for `Divariant`
4) add type class Trifunctor and instances for it (as there are a lot of them, that is natural as in general in `ZIO` we abstract over some input, output and error) as described in #291 

Removing `Either` and `Tuple2` instances from `Covariant` sound reasonable for me because when you work with them you use build-in API or you want extra syntax: `bimap`, `mapLeft` from Bicovariant.

My guess similar case would be with Trifunctor.

lemastero:
Oh now I get it! I will see what is possible with aliases.

src/main/scala/zio/prelude/Bicovariant.scala
lemastero: LeftContravariant will be added in later PR as:
`trait TriLeftCovariant[Z[_, +_, _]]`
`trait TriLeftMostCovariant[Z[+_, _, _]]`
ATM there is no reason to add it when we have 2 type params.

BTW RightCovariant will be replaced by [_,_,+_] version.

lemastero:
@sideeffffect I could use re-review / approval. I believe it is ready to be merged.

adamgfraser (approved):
I think we're probably going to have continued movement on this with the ongoing work on the trifunctor approach but let's get this in and we can build on it.

Issue #215, opened 2020-08-17
Consolidate the implementation of Type class instances


sideeffffect:
In the same spirit as https://github.com/zio/zio-prelude/pull/212#issuecomment-673987171 I would like to propose we do that to all other types:
We remove the redundant instances and put the implementation for most concrete Type class to the companion object of the most abstract trait. That way, there's just one, and thus always coherent, implementation for a given Type class hierarchy.
I volunteer to do the work!

What do you guys think?

It is probably a good idea not to do it all at once, but in batches/groups. Any idea what would be the most natural grouping for this endeavor?

---

- [x] https://github.com/zio/zio-prelude/issues/218 `Equal`
- [x] https://github.com/zio/zio-prelude/issues/219 `Associative`
- [x] https://github.com/zio/zio-prelude/issues/220 `AssociativeBoth`
- [x] https://github.com/zio/zio-prelude/issues/282 `AssociativeEither`
- [x] https://github.com/zio/zio-prelude/issues/283 `Invariant`
- [x] https://github.com/zio/zio-prelude/issues/284 `AssociativeFlatten`
- [x] https://github.com/zio/zio-prelude/issues/285 `AssociativeCompose`

adamgfraser:
@sideeffffect Good idea! We can probably break it up into the `Equal` hierarchy, the `Associative` hierarchy, and the `AssociativeBoth` / `AssociativeEither` hierarchies. I am also happy to do some of it.

sideeffffect:
:fireworks: 
awesome work everybody!
:muscle: :muscle: :muscle:

PR #280, opened 2020-09-15
WIP unify AssociativeBoth and AssociativeEither instances

### LINKED ISSUE #220: Consolidate the implementation of Type class instances: AssociativeBoth
Consolidate the `AssociativeBoth` type class hierarchy

parent issue: https://github.com/zio/zio-prelude/issues/215

lemastero:
Resolve #220

* combine instances present in `AssociativeBoth` and `AssociativeEither` and place them in `AssociativeEither` (in AssociativeBoth there is lots of code for Tuple2 ... Tuple22

* add type alias:
```scala
trait AssociativeEitherBoth[F[_]] extends AssociativeEither[F] with AssociativeBoth[F]
```
so I can express this:
```scala
implicit def EitherAssociativeBothEither[L]:
    AssociativeEither[({ type lambda[+r] = Either[L, r] })#lambda]
    with AssociativeBoth[({ type lambda[+r] = Either[L, r] })#lambda] =
    new AssociativeEither[({ type lambda[+r] = Either[L, r] })#lambda]
     with AssociativeBoth[({ type lambda[+r] = Either[L, r] })#lambda] {
```
as
```scala
implicit def EitherAssociativeBothEither[L]: AssociativeEitherBoth[({ type lambda[+r] = Either[L, r] })#lambda] =
    new AssociativeEitherBoth[({ type lambda[+r] = Either[L, r] })#lambda]
```

The description was short, if I misunderstood sth or sth more needs to be done please let me know :)

### DIFF
--- a/src/main/scala/zio/prelude/AssociativeBoth.scala
+++ b/src/main/scala/zio/prelude/AssociativeBoth.scala
@@ -1,7 +1,6 @@
 package zio.prelude
 
 import scala.annotation.implicitNotFound
-import scala.concurrent.Future
 import scala.util.Try
 
 import zio._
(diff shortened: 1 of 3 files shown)


sideeffffect:
I think you have misunderstood the ticket. But that's my fault, I should have described it better.
Let me now at least explain the point of https://github.com/zio/zio-prelude/issues/215 with an example:

There is unnecessary code duplication when we define instances for various Type classes that are in a hierarchy.
Remember that `IdentityBoth` <: `AssociativeBoth`. When you look at how the instances are defined for these, you'll see that there's
 * `AssociativeBoth.OptionAssociativeBoth: AssociativeBoth[Option]` and there's
 * `IdentityBoth.OptionIdentityBoth: IdentityBoth[Option]`

But that is redundant, since `OptionIdentityBoth` already subsumes all functionality that is in `OptionAssociativeBoth`. There should be only one, so that there's no space for incoherence and less space for bugs.
`OptionIdentityBoth` should move to `AssociativeBoth`, so that there would be only `AssociativeBoth.OptionIdentityBoth: IdentityBoth[Option]`.

Analogously, `CommutativeEither` <: `AssociativeEither` and there's both
 * `AssociativeEither.ZIOAssociativeEither[R, E]: AssociativeEither[ZIO[R, E, +_]]` and
 * `CommutativeEither.ZIOCommutativeEither[R, E]: CommutativeEither[ZIO[R, E, +_]]`

when having `ZIOCommutativeEither` is enough -- but it should live in `AssociativeEither`.

The reason is, that Scala implicit resolution mechanism is searching in companion objects of super types too.
See this blog post for a more detailed explanation: https://meta.plasm.us/posts/2019/09/30/implicit-scope-and-cats/

Do you think we could do this PR only the `AssociativeBoth` hierarchy? It is a mechanical work, but is not a small work. Then we could do `AssociativeEither` in another PR.

lemastero:
Thanks for your explanation @sideeffffect. Now it makes a lot more sense :) On my way to introduce changes as you described.

lemastero:
Replaced by https://github.com/zio/zio-prelude/pull/288

Issue #285, opened 2020-09-15
 Consolidate the implementation of Type class instances: AssociativeCompose


sideeffffect:
Consolidate the `AssociativeCompose` type class hierarchy

parent issue: #215

PR #286, opened 2020-09-15
Consolidate the implementation of Type class instances: AssociativeCompose

sideeffffect:
closes https://github.com/zio/zio-prelude/issues/285

/cc @lemastero

### DIFF
--- a/src/main/scala/zio/prelude/AssociativeCompose.scala
+++ b/src/main/scala/zio/prelude/AssociativeCompose.scala
@@ -14,9 +14,12 @@ trait AssociativeCompose[:=>[-_, +_]] {
     eq.equal(ad1, ad2)
   }
 }
-object AssociativeCompose             {
-  implicit val FunctionAssociativeCompose: AssociativeCompose[Function] =
-    new AssociativeCompose[Function] {
+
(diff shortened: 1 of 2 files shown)


src/main/scala/zio/prelude/AssociativeCompose.scala
```diff
@@ -14,9 +14,12 @@ trait AssociativeCompose[:=>[-_, +_]] {
...
 }
-object AssociativeCompose             {
-  implicit val FunctionAssociativeCompose: AssociativeCompose[Function] =
-    new AssociativeCompose[Function] {
+
+object AssociativeCompose {
+  implicit val FunctionIdentityCompose: IdentityCompose[Function] =
+    new IdentityCompose[Function] {
+      def identity[A]: A => A = (a: A) => a
+
```
lemastero, line 22: WDYT about removing instance in here and restore the one in IdentityCompose ?

lemastero: I have study others PR related to this. Now I understand why that way and not the other :)

lemastero (approved):

adamgfraser (approved):

PR #292, opened 2020-09-17
remove unused type param in ZQueueContravariant

lemastero:
`ZQueueContravariant` does not need type param `A`


sideeffffect (approved):
Nice

adamgfraser (approved):
Good catch!

Issue #273, opened 2020-09-12
Rethink Equality for contravariant types (not only) because of Dotty


sideeffffect:
@adamgfraser https://github.com/zio/zio-prelude/pull/246#discussion_r487353697 https://github.com/lampepfl/dotty/issues/9778#issuecomment-691390362

> I think that is right. Equality for contravariant types is tricky in general. If the input type `A` can be enumerated then you can say that two `F[A]` values are equal if they produce the same outputs for all inputs. If they can't be enumerated you could at least try to test a subset of the values. But either way you need a notion of equality for whatever you are converting the `A` values into, not the `A` values. So I think this is a case where we too quickly took logic from the covariant case to the contravariant one. Thanks for the good thought!

adamgfraser:
I will work on this.

PR #298, opened 2020-09-19
Support Equality For Contravariant Types

### LINKED ISSUE #92: How Should We Support "Test" Equal Instances?
We're starting to add instances for more data types that don't have well defined `Equal` instances. For example, `Ord` isn't a data type so we can't just look at two `Ord` values and say whether they are equal. At the same time, we would like to be able to test laws for them to make sure our instances are well behaved.

One approach to do this would be to allow defining some "test" `Equal` instances. For example, given two `Ord[A]` values we could say they are equal "as far as we can tell" if they return the same ordering for a random sample of values. We could not know for certain the values are actually equal unless the domain was very small, but this could still probably provide a high degree of confidence in the correctness of instances.

This would probably require performing effects in determining the "to our best knowledge" equality, especially for effect types but would require defining `Equal` instances that are not "true equality" in some sense, at least within our own internal test suites.

Does this make sense or is there another way we should approach this?

adamgfraser:
Work in process to resolve #92 and #273.

I added a new `Enumerable` type class to describe types for which all values can be enumerated.  With this we can assert that types such as functions are equal if two functions produce equal outputs for every possible input.

With this we should be able to test laws for contravariant types, but I think we still have to do some more on the implementation of laws in ZIO Test for contravariant types.

In particular, laws are currently based on the concept of a `GenF`, which says given a generator of `A` values we can generate `F[A]` values. This makes a lot of sense for covariant types like `Option` where there are lots of different values for any `Option[A]` (e.g. an `Option[Int]` could be `None`, `Some(1)`, `Some(2)`, etc...). But for contravariant types there is generally only a single value for a given type (e.g. a single `Equal[Int]`) that we are testing and we can't derive it from having a way to generate `Int` values. So I think we need to update the laws framework for contravariant types for this. I will work on that.


lemastero (approved):
LGTM

jdegoes (approved):

jdegoes:
@adamgfraser I think also we should move all the laws into ZIO Prelude. They're basically ZIO Prelude laws, anyway.

src/main/scala/zio/prelude/Enumerable.scala
```diff
@@ -0,0 +1,17 @@
+package zio.prelude
+
+trait Enumerable[+A] {
```
jdegoes, line 3: This one could have `ordinal` and `fromOrdinal`, together with `size`, which would allow a law that ties these together to `enumerate`, as well as helpers like `min` and `max`.

src/main/scala/zio/prelude/Equal.scala
```diff
@@ -300,6 +307,18 @@ object Equal extends Lawful[Equal] {
   implicit def EitherEqual[A: Equal, B: Equal]: Equal[Either[A, B]] =
     Equal[A] either Equal[B]
 
+  /**
+   * Derives an `Equal[Equal[A]]` given an `Enumerable[A]`.
+   */
+  implicit def EqualEqual[A: Enumerable]: Equal[Equal[A]] =
```
jdegoes, line 313: 😆

src/main/scala/zio/prelude/Equal.scala
```diff
@@ -300,6 +307,18 @@ object Equal extends Lawful[Equal] {
...
+    Equal.make { (l, r) =>
+      val pairs = for {
+        a1 <- Enumerable[A].enumerate
+        a2 <- Enumerable[A].enumerate
+      } yield l.equal(a1, a2) == r.equal(a1, a2)
```
jdegoes, line 318: I think we want `a1.zip.a2(_ === _).forall(identity)`.

jdegoes: Or maybe I misunderstand...

src/main/scala/zio/prelude/Enumerable.scala
adamgfraser: I like that.

src/main/scala/zio/prelude/Equal.scala
adamgfraser: I think ideally we want the Cartesian product here because we want to say that for all combinations of values the two equal instances give the same answer.

sideeffffect:
This may be worth resurrecting :thinking:  /cc @jdegoes

Issue #299, opened 2020-09-19
Create Validation from predicate


vinhhv:
Ability to create a Validation from a predicate, similar to `Either`'s `.cond`:

```
def cond[A, B](test: Boolean, right: => B, left: => A): Either[A, B]
```

Suggestion:
```
def fromPredicate[E, A](predicate: Boolean, value: => A, error: => E): Validation[E, A]
```

PR #300, opened 2020-09-19
Add fromPredicate

vinhhv:
https://github.com/zio/zio-prelude/issues/299

### DIFF
--- a/src/main/scala/zio/prelude/Validation.scala
+++ b/src/main/scala/zio/prelude/Validation.scala
@@ -293,7 +293,16 @@ object Validation extends LowPriorityValidationImplicits {
   def fromOption[A](value: Option[A]): Validation[Unit, A] =
     value.fold[Validation[Unit, A]](fail(()))(succeed)
 
-  def fromPredicate[E, A](error: E, value: A)(f: A => Boolean): Validation[E, A] =
+  /**
+   * Constructs a `Validation` from a predicate, failing with None.
+   */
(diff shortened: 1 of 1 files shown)


src/main/scala/zio/prelude/Validation.scala
```diff
@@ -293,6 +293,10 @@ object Validation extends LowPriorityValidationImplicits {
   def fromOption[A](value: Option[A]): Validation[Unit, A] =
     value.fold[Validation[Unit, A]](fail(()))(succeed)
 
+  def fromPredicate[E, A](predicate: Boolean, value: => A, error: => E): Validation[E, A] =
```
jdegoes, line 296: How about this signature:

```scala
def fromPredicate[E, A](error: E, value: A)(f: A => Boolean): Validation[E, A] = ???
```

Or possibly even better:

```scala
def fromPredicate[E, A](value: A)(f: A => Boolean): Validation[None.type, A] = ???
```

vinhhv: I think the first one makes sense; allows the user to control the error returned.

adamgfraser: Maybe we need two versions here, `fromPredicate` that just provides a default failure value and then `fromPredicateWith` that lets the user specify an error to fail with if the predicate is not satisfied.

vinhhv: Good idea, I'll add that.

adamgfraser (approved):
Thanks for adding!

PR #303, opened 2020-09-23
make SubtypeSmart[A]#Type constrained as subtype of A

gurghet:
Continuing from https://discordapp.com/channels/629491597070827530/738827844049961010/758265964738904104
this is more like a note for @adamgfraser to check this


sideeffffect (approved):
Nice catch :+1:

adamgfraser (approved):
Great fix!

PR #305, opened 2020-09-23
Newtype Tag extends Any

### LINKED ISSUE #272: Rethink Newtype.Tag (not only) because of Dotty
@adamgfraser https://github.com/zio/zio-prelude/pull/246#issuecomment-691368551

> I think the issue with the new types in particular will require more work from us. A relatively minimized reproducer of the issue is below:
> 
> ```scala
> sealed trait NewtypeModule {
> 
>   def newtype[A]: Newtype[A]
> 
>   sealed trait Newtype[A] {
>     type Type
>     def apply(value: A): Type
>   }
> }
> 
> object NewtypeModule {
>   val instance: NewtypeModule =
>     new NewtypeModule {
>       def newtype[A]: Newtype[A] =
>         new Newtype[A] {
>           type Type = A
>           def apply(value: A): Type = value
>         }
>     }
> }
> 
> trait NewtypeExports {
>   import NewtypeModule._
> 
>   abstract class Newtype[A] extends instance.Newtype[A] {
>     val newtype: instance.Newtype[A] = instance.newtype[A]
>     trait Tag
>     type Type = newtype.Type with Tag
>     def apply(value: A): Type = newtype(value).asInstanceOf[Type]
>   }
> }
> 
> object Example extends App with NewtypeExports {
>   
>   object Meter extends Newtype[Double]
>   type Meter = Meter.type
>   
>   Meter(3.4) // class cast exception
> }
> ```
> 
> The problem is that in `NewTypeExports` we define `trait Tag` and then `type Type = new type.Type with Tag`. Then we cast a value of type `A` to a value of type `Type`. `A` is indeed identical to `newtype.Type` so this part of the cast is valid but `Tag` is a completely arbitrary type that we just made up so I don't think it is actually sound to cast `A` to be an instance of this type. It ends up not mattering because `Tag` has no methods defined on it and we never match against `Tag` so we never do anything with it that would be invalid, but I can understand the cast being invalid at this point.
> 
> It seems like to some extent we were relying on an idiosyncracy of the Scala 2 compiler so I don't think we really have a basis for going to the Dotty team and saying there is something wrong here, if anything I think they fixed an incorrect behavior we were relying on for our trick here.
> 
> We probably need to spend some more time on this. We originally added so we could support defining type class instances for new type. Why don't we mark this test as `scala2Only` until we resolve this?
> 
> Copying @jdegoes.

joroKr21:
Enables `NewtypeSpec` on Dotty.
If we drop Scala 2.11 we could use `type Tag` instead of `trait Tag`.
Not sure if you want to close #272 but `Tag` should be universal anyway.

### DIFF
--- a/src/main/scala/zio/prelude/NewtypeModule.scala
+++ b/src/main/scala/zio/prelude/NewtypeModule.scala
@@ -308,7 +308,7 @@ trait NewtypeExports {
   abstract class Newtype[A] extends instance.Newtype[A] {
     val newtype: instance.Newtype[A] = instance.newtype[A]
 
-    trait Tag
+    trait Tag extends Any
     type Type = newtype.Type with Tag
 
     def wrapAll[F[_]](value: F[A]): F[Type] = newtype.wrapAll(value).asInstanceOf[F[Type]]
@@ -332,7 +332,7 @@ trait NewtypeExports {
   abstract class NewtypeSmart[A](assertion: Assertion[A]) extends instance.NewtypeSmart[A] {
     val newtype: instance.NewtypeSmart[A] = instance.newtypeSmart[A](assertion)
 
-    trait Tag
+    trait Tag extends Any
     type Type = newtype.Type with Tag
 
     def makeAll[F[+_]: Traversable](value: F[A]): Validation[String, F[Type]] =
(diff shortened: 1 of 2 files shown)


adamgfraser (approved):
Fantastic!

PR #306, opened 2020-09-24
add test for SubtypeSmart subtyping capability

gurghet:
Follow up to https://github.com/zio/zio-prelude/pull/303 to add a test


src/test/scala/zio/prelude/NewtypeSpec.scala
```diff
@@ -28,6 +28,12 @@ object NewtypeSpec extends DefaultRunnableSpec {
           assert(Natural.make(-1))(isFailureV(equalTo(expected)))
         }
       ),
+      suite("SubtypeSmart")(
+        test("subtypes values") {
+          val _ = implicitly[Natural <:< Int]
+          assert(true)(isTrue)
```
gurghet, line 34: or this could be `assertCompleted` I think it will be weird in any case because this is not a proper unit test but will fail at compile time

adamgfraser: `assertCompletes` would probably make the intention slightly more clear here. You could also just call a method defined on `Int` and make sure it works, like adding two numbers.

ZIO Test does have an `assertCompiles` you can use to check that code does or does not compile but that is probably heavier weight than we need here.

adamgfraser (approved):
Looks great! Thanks for adding this!

PR #304, opened 2020-09-23
add Zivariant

lemastero:
Fix #291 add Zivariant abstraction

PR #293, opened 2020-09-17
add laws for Divariant using methods returning booleans

lemastero:
Resolve first part of #290 - simple implementation of the laws (no zio-test law testing yet)

Changes:
* add  `composition` and `identity` laws for `rightMap`:
```scala
rightMap(g compose f) == rightMap(g) compose rightMap(f)
rightMap(identity) == identity
```
* add `composition` and `identity` laws for `leftContramap`:
```scala
leftContramap(f compose g) === leftContramap(g) compose leftContramap(f)
leftContramap(identity) === identity
```
* add `composition` and `identity` laws for dimap
```scala
dimap(f compose g, h compose i) === dimap(g, h) compose dimap(f, i)
dimap(identity, identity) == identity
```
* add law that enforce coherence between dimap and righMap and leftContramap
```scala
dimap(f, g)  === rightMap(g) andThen rleftContramap(f)
```
* derive `Function1Contravariant` and `Function1Covariant` instead of manually implement them
* type lambda are not neccessary in
`Divariant[({ type lambda[-A, +B] = A => B })#lambda]` - it is enough to `Divariant[Function1]`
* extra new line between trait and object avoid akward formatting
* rename `leftMap` to `leftContramap` (if we add Bifunctor like thing then we would have a name clash)

PR #309, opened 2020-09-26
Zivariant with Bicovariant and Divariant as type aliases

lemastero:
* Bicovariant and Divariant are replaced by type aliases:
```scala
 type Divariant[:=>[-_, +_]]   = Zivariant[({ type lambda[-R, +E, +A] = R :=> A })#lambda]
 type Bicovariant[:=>[+_, +_]] = Zivariant[({ type lambda[-R, +E, +A] = E :=> A })#lambda]
```
* When there are `Zivariant` instances then simpler `Covariant`, `Contravariant` instances are derived from them
* Bicovariant instances created for `Tuple2`, `Either`, `Exit`. From those Covariant instances are derived (including Failed ones)
* For `Function1` there is Divariant instance and Covariant and Contravariant instances are derived
* To easily define Bicovariant and Divariant instances there are helper interfaces: `BicovariantInstance` and `DivariantInstance`

Cons:
* When you implement instance for Divariant (same with Bicovariant) you need to deal with extra type like `E` here:
```scala
new DivariantInstance[Function1] {
  override def dimap[R, E, A, R1, A1](r: R1 => R, a: A => A1): (R => A) => R1 => A1 =
     r andThen _ andThen a
```
but thanks to Syntax it will not affect the user experience:
```scala
def leftContramap[C](ca: C => A)(implicit divariant: Divariant[:=>]): C :=> B =
  divariant.contramap(ca)(f)
```


src/main/scala/zio/prelude/package.scala
```diff
@@ -61,6 +62,9 @@ package object prelude
   type DeriveInverse[F[_]]     = Derive[F, Inverse]
   type DeriveOrd[F[_]]         = Derive[F, Ord]
 
+  type Divariant[:=>[-_, +_]]   = Zivariant[({ type lambda[-R, +E, +A] = R :=> A })#lambda]
+  type Bicovariant[:=>[+_, +_]] = Zivariant[({ type lambda[-R, +E, +A] = E :=> A })#lambda]
```
sideeffffect, line 66: I would advise against using `:=>` as a type parameter in `Bicovariant`. `:=>` is (intentionally) similar to `=>` and thus hints that the first parameter is _contra_-variant, but that's not what happens here in `Bicovariant`, right?

src/main/scala/zio/prelude/package.scala
```diff
@@ -61,6 +62,9 @@ package object prelude
   type DeriveInverse[F[_]]     = Derive[F, Inverse]
   type DeriveOrd[F[_]]         = Derive[F, Ord]
 
+  type Divariant[:=>[-_, +_]]   = Zivariant[({ type lambda[-R, +E, +A] = R :=> A })#lambda]
```
sideeffffect, line 65: This raises the question, whether we should encode mere `Covariant` and `Contravariant` as special cases of `Zivariant`.
:thinking: 
Perhaps that's something we don't need to decide right now and can leave this to a future PR.

sideeffffect:
Btw, does this make https://github.com/zio/zio-prelude/pull/279 and https://github.com/zio/zio-prelude/pull/304 obsolete?
Otherwise awesome job! :+1: :rocket:

src/main/scala/zio/prelude/package.scala
lemastero: It bothers me too. Do you have a suggestion?

sideeffffect: Would something like plain `F` do?

lemastero: It is inconvenient to write all those [] that is why I prefer symbols rather than letters. Would `<=>` work?
https://github.com/zio/zio-prelude/pull/309/commits/1a39ee3b9000d4c3ef785e6fcf00a2f94dfd1b73

src/main/scala/zio/prelude/package.scala
lemastero: Yes :)

There are few things to consider when pushing this encoding further:
- laws for Bicontravariant, laws for Divariant
- Covariant/Contravariant extends some other types and have extra methods, add them in syntax or expand the hierarchy of Trivariant 
- use asInstanceOf instead of deriving (currently it does not work b/c Covariant/Contravarian are not type aliases)
- encode the same thing with Invariant

Those will be pretty big PR's, hard to review if done at once.

lemastero:
> Btw, does this make #279 and #304 obsolete?
> Otherwise awesome job! 👍 🚀

For me there is natural evolution #293,  #279 => #304 => #309. But merging this and closing the others works for me as too.

lemastero:
cc: @neko-kai @Odomontois @adamgfraser @jdegoes

src/main/scala/zio/prelude/package.scala
sideeffffect: `<=>` suggests `Equivalence` which is invariant :disappointed: 
https://github.com/zio/zio-prelude/blob/8074366be3814b16334bfd3f98aeee14b2dff174/src/main/scala/zio/prelude/package.scala#L34
`<*>`? `&&`? Or give up and use `F`?

lemastero:
There are 2 problematic things here:
1. instance for Traversable and Bicovariant
```scala
[error] /zio-prelude/src/main/scala/zio/prelude/Invariant.scala:92:9: illegal inheritance;
[error]  <$anon: zio.prelude.Traversable[[+a]scala.util.Either[E,a]] with zio.prelude.Bicovariant[Either]> inherits different type instances of trait TriCovariant:
[error] zio.prelude.TriCovariant[[R, +E, +A]Either[E,A]] and zio.prelude.Covariant[[+a]scala.util.Either[E,a]]
[error]     new Traversable[({ type lambda[+a] = Either[E, a] })#lambda] with Bicovariant[Either] {
```
Solution: port Traversable to trifuncor encoding (you can traverse over result).
In general, it is possible to define traverse over error but I do not know if that is useful.

2. I regular encoding you have Invariant as a base class for Covariant and Contravariant. In trifunctor encoding they overlap, so one could define 2 different version of imap with 2 different names (this is suprising for users IMHO). I just ignored `invmap` in trifunctor encoding (it is present in Covariant and Contravariant thanks to the CovariantInstance (ContravariantInstance).

jdegoes:
@adamgfraser @lemastero I think these should be unified with the other type classes in `fx` and shipped as part of that module. Thoughts???

Btw am closing for now for hygiene but would like this work merged in at some point.

adamgfraser:
@jdegoes @lemastero Yes I think it would definitely be good to get this in there. Maybe we should focus it a little more on "shapes" for data types that exist in ZIO to avoid the combinatorial explosion of different combinations of variance but otherwise looks great.

PR #313, opened 2020-09-27
Improve NonEmptyList and Validation instances

sideeffffect:
`Hash` instance should be preferred over `Ord`, because it could contain short-circuiting `equal` implementation

### DIFF
--- a/src/main/scala/zio/prelude/NonEmptyList.scala
+++ b/src/main/scala/zio/prelude/NonEmptyList.scala
@@ -476,10 +476,16 @@ object NonEmptyList extends LowPriorityNonEmptyListImplicits {
     }
 
   /**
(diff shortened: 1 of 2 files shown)


sideeffffect:
That one red build is just a flaky one.
Otherwise this is ready to be merged.

adamgfraser (commented):
Couldn't either instance short circuit?

sideeffffect:
Theoretically, yes.
But the `Hash` instance is guaranteed to "do the right thing". `Ord` instance might implement `equal` in terms of `compare`, which could sometimes be slow (because sometimes it's more expensive to determine if `<` or `>`, even though you quickly know that `!==`).

adamgfraser (approved):
Okay. We need to prioritize one over the other to avoid ambiguity so might as well do it this way.

PR #314, opened 2020-09-27
Rename AssertionSyntax equal to equalTo

sideeffffect:
`AssertionSyntax.equal` alises with `Equal.equal`, which causes often problems typically when the program does not typecheck

### DIFF
--- a/src/main/scala/zio/prelude/Assertions.scala
+++ b/src/main/scala/zio/prelude/Assertions.scala
@@ -13,7 +13,7 @@ trait Assertions {
   /**
    * Makes a new assertion that requires a value equal the specified value.
    */
-  def isEqualTo[A: Equal](expected: A): Assertion[A] =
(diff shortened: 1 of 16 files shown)


src/main/scala/zio/prelude/Assertions.scala
```diff
@@ -13,7 +13,7 @@ trait Assertions {
   /**
    * Makes a new assertion that requires a value equal the specified value.
    */
-  def equalTo[A: Equal](expected: A): Assertion[A] =
+  def isEqualTo[A: Equal](expected: A): Assertion[A] =
```
adamgfraser, line 16: I think it would be nice to keep this one as `equalTo` for symmetry with the default `equalTo` assertion in ZIO Test.

sideeffffect: done :heavy_check_mark:

adamgfraser (approved):

PR #315, opened 2020-09-27
Simplify the type for `default` in `Equal` and `Hash`

sideeffffect:


adamgfraser (commented):
I'm not sure abut this one. You're right that we can use default equality or hash code to compare any values, but a lot of the time we want to use the default notion of equality / hash code but still only allow comparisons of values of the same type. Of course since it is contravariant the user can always narrow it with a type ascription but I'm not sure it is great to have an `Equal[Any]` floating around, especially with a name like `default`.

sideeffffect:
This isn't any different from what we have currently in practical terms. `default: Equal[Any]` is no more or less dangerous than `default[A]: Equal[A]`.
I just wanted to do a little cleanup, so that (if possible), we use variance instead of free type variables -- the ZIO motto at large.

adamgfraser:
I think there is some difference. You're right that as currently written if we just do `val equal = Equal.default` the compiler will infer that type as `Equal[Any]`. However, right now if I want equality for a more specific type I can get that by specifying the type:

```scala
val intEqual = Equal.default[Int] // gives me an equal instance that only compares integers
```

With this change that would no longer work and you would need an explicit type ascription:

```scala
val intEqual = Equal.default // gives me an equal instance that compares unrelated types
val intEqual2: Equal[Int] = Equal.default // now I am okay
```

Another example would be today:

```scala
Equal.default[Int].equal(x, y) // fails to compile if x and y are not both integers
```

Versus with this change:

```scala
(Equal.default: Equal[Int]).equal(x, y) // need a type ascription to get the same effect
```

I'm not sure it is a huge difference but the thing that I think separates this from other uses of variance is that universal equality is powerful enough to compare completely unrelated types but we don't actually want it to be that powerful. Most of the time you don't actually want the equal instance to be parameterized on `Any` and it seems like doing that through a type parameter versus ascription is more straightforward.

But I see your point. If we did change these from being a `def` to ` val`.

sideeffffect:
Oh, now I see that having type parameter can make some things easier (like limiting the "scope" of the instance as you mentioned).
Thanks for the explanation :+1:

Issue #283, opened 2020-09-15
Consolidate the implementation of Type class instances: Invariant


sideeffffect:
Consolidate the `Invariant` type class hierarchy

parent issue: #215

PR #316, opened 2020-09-27
Consolidate the implementation of Type class instances: Invariant

sideeffffect:
fixes https://github.com/zio/zio-prelude/issues/283


sideeffffect:
@adamgfraser this is ready to be merged. Could you please have a look at it? :pray:

adamgfraser:
@sideeffffect Yes. Will do tonight!

adamgfraser (commented):
@sideeffffect I'm a little worried about the impact on discoverability here.

There are a bunch of benefits to the approach we have been taking:
1. We consolidate all the logic related to functionality for a data type in one place
2. We eliminate the possibility of inconsistent implementations
3. We make sure that instances of subtypes are in scope when summoning a supertype

However, I think there is also a downside here that to facilitate this we are making it significantly less straightforward for the reader of the codebase. Before if you wanted the `Traversable` instance for `List` you would look in the companion object and it would be right there. Now if you want to find that instance you have to go to `Invariant`.

There is definitely logic to this in that `Invariant` is the most general super type. But now you have to understand that logic. And more than that, we're going from `Traversable`, which is a well known abstraction that is one of the go to reasons for people to use a functional programming library, to `Invariant`, which while very well defined is I think used quite infrequently in practice. And it requires having some knowledge of the entire hierarchy to understand what the super types are instead of just saying "today I understand Traversable and want to just deal with that".

This shows up most concretely in changes like going from being able to call `Traversable.ListTraversable.zipWithIndex(as)` to `Invariant.ListTraversable.zipWithIndex(as)` (using the summoner would get around this).

What do you think about doing something like?

```scala
object Invariant {

  implicit val ListInvariant: Traversable[List] =
    ???
}

object Traversable {
  implicit val listTraversable: Traversable[List] =
    Invariant.listInvariant
}
```

That way we do the "right" thing in terms of defining our logic in one place but also let users find the instances on the objects they are working with and provide them a breadcrumb to the full implementation.

sideeffffect:
I'm afraid that won't be a good way forward :disappointed: 
Let's take an example, this is what we would have to do for `NonEmptyChunk`:

```scala
object Invariant {
  implicit val NonEmptyChunkInvariant: NonEmptyTraversable[NonEmptyChunk] = ???
}
object Covariant {
  implicit val NonEmptyChunkCovariant: NonEmptyTraversable[NonEmptyChunk] = Invariant.NonEmptyChunkInvariant
}
object Traversable {
  implicit val NonEmptyChunkTraversable: NonEmptyTraversable[NonEmptyChunk] = Covariant.NonEmptyChunkCovariant
}
object NonEmptyTraversable {
  implicit val NonEmptyChunkNonEmptyTraversable: NonEmptyTraversable[NonEmptyChunk] = Traversable.NonEmptyChunkTraversable
}
```

It doesn't really scale well to copy the instance everywhere. There would be just too much boilerplate (and it would be a lot of work). And the benefit is not that great.
Scala developers are already familiar with the implicit resolution mechanism. Scala compiler behaves as if these "forwarding instances" were already there, so we don't need to put them there manually ourselves.
I worry that if we went against what Scala/the Scala compiler naturally does, it would create only more confusion and boilerplate code to read through.

regiskuckaertz:
Sorry for chiming in uninvited: isn't [given imports](https://dotty.epfl.ch/docs/reference/contextual/given-imports.html) going to throw a spanner into all that work? I understand they will need to be explicitly put in scope starting from Scala 3.

adamgfraser:
@regiskuckaertz No, more perspectives are always good! My understanding is that as long as something is in the companion object of the data type or type class it is still automatically in implicit scope. So I think the vast majority of the time nothing should change. We're running the test suite for ZIO Prelude on Scala 3 at this point and I think a lot of things would be blowing up if that wasn't the case.

So I think the change is really just if you want to explicitly import an instance then you have to use slightly different syntax. That may impact in particular the implicit syntax classes which right now you do get by `import zio.prelude._` and I think in the future would require `import zio.prelude.{ given._, _ }`.

adamgfraser:
@sideeffffect That actually seems fine to me. 😃

The total number of aliases is comparable to the number of instances we originally had, there is no code duplication since they are all aliases, and I think It is a relatively small one time cost to implement them. I would think of them kind of like the type alias companion objects in ZIO.

But we can handle that in a follow up PR if we want to.

src/main/scala-2.11-2.12/zio/prelude/InvariantVersionSpecific.scala
```diff
@@ -2,7 +2,7 @@ package zio.prelude
 
 import scala.collection.generic.CanBuildFrom
 
-trait TraversableVersionSpecific {
```
adamgfraser, line 5: Do we need to rename these? They deal exclusively with `Traversable` so that name seems to be more specific. We can still have the `Invariant` companion object extend it.

adamgfraser (approved):
Looks great! Just had one very minor comment.

sideeffffect:
@regiskuckaertz _given imports_ shouldn't change anything for us. We're placing the type class implementations in the [implicit scope](https://github.com/liancheng/notes/blob/master/scala-implicit-resolution-mechanism.rst#implicit-resulotion) (i.e. companion objects) precisely in order not to bother with any kind of imports (as used to be the case with ScalaZ or [Cats](https://meta.plasm.us/posts/2019/09/30/implicit-scope-and-cats/)).

src/main/scala-2.11-2.12/zio/prelude/InvariantVersionSpecific.scala
sideeffffect: I would feel uneasy about naming these `TraversableVersionSpecific`.
Using `*VersionSpecific` kinda like a hack, to support multiple incompatible Scala versions.

If we didn't support multiple versions, all the members would be in `Invariant`. But we want multiple Scala versions, so we need to put some members into `InvariantVersionSpecific`. But those members are still "logically" members of `Invariant`, so that's why I would prefer to reflect that in the `*VersionSpecific` trait name. Not doing that would likely confuse other people.

sideeffffect:
One dead build needs restarting. Otherwise, this is ready to be merged :rocket:

src/main/scala-2.11-2.12/zio/prelude/InvariantVersionSpecific.scala
adamgfraser: Yeah, I was thinking more about the contents of the file, which while they are `Invariant` instances are more specifically `Traversable` instances. But I don't think it really matters.

adamgfraser (approved):

PR #318, opened 2020-09-28
Implement FlatMap Syntax

### LINKED ISSUE #317: Add flatMap
Not having `flatMap` on contextual abstractions prevents users from co-opting the syntactic sugar provided by the language. I know the reasoning behind it but I think this is a severe impediment to usability. Why not add it as an extension method on `Monad`?

adamgfraser:
Resolves #317. If instances of `IdentityFlatten` and `Covariant` are defined for a type then `flatMap` should be available through implicit syntax.


regiskuckaertz (approved):
That was fast, thank you 🙏

Issue #323, opened 2020-09-30
 Document the Type Class hierarchies with chart: Associative


sideeffffect:
Document the `Associative` hierarchy in a visual way.

parent issue: https://github.com/zio/zio-prelude/issues/321

PR #337, opened 2020-10-10
Add diagram for the Associative hierarchy

sideeffffect:
Missing instances that were uncovered in the process are added.
Doc comments are also improved.

fixes https://github.com/zio/zio-prelude/issues/323

This turned out to me a much longer and wilder ride than I originally anticipated. Sorry for the amount of changed code. Please let me know @adamgfraser if you would like me to split it into smaller PRs.

[Rendered here](https://github.com/sideeffffect/zio-prelude/blob/323-type-class-diagram-Associative/docs/overview/diagrams.md)


sideeffffect:
@adamgfraser Do you think you could have a look at this?

adamgfraser (approved):
Looks great!

PR #338, opened 2020-10-14
Start using release-drafter

sideeffffect:


sideeffffect:
@adamgfraser What do you think about starting using it?

adamgfraser:
@sideeffffect I think it is a good idea. We used it in ZIO for the last minor release we did. At least in that case I found myself making a significant number of manual edits but just the fact that it gathers all the PRs with the author and name of the PR is a big time saver as that can take significant effort.

adamgfraser (approved):

PR #339, opened 2020-10-14
Derive (Associative/Identity)Both from Covariant and (Associative/Identity)Flatten

sideeffffect:
Is this what you had in mind @felher?
As mentioned on Discord, these derivations can't be automatic, because that would conflict with existing `Associative`/`Identity` `-Flatten` instances.
/cc @adamgfraser


sideeffffect:
related to https://github.com/zio/zio-prelude/issues/340

No matter how I tried, wherever I placed the derivations or instances... when I make the derivations automatic (i.e. `implicit`), there are ambiguities.
Perhaps somebody will be able to crack this nut, but so far this (non-automatic derivation) seems to me like the only way.

sideeffffect:
@adamgfraser I think we can merge this as is. Maybe in the future, somebody will figure out how to make the derivations automatic.

adamgfraser:
@sideeffffect Sounds good to me!

adamgfraser (approved):

PR #350, opened 2020-10-21
Remove ThrowableHash out of implicit scope

sideeffffect:
It would allow comparing _all_ Throwables across hierarchies defined by users, which would typically be a mistake.


jdegoes (approved):

jdegoes:
Agreed, and good catch!

PR #351, opened 2020-10-23
Bring Ring-like structures to ZIO Prelude

sideeffffect:
This is the foundation to have Ring (and the whole family, like Semiring, Field, ...) algebraic structures and instances for Scala library/ZIO types.
You can see the skeleton of the hierarchy here:
https://github.com/sideeffffect/zio-prelude/blob/ring/docs/functional-abstractions/abstraction-diagrams.md (named _DistributiveProd_)
More is still to come, mainly comments/documentation and the instances for more types, currently there is just `Int` and `Double` to try things out.
I would be interested in your feedback :smiley: 

/cc @adamgfraser @jdegoes


jdegoes:
I think the cornerstone of this hierarchy should be `Distributive` (a semiring), in keeping with other naming conventions.

In addition, a distributive is really a pair of other binary operators that satisfy certain laws. So that would argue the type class itself should contain a proof of those, possibly in the form of instances for the other two binary operators.

In fact, maybe there's some encoding where a type class is parameterized by the properties of the two binary operators, e.g.:

```scala
trait Distributive[Add[x] <: Associative[x], Multiply[x] <: Associative[x], A] {
  def add: Add[A]
  def multiply: Multiply[A]
}
```

sideeffffect:
Thanks John for looking at this!
If we look at [`DistributiveMultiply`](https://github.com/sideeffffect/zio-prelude/blob/ring/src/main/scala/zio/prelude/DistributiveMultiply.scala), we can see that it this will indeed be the most useful type class for most users' needs.
But I wanted to be able to express a general [Semiring](https://en.wikipedia.org/wiki/Semiring). For this, a mere
```scala
DistributiveMultiply[A, Addition = CommutativeMonoid, Multiplication = Identity]
```
isn't enough. There is also another law specifying that 0 (neutral element for addition) annihilates when used in multiplication. This is where [`AnnihilatingZero`](https://github.com/sideeffffect/zio-prelude/blob/ring/src/main/scala/zio/prelude/AnnihilatingZero.scala) comes into play, for Semiring, it is
```scala
AnnihilatingZero[A, Addition = CommutativeMonoid, Multiplication = Identity]
```

Both `DistributiveMultiply` and `AnnihilatingZero` share a common structure, they need to allow two operations: addition and multiplication, where each may satisfy some laws. This shared structure is captured in [`AddMultiply`](https://github.com/sideeffffect/zio-prelude/blob/ring/src/main/scala/zio/prelude/AddMultiply.scala), which doesn't constrain these two operations besides being `Associative` (which is the top of its hierarchy).

(I opted for more verbose `DistributiveMultiply`, because there is another distributive comming, a `DistributiveJoinMeet`, when we will be adding Lattices & co.)

Does that make sense?

---

Btw one thing I was considering is whether the Addition and Multiplication proofs should be straight on `A`
```scala
  def Addition: Addition[A]
  def Multiplication: Multiplication[A]
```
but I settled on `Sum`/`Prod`
```scala
  def Addition: Addition[Sum[A]]
  def Multiplication: Multiplication[Prod[A]]
```
The reason is that since "adding" and "multiplying" the vales are both valid ways of combining them, us (ZIO Prelude developers) _and_ our users, will defined Type class instances not directly on `A`s, but in terms of `Sum`/`Prod` and will be able to directly reuse them here.
There are helper methods `add`/`multiply`, which can be overridden to enable more efficient implementation (bypassing the `apply` and `unwrap` methods on `Sum`/`Prod`).

sideeffffect:
I've reworked the proposal a bit, can you @jdegoes have another look, please?

One can start by having a look at the diagram (linked in the PR description). Now, there are only `DistributiveMultiply` and `AnnihilatingZero` as prominent _type classes with laws_.

The other traits were demoted to mere "Shape" traits. These are just helper traits containing some useful type parameters and/or methods. I've borrowed the term from this ticket https://github.com/lampepfl/dotty/issues/9028 (which borrows it from the _Getting F-Bounded Polymorphism into Shape_ paper).
I think this move was a great improvement, because keeps with the ZIO Prelude principle that each Type class _must_ come with a set of laws. One way of looking at it is that the (nominal) type of the Type class trait _is_ the law(s). We double down on deconstructing things and composing them (these individual laws) horizontally via Scala's intersection types, instead of creating vertical hierarchies.

src/main/scala/zio/prelude/AddMultiplyShape.scala
```diff
@@ -0,0 +1,100 @@
+package zio.prelude
+
+import zio.prelude.newtypes.{ Prod, Sum }
+
+trait AddMultiplyShape[A, +Addition[x] <: Associative[x], +Multiplication[x] <: Associative[x]] {
```
jdegoes, line 5: This is the one I think should be called `Distributive`, because the fundamental lie here that binds the two binary operators is `a * (b + c) == a * b + a * c`. Then other laws add additional structure.

src/main/scala/zio/prelude/AddMultiplyShape.scala
```diff
@@ -0,0 +1,100 @@
...
+
```
jdegoes, line 6: As an experiment, we could see about moving `Addition` and `Multiplication` to type members, and using the `Aux` pattern (refinement typing), introduce type aliases for common combinations.

This is only practical if Scala can find the implicits for the common combinations of properties of the two binary operators.

The possible benefits (if workable):

 - Fewer type parameters
 - Easy to make type aliases for common combinations and require those concisely

src/main/scala/zio/prelude/AddMultiplyShape.scala
```diff
@@ -0,0 +1,100 @@
...
+  def add(l: => A, r: => A): A =
+    Sum.unwrap(Addition.combine(Sum(l), Sum(r)))
+
+  def multiply(l: => A, r: => A): A =
+    Prod.unwrap(Multiplication.combine(Prod(l), Prod(r)))
+
+  def Addition: Addition[Sum[A]]
```
jdegoes, line 13: After looking at this, probably lower case `A` and `M` for `Addition` and `Multiplication` fields.

src/main/scala/zio/prelude/AnnihilatingZero.scala
```diff
@@ -0,0 +1,81 @@
+package zio.prelude
+
+import zio.prelude.coherent.AnnihilatingZeroEqual
+import zio.prelude.newtypes.{ Prod, Sum }
+import zio.test.TestResult
+import zio.test.laws.{ Lawful, Laws }
+
+trait AnnihilatingZero[A, +Addition[x] <: Identity[x], +Multiplication[x] <: Associative[x]]
```
jdegoes, line 8: In Prelude, `Identity` is used for the identity law, and `identity` for the identity element.

This suggests, perhaps, that `Annihilation` would be the name of this type class, and `annihilation` the name of the "anihilator" element.

src/main/scala/zio/prelude/DistributiveMultiply.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * The left distributivity law states that for operators `+` and `*`, for all
+   * values `a1`, `a2`, and `a3`, the following must hold:
+   *
+   * {{{
+   * a1 * (a2 + a3) === (a1 * a2) + (a1 * a3)
+   * }}}
+   */
+  val leftDistributivityLaw: Laws[DistributiveMultiplyEqual] =
```
jdegoes, line 20: Why are these laws introduced here instead of in AddMultiply ("Distributive")? What useful (non-esoteric) structures are there without the distributive law?

src/main/scala/zio/prelude/DivideShape.scala
```diff
@@ -0,0 +1,49 @@
+package zio.prelude
+
+import zio.prelude.newtypes.{ Prod, Sum }
+
+trait DivideShape[A, +Addition[x] <: Associative[x], +Multiplication[x] <: InverseNonZero[x]]
```
jdegoes, line 5: I'd like to brainstorm a bit on the names here. This definitely takes it down a certain path. 

In addition, the alternate (type member) encoding could clean things up a bit here, if it ends up working out.

src/main/scala/zio/prelude/InverseNonZero.scala
```diff
@@ -0,0 +1,909 @@
...
+ *
+ * Because `InverseNonZero` defines a binary rather than a unary operator it can be
+ * used to describe inverse operations for types that do not have inverse
+ * values. For example, the natural numbers do not have inverses because the
+ * set of natural numbers does not include negative numbers. But we can still
+ * define a subtraction operation that is the inverse of addition for the
+ * natural numbers, since subtracting a number from itself always returns
+ * zero.
+ */
+trait InverseNonZero[A] extends Identity[A] {
```
jdegoes, line 24: Unfortunately, I don't think this is very well defined, because the type class does not define a "zero" element. In general for an inverse, you don't need to speak about zero elements. 

We should think about other ways to achieve the goal of making this definable for division (or giving up). Such as `Inverse[A, B <: A]` or something.

src/main/scala/zio/prelude/package.scala
```diff
@@ -73,6 +76,18 @@ package object prelude
     type Semilattice[A]        = Commutative[A] with Idempotent[A]
     type BoundedSemilattice[A] = Semilattice[A] with Identity[A]
 
+    type Semiring[A] =
+      AnnihilatingZero[A, CommutativeMonoid, Identity] with DistributiveMultiply[A, CommutativeMonoid, Identity]
+    type Ring[A]     =
+      AnnihilatingZero[A, AbelianGroup, Identity]
+        with DistributiveMultiply[A, AbelianGroup, Identity]
+        with SubtractShape[A, AbelianGroup, Identity]
+    type Field[A]    =
```
jdegoes, line 85: These "proof of concepts" are useful and we should definitely include them.

src/main/scala/zio/prelude/AddMultiplyShape.scala
sideeffffect: Both `AnnihilatingZero` and `DistributiveMultiply` type classes share a common "shape". I introduced the trait `AddMultiplyShape` so that both type classes can inherit from it.

src/main/scala/zio/prelude/AddMultiplyShape.scala
sideeffffect: I actually tried that, but there were some downsides and taken all together, I didn't feel the pros outweigh the cons (mainly verboseness and it's an all new concept, not yet used in ZIO Prelude).
But I didn't do it with the `Aux` helpers!

I will give it another try, now with `Aux` and will report with results :+1:

src/main/scala/zio/prelude/AddMultiplyShape.scala
sideeffffect: That is definitely a reasonable proposition.
But then I don't know, how we will do it with lattices. Here, the operation ("addition") has a different name that the (imperative) method ("add"). But in a lattice, the operation and the method are both called "join".
So how do we resolve this?

src/main/scala/zio/prelude/AddMultiplyShape.scala
```diff
@@ -0,0 +1,100 @@
...
+
+    override def Multiplication: Commutative[Prod[Int]] with Identity[Prod[Int]] =
+      Associative.IntProdCommutativeIdentity
+  }
+
+  implicit val DoubleAnnihilatingZeroDistributiveMultiply: classic.Field[Double] = new Field[Double] {
+    override def add(l: => Double, r: => Double): Double      = l + r
+    override def divide(l: => Double, r: => Double): Double   = l / r
+    override def multiply(l: => Double, r: => Double): Double = l * r
+    override def subtract(l: => Double, r: => Double): Double = l - r
```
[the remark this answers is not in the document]
sideeffffect: :smile_cat:

src/main/scala/zio/prelude/AddMultiplyShape.scala
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * Add two values.
+     */
+    def add(r: => A)(implicit associative: AddMultiplyShape[A, Associative, Associative]): A =
+      associative.add(l, r)
+
+    /**
+     * A symbolic alias for `multiply`.
+     */
+    def ***(r: => A)(implicit associative: AddMultiplyShape[A, Associative, Associative]): A =
```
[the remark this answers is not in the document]
sideeffffect: There's not. This was just my invention, because mere `+` didn't work. So that's why I opted for a triplet `+++`. `*` works fine, but I wanted the symbols to be "symmetrical".
I suspect that I might have done some mistake. Do you know what I did wrong and/or how to make `+` work as an extension method?

src/main/scala/zio/prelude/AnnihilatingZero.scala
sideeffffect: Good idea! :+1:

src/main/scala/zio/prelude/DistributiveMultiply.scala
sideeffffect: > What useful (non-esoteric) structures are there without the distributive law?
If there are any, I have no idea of their existence. (I'm no mathematician, though :smile: )
But that was not the reason why I introduced both `AnnihilatingZero` and `DistributiveMultiply`.

The traditional algebraic structure _Semiring_ requires (among others) these two, independent, laws: (1) annihilating zero and (2) distributing multiply. That's why I introduced two independent type classes: `AnnihilatingZero` and `DistributiveMultiply`.

Do we want to express it in another way? What would be an alternative? `AnnihilatingZero` inheriting from `DistributiveMultiply`? Or vice versa? :thinking:

Btw, I intentionally named it verbosely `DistributiveMultiply`, so that it's distinct from `DistributiveJoinMeet` (for Lattices).

src/main/scala/zio/prelude/DivideShape.scala
sideeffffect: Happy to change the naming! Any suggestions?
(I will definitely try the type member encoding again.)

src/main/scala/zio/prelude/InverseNonZero.scala
sideeffffect: I have to admit that `InverseNonZero` is quite un-elegant and ugly :crying_cat_face:, I just didn't know how to do it better.

> Such as Inverse[A, B <: A] or something.

I'm not sure what you mean by this. Could you please expand on this? Hopefully this will get us to a better place :crossed_fingers:

src/main/scala/zio/prelude/package.scala
sideeffffect: :+1:

src/main/scala/zio/prelude/AddMultiplyShape.scala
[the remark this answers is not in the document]
sideeffffect: Btw the errors I get look plainly like this
```scala
[error] /home/ondra/Projects/zio-prelude/src/main/scala/zio/prelude/DistributiveMultiply.scala:23:21: value + is not a member of type parameter A
[error]         (a1 *** (a2 + a3)) <-> ((a1 *** a2) + (a1 *** a3))
[error]                     ^
[error] /home/ondra/Projects/zio-prelude/src/main/scala/zio/prelude/DistributiveMultiply.scala:23:45: value + is not a member of type parameter A
[error]         (a1 *** (a2 + a3)) <-> ((a1 *** a2) + (a1 *** a3))
[error]                                             ^
[error] /home/ondra/Projects/zio-prelude/src/main/scala/zio/prelude/DistributiveMultiply.scala:37:14: value + is not a member of type parameter A
[error]         ((a1 + a2) *** a3) <-> ((a1 *** a3) + (a2 *** a3))
[error]              ^
[error] /home/ondra/Projects/zio-prelude/src/main/scala/zio/prelude/DistributiveMultiply.scala:37:45: value + is not a member of type parameter A
[error]         ((a1 + a2) *** a3) <-> ((a1 *** a3) + (a2 *** a3))
[error]                                             ^
[error] four errors found
```

src/main/scala/zio/prelude/AnnihilatingZero.scala
sideeffffect: Renamed :heavy_check_mark:

src/main/scala/zio/prelude/AddMultiplyShape.scala
sideeffffect: I've reworked the codebase, changing those type parameters to type member while utilizing the `Aux` pattern, where handy.
And I think it's an improvement. The code feels more Scala-y and I like it :sunglasses: 
If you're not worried it would feel overwhelming to new ZIO Prelude users, I think we can go forward with it.

Please have a look yourself and tell me what you think :wink:

adamgfraser:
Following up on a conversation @sideeffffect and I had today, I think one encoding of this we could explore is:

```scala
package zio.prelude

trait Semiring[A] {
  def add(a1: A, a2: A): A
  def multiply(a1: A, a2: A): A
}

object Semiring {

  val laws = ???
  // Distributive law

  implicit def derive(implicit sum: Associative[Sum[A]], product: Product[A]): Semiring[A] =
    new Semiring[A] {
      def add(a1: A, a2: A): A
      def multiply(a1: A, a2: A): A
    }
}

trait SemiringWithIdentity[A] extends Semiring {
  def identity: A
}

object SemiringWithIdentity {

  val laws = ???
  // annihilation

  implicit def derive(implicit sum: Inverse[Sum[A]], product: Product[A]): SemiringWithIdentity[A] =
    ???
}
```

The idea is that we should be able to derive a `Semiring` where multiplication distributes over addition given an `Associative[Sum[A]]` and an `Associative[Prod[A]]`. Similarly we should be able to derive a version where there is also a zero element whenever we have an `Associative[Inverse[A]]` and an `Associative[Prod[A]]`.

I don't think this adds functionality beyond what you already have but it would be significantly simpler (no type members, no Aux pattern, etc...). I also like it conceptually in that it makes the individual binary operations fundamental and then this type class just describes that there are two binary operations that are related in a certain way. It also avoids the need for any new implementations of the type classes because we can just delegate to the existing instances.

One consideration is that the validity of the instances here depend on the validity of the instances that the user defines for `Sum` and `Prod`. For example it would be possible to implement versions of `Sum` and `Prod` that don't satisfy the laws, say by implementing `Sum` in terms of multiplication and `Prod` in terms of addition. But I think it is fine to say that an erroneous definition. If you are going to define instances of `Sum` and `Prod` for the same type then those should relate to each other in the way we would expect in terms of multiplication distributing. If that is true then this derivation is valid.

jdegoes:
@adamgfraser I like this SOOO much more and hope it works out.

adamgfraser:
@sideeffffect Did you get a chance to explore this more? Do you want me to try to flesh it out?

jdegoes:
Other comments:

 - We still need to think about names
 - Technically the mere existence of product and sum should not necessarily imply that products distribute over sums, perhaps, but we could solve that if we wanted to with a no-op type class such as:
    ```scala
    trait Distributive[A]
    object Distributive extends Lawful[...] { ... }
    ```
    Then merely "declaring" a distributive for some type, together with the sum and product, would derive the distributive. Maybe even the distributive would contain the add / multiply inside them. But all sum / product operations would go through the sum / product new types.
- We should have special case combine operators, one for product, and one for sum. These would "go through" the appropriate newtypes

core/shared/src/main/scala/zio/prelude/InverseNonZero.scala
```diff
@@ -0,0 +1,930 @@
...
+ *
+ * Because `InverseNonZero` defines a binary rather than a unary operator it can be
+ * used to describe inverse operations for types that do not have inverse
+ * values. For example, the natural numbers do not have inverses because the
+ * set of natural numbers does not include negative numbers. But we can still
+ * define a subtraction operation that is the inverse of addition for the
+ * natural numbers, since subtracting a number from itself always returns
+ * zero.
+ */
+trait InverseNonZero[A] extends Identity[A] {
```
jdegoes, line 26: I would say, instead of this, let's define a type class called `PartialInverse`, which has `(A, A) => Option[A]`. Then ordinary `Inverse` can extend that.

jdegoes: Not saying it's a great idea but if we saw this pattern a lot, we could think about `InverseM`, `OrdM`, etc., with an "effectful" version of the ordinary type class. Then one common specialization would be with `F[_] = Option`, although there are certain cases where, for example, you can only tell if two things are equal by performing an effect.

jdegoes: /cc @adamgfraser

experimental/shared/src/main/scala/zio/prelude/experimental/package.scala
```diff
@@ -0,0 +1,24 @@
...
+
+    type Semiring[A] = Annihilation[A] with DistributiveMultiply[A] {
+      type Addition[x] <: CommutativeMonoid[x]
+      type Multiplication[x] <: Identity[x]
+    }
+    type Ring[A]     = Semiring[A] with SubtractShape[A] {
+      type Addition[x] <: AbelianGroup[x]
+      type Multiplication[x] <: Identity[x]
+    }
+    type Field[A]    = Ring[A] with DivideShape[A] {
```
jdegoes, line 17: No matter how we change this PR, we should keep in mind our end goal is this level of modularity. ❤️

sideeffffect:
I've implemented the `PartialInverse`. I will need more time to think through and implement the other ideas you guys have suggested. Hopefully I will have more time to spend on this in the near future.

sideeffffect:
I can imagine you guys are super busy with ZIO 2 right now, but is there something I can do to help this move forward?
@jdegoes @adamgfraser

adamgfraser:
@sideeffffect I'm good with merging this into the `experimental` package. Can we move everything in there? Looks like there is still some stuff in `core`.

sideeffffect:
Unfortunately, for this Ring-like thing to work, `PartialInverse` is necessary and that needs to live in `core`. Do you have concerns about any particular change in `core` that I could fix?

adamgfraser:
Okay. Let me look in more detail. My concern is more about overall increase in complexity of the hierarchy for core functional abstractions by adding new ones that I think largely only have applicability in the math context (you basically want to describe division).

jdegoes:
@adamgfraser Agree with Adam! Would be good to get this in, and just under an experimental package so we have the flexibility to evaluate it for a while. Thank you for your persistence on this!

sideeffffect:
I know this has literally been years, but I'd appreciate if any interested parties could have another look at this :pray:

kyri-petrou:
@sideeffffect I'm not sure I'm the best person to review this because my understanding of algebras is basic at best.

Are there any specific parts of the code you'd like me to focus on?

jdegoes (dismissed):

sideeffffect:
The most important thing to check is that it doesn't break any thing in the non-experimental parts. So if some match in the experimental part is "broken", doesn't really matter that much. We can always fix it later. The experimental parts are wild-west with not backward guarantees, but wouldn't want to release ZIO Prelude 1.0 with broken non-experimental parts.
I'll double check tomorrow. But if somebody feels brave, you can also check it yourself.

After this is merged, it would be wise to release another RC and give some time for people to test it out. Only after that, we should release 1.0.0. Does that sound good @ghostdogpr ?

ghostdogpr:
Sounds good 👍

sideeffffect:
I'll need one more approval and then let's merge :muscle:

jdegoes (approved):

PR #352, opened 2020-10-23
Syntax to reduce associative elements in a parallel manner

sideeffffect:


src/main/scala/zio/prelude/Associative.scala
```diff
@@ -1272,4 +1272,16 @@ trait AssociativeSyntax {
...
+
+  /**
+   * Provides syntax for combining values in a `ParSeq` which is thus done in a parallel manner.
+   */
+  implicit class AssociativeParSeqOps[A](v: collection.parallel.immutable.ParSeq[A]) {
+
+    /**
+     * Associatively combines the values in a parallel manner.
+     */
+    def reduceAssociative(implicit associative: Associative[A]): A =
```
jdegoes, line 1284: How about `reducePar`?

sideeffffect: I assume people will very often use it with `.par` right in front of the reduction, like `someSeqCollection.par.reduceAssociative`.
If that that is the case, wouldn't you be worried that `someSeqCollection.par.reducePar` could feel a bit repetitive on the "par" part? :)

I don't really mind, so if you think it's a good idea, let's do this :+1:

sideeffffect:
We're green :tada: Let me know what you think of this, @jdegoes :pray:

src/main/scala/zio/prelude/Associative.scala
jdegoes: How about just `reduce`, then?

sideeffffect: That method is already present "natively" in `ParIterable`, so I would be worried that could cause some confusion/ambiguity for our users, or won't work outright.
What do you think?
(I've reworked the code a bit, so you might want to have another look.)

sideeffffect: I've just tried renaming `IdentityParIterableOps#reduceIdentityBlocking` to just `reduce` and compilation in `ParIterable non-empty returns a value` fails with
```scala
[error] /home/ondra/Projects/zio-prelude/src/test/scala/zio/prelude/IdentitySpec.scala:63:61: missing argument list for method reduce in trait ParIterableLike
[error] Unapplied methods are only converted to functions when a function type is expected.
[error] You can make this conversion explicit by writing `reduce _` or `reduce(_)` instead of `reduce`.
[error]             assert(List(Sum(1), Sum(2), Sum(3), Sum(4)).par.reduce)(equalTo(Sum(10)))
[error]                                                             ^
[error] one error found
```

src/main/scala/zio/prelude/Identity.scala
```diff
@@ -817,4 +817,24 @@ trait IdentitySyntax {
...
+     * while blocking the thread.
+     */
+    def reduceIdentityBlocking(implicit identity: Identity[A]): A =
+      p.fold(identity.identity)(identity.combine(_, _))
+
+    /**
+     * Returns an effect, that associatively combines the values in a parallel manner, returning the `identity` element if empty,
+     * while ensuring the current thread isn't blocked.
+     */
+    def reduceIdentity(implicit identity: Identity[A]): zio.RIO[zio.blocking.Blocking, A] =
```
sideeffffect, line 837: AFAIK, ZIO Prelude is moving to having ZIO proper only as an _optional_ dependency. In light of that, do you think we should change this signature?
Signature like this
```scala
    def reduceIdentity(blocker: zio.blocking.Blocking)(implicit identity: Identity[A]): zio.Task[A]
```
would "force" the caller to supply a ZIO object. That could prevent runtime error in cases ZIO is not on classpath (which could happen, since it would be an optional dependency).

@jdegoes @adamgfraser

jdegoes: I think that makes it more painful to use without any real gain. A user can always add ZIO if they want to use these optional methods.

sideeffffect:
This is then ready to merge from my POV. Only one build needs to be restarted.

src/main/scala/zio/prelude/Associative.scala
```diff
@@ -1272,4 +1272,24 @@ trait AssociativeSyntax {
...
+  /**
+   * Provides syntax for combining values in a `ParIterable` which is thus done in a parallel manner.
+   */
+  implicit class AssociativeParIterableOps[A](private val p: collection.parallel.immutable.ParIterable[A]) {
+
+    /**
+     * Associatively combines the values in a parallel manner,
+     * while blocking the thread.
+     */
+    def reduceAssociativeBlocking(implicit associative: Associative[A]): Option[A] =
```
jdegoes, line 1285: Why the name 'reduceAssociativeBlocking' for this one?

jdegoes: Nevermind, I see why. I suggest `reduceAssociative` for this one, and `reduceAssociativeM` for the other one, keeping with the ZIO convnetion of `M` = effectful variant.

src/main/scala/zio/prelude/Identity.scala
```diff
@@ -817,4 +817,24 @@ trait IdentitySyntax {
...
+  /**
+   * Provides syntax for combining values in a `ParIterable` which is thus done in a parallel manner.
+   */
+  implicit class IdentityParIterableOps[A](private val p: collection.parallel.immutable.ParIterable[A]) {
+
+    /**
+     * Associatively combines the values in a parallel manner, returning the `identity` element if empty,
+     * while blocking the thread.
+     */
+    def reduceIdentityBlocking(implicit identity: Identity[A]): A =
```
jdegoes, line 830: Ditto for these names.

src/main/scala/zio/prelude/Associative.scala
sideeffffect: :heavy_check_mark:

src/main/scala/zio/prelude/Identity.scala
sideeffffect: :heavy_check_mark:

jdegoes (approved):

PR #353, opened 2020-10-26
[RFC] Lattice foundation

sideeffffect:
This is the foundation to have Lattice (and the whole family, like Distributive lattice, ...) algebraic structures and instances for Scala library/ZIO types.
You can see the skeleton of the hierarchy here:
https://github.com/sideeffffect/zio-prelude/blob/lattice/docs/overview/diagrams.md (named _JoinMeet_)
More is still to come, other laws, comments/documentation and instances for more types, currently there is just `Set[A]` to try things out.
I would be interested in your feedback :smiley:

/cc @adamgfraser @jdegoes

There are some similarities to Ring-like structures, which are discussed in this PR https://github.com/zio/zio-prelude/pull/351 so maybe we can gain something when we will be able how the design decisions play out in both Type class hierarchies.


adamgfraser (approved):
Sounds good. Let's get this in and then I agree we will have to see how it works.

sideeffffect:
:tada: :tada: :tada:
:heart:

PR #354, opened 2020-10-28
Ord: Add eqConsistencyLaw

sideeffffect:


sideeffffect:
It seems like this law missing has been a mere oversight, a `Hash` type class has an analogous law.
What do you think about adding it @adamgfraser ?

adamgfraser (approved):
Yes I think this makes sense.

PR #355, opened 2020-10-29
AssociativeBothTupleOps

sideeffffect:
/cc @DamianReeves


sideeffffect:
I think this also can go in. Or would you like to see some changes @adamgfraser ? I remember you mentioned something about the type annotations :smiley:

adamgfraser (approved):
@sideeffffect I think since this has been open for a while we should get it in and if we can figure out a way to clean up the type signatures later great.

PR #357, opened 2020-10-31
Functions A => A form a Monoid

sideeffffect:


adamgfraser (approved):
You mean it can be combined with an associative operation that has an identity element? 😉

Issue #329, opened 2020-09-30
 Document the Type Class hierarchies with chart: AssociativeCompose


sideeffffect:
Document the `AssociativeCompose` hierarchy in a visual way.

parent issue: https://github.com/zio/zio-prelude/issues/321

    ```mermaid
    classDiagram
      AssociativeCompose~:=>[-_, +_]~ <|-- IdentityCompose~:=>[-_, +_]~
      class AssociativeCompose~:=>[-_, +_]~{
      }
      class IdentityCompose~:=>[-_, +_]~{
      }
    ```

sideeffffect:
Note that this has already open PR :point_up:

ernestochero:
@sideeffffect 
Is possible to take this, what is the status of #358 ?

sideeffffect:
Hopefully close to merging. But worry not, there are other, very similar documentation tickets. Look at the parent issue. It would immensely help us if you documented a Type class hierarchy that hasn't been documented yet.

I'm here for you, if you'd need any help with it :wink:

PR #358, opened 2020-10-31
AssociativeCompose: add Application, Both, Either and diagram

sideeffffect:
fixes https://github.com/zio/zio-prelude/issues/329

Overview rendered [here](https://github.com/sideeffffect/zio-prelude/blob/compose-hierarchy/docs/overview/diagrams.md#associativecompose)


sideeffffect:
I think this would be a great addition to the `AssociativeCompose` hierarchy. What do you think @adamgfraser ?
I'm not sure about the naming and which parameters should be _by-name_, though, so I would especially appreciate feedback on that :pray:

src/main/scala/zio/prelude/EitherCompose.scala
```diff
@@ -0,0 +1,32 @@
...
+
+    /** A symbolic alias for `fromEither`. Composes `A -> C` with `B -> C` to form `A or B -> C`. */
+    def |||[B, :+:[+_, +_]](implicit either: EitherCompose[:=>, :+:]): (=> B :=> C) => ((A :+: B) :=> C) =
+      either.fromEither(a2b)
+
+    /** Composes `A -> C` with `B -> C` to form `A or B -> C`. */
+    def fromEither[B, :+:[+_, +_]](implicit either: EitherCompose[:=>, :+:]): (=> B :=> C) => ((A :+: B) :=> C) =
+      either.fromEither(a2b)
+  }
+}
```
lemastero, line 32: For EitherCompose and BothCompose it should be possible to have instance for:
* Function1, Either, Void
* Function1, Tuple2, Unit
You would have to change Nothing to be polymorphic.

lemastero: If you combine both EitherCompose and BothCompose using the same types you would end up with MonoidalCategory and it has this nice property that it does not matter how you stack together :*: and :=> it always makes sense.

Instances mentioned ☝️ can be used as utilities to re-pack tuples and either.

adamgfraser:
Should we move this to experimental package and get it in?

src/main/scala/zio/prelude/EitherCompose.scala
sideeffffect: Sorry for responding so late, I've been a bit busy. How have you been doing? :smile_cat: 

I'm not sure I understand :thinking: 
There are already instances 
 * `Function[-*, +*] [type :*: = ❨_, _❩; type :--> = Function]` for `ApplicationCompose`
	* How would you make it work with `Any`? Where would you put this `Any`?
 * `Function[-*, +*] [type :+: = Either]` for `EitherCompose`
	* How would you make it work with `Nothing`? Where would you put this `Nothing`?

sideeffffect:
@adamgfraser this is good to be merged :+1:

experimental/shared/src/main/scala/zio/prelude/experimental/ApplicationCompose.scala
```diff
@@ -0,0 +1,53 @@
+package zio.prelude
+package experimental
+
+trait ApplicationCompose[:=>[-_, +_]] extends BothCompose[:=>] {
+
+  type :-->[-_, +_]
```
jdegoes, line 6: This is a lot of generality.  What are motivating cases to support `:-->` and `:=>` simultaneously?

experimental/shared/src/main/scala/zio/prelude/experimental/EitherCompose.scala
```diff
@@ -0,0 +1,62 @@
+package zio.prelude
+package experimental
+
+trait EitherCompose[:=>[-_, +_]] extends AssociativeCompose[:=>] {
```
jdegoes, line 4: Why not `ComposeEither`, `ComposeBoth`?

experimental/shared/src/main/scala/zio/prelude/experimental/package.scala
```diff
@@ -0,0 +1,15 @@
...
+
+package object experimental extends ApplicationComposeSyntax with BothComposeSyntax with EitherComposeSyntax {
+
+  object classic {
+
+    import zio.prelude.classic._
+
+    type CartesianCategory[:=>[-_, +_], :*:[+_, +_]]                     = Category[:=>] with BothCompose.Aux[:=>, :*:]
+    type ClosedCartesianCategory[:=>[-_, +_], :*:[+_, +_], :-->[-_, +_]] = CartesianCategory[:=>, :*:]
+      with ApplicationCompose.Aux[:=>, :*:, :-->]
```
jdegoes, line 11: Like to see this! ❤️

experimental/shared/src/main/scala/zio/prelude/experimental/EitherCompose.scala
sideeffffect: To keep things regular. It's always `*Compose`, like `AssociativeCompose` or `IdentityCompose`. It is like this for all other type class names, like `*Both`, `*Either`, `*Flatten`, etc.
Flipping the order just here in this case would most likely only confuse people.

experimental/shared/src/main/scala/zio/prelude/experimental/ApplicationCompose.scala
sideeffffect: I don't know, to be honest. I just know that that's the way CCC is usually defined (in Haskell).
https://hackage.haskell.org/package/categories-1.0.7/docs/Control-Category-Cartesian-Closed.html
I learned about this topic here, btw https://youtu.be/SVRYcrhRCes?t=449

sideeffffect:
Open question is, what would a Co-_Cartesian Closed Category_ look like...
```scala
trait UnapplicationCompose[:=>[-_, +_]] extends EitherCompose[:=>] {

  type :<--[-_, +_]

  def unapplication[A, B]: B :=> ((A :<-- B) :+: A)
  def cocurry[A, B, C](f: C :=> (A :+: B)): (B :<-- C) :=> A
  def uncocurry[A, B, C](g: (B :<-- C) :=> A): C :=> (A :+: B)

  def unapplicationCompose[A, B, C](
    f: C :=> (A :+: B),
    g: (B :<-- C) :=> A
  )(implicit eqF: Equal[C :=> (A :+: B)], eqG: Equal[(B :<-- C) :=> A]): Boolean =
    ???
}
```
I'm guessing it could look something like this, but I have essentially no idea about the meaning of this or what could be the instances.

experimental/shared/src/main/scala/zio/prelude/experimental/EitherCompose.scala
adamgfraser: I actually think this cuts the other way. When looking for relevant type classes for something it is nice to have things grouped together. If we had `ComposeEither`, `ComposeBoth`, and so on it would be easy to know where you should look for type classes for "function like" things and that those are different from type classes for other things.

sideeffffect: :thinking: are you suggesting that we rename `AssociativeCompose` -> `ComposeAssociative` and `IdentityCompose` -> `ComposeIdentity`?

adamgfraser: That's at least an idea.

sideeffffect: I think it would be better to change the naming scheme in a dedicated PR
https://github.com/zio/zio-prelude/issues/479

sideeffffect:
There are two open questions:
 * how to express Co-Cartesian Closed Category
 * Change the naming in the `AssociativeCompose` hierarchy using a non-typical scheme

I think we can deal with both in dedicated tickets and so this PR is ready to be merged.

https://github.com/zio/zio-prelude/issues/480
https://github.com/zio/zio-prelude/issues/479

What do you guys think?

sideeffffect:
Ping :smile_cat:

adamgfraser (approved):

PR #377, opened 2020-11-15
Fix Publishing Of Website

adamgfraser:
It looks like the website is failing to publish because `unidoc` fails to compile. It appears there are some other issues where there can be spurious compilation fails with `unidoc`. I added type annotations where necessary to resolve the failures. Obviously these are ugly but they don't impact the user API so I think it is best to get this in so we can get the website publishing and start adding content to it.

With this change you can run the following commands to create the site:

```
sbt docs/docusaurusCreateSite
cd website
yarn sta
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,7 +173,7 @@ microsite: &microsite
   steps:
       - add_ssh_keys:
           fingerprints:
-            - "11:2c:fa:98:69:d8:a1:d5:76:2e:38:d0:91:39:85:ce"
+            - "ea:30:ee:48:f1:96:e0:60:fd:8f:53:32:79:d5:7d:29"
       - checkout
       - <<: *load_cache
       - <<: *install_nodejs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -341,7 +341,7 @@ workflows:
       - microsite:
           context: Website
           requires:
-            - microsite
+            - release
           filters:
             <<: *filter_tags
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,25 +70,25 @@ save_cache: &save_cache
           - "~/.cache"
           # - "~/website/node_modules"
 
-# install_yarn: &install_yarn
-#   - run:
-#       name: Install Yarn
-#       command: |
-#         curl -sL https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
-#         sudo bash -c 'echo "deb https://dl.yarnpkg.com/debian/ stable main" > /etc/apt/sources.list.d/yarn.list'
-#         sudo apt update && sudo apt install yarn -y
-#         yarn policies set-version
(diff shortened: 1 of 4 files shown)


sideeffffect (approved):

sideeffffect:
Btw, we're doing it the same way (with `unidoc` & co.) as it's done in ZIO, ZIO Query, etc, right?

adamgfraser:
@sideeffffect Yes. I think the issue is we are doing more "fancy stuff" than those libraries are. All of the issues are with type inference for partially applied types which doesn't really come up in those libraries do to the avoidance of higher kinded types.

sideeffffect:
Shouldn't we also add the step that generates the website to `ci/circleci: mdoc`, so that we don't introduce a regression in a future PR?

adamgfraser:
@sideeffffect Good idea. There is a `microsite` action in the `config.yml` file that is run in ZIO Core but was commented out here that does that. I will uncomment that and then once we get that working we can add it to the branch protection rules.

sideeffffect (approved):

sideeffffect:
Great job, thanks!

.circleci/config.yml
```diff
@@ -169,33 +169,33 @@ release: &release
...
-#             node -v
-#             ./sbt ++${SCALA_VERSION}! docs/docusaurusCreateSite
-#             ./sbt ++${SCALA_VERSION}! docs/docusaurusPublishGhpages
-#       - <<: *clean_cache
-#       - <<: *save_cache
+microsite: &microsite
+  steps:
+      - add_ssh_keys:
+          fingerprints:
+            - "11:2c:fa:98:69:d8:a1:d5:76:2e:38:d0:91:39:85:ce"
```
mijicd, line 176: Please verify this is the fingerprint set for prelude.

mijicd (commented):
One small question, other than that looks good to me!

mijicd (approved):

sideeffffect (approved):

sideeffffect:
Another thing that comes to mind is, shouldn't we also add a new CircleCI job dedicated only to _building_ the site (so only `docs/docusaurusCreateSite`), and this job will be a prerequisite of `release` job and required by GitHub to pass for merging PRs?
Should we do it in this PR, or should we merge it right now and I will open a new PR for that?

mijicd:
@sideeffffect please let's keep the CI layout unified across the projects.

sideeffffect:
@mijicd alright :ok_hand: 
Do you think ZIO projects in general could benefit from this additional `docs/docusaurusCreateSite` job? It seems we could accidentally merge a PR that will fail building the website, which would be unfortunate. Should we discuss this in a separate ticket/PR (since that would be more general discussion, not related to this specific PR)?

mijicd:
Let's move it to #devops-lair on discord, as this is a bit broader topic :).

sideeffffect:
Is there anything that blocks us from merging this PR?

Issue #386, opened 2020-11-18
Add repeat method to Identity


lemastero:
_Difficulty: Easy_

Add method repeat to Identity that will combine input n elements a.
If n is below 1 it should return `identity` element.
```scala
def repeat(a: A, n: Int): A = ???
```

shankarshastri:
@sideeffffect , I would like to give it a try for this hackathon

shankarshastri:
@sideeffffect , I was trying to go through Identity, should the implementation be inside the Identity trait?

```scala
trait Identity[A] extends Associative[A] {

  /**
   * The identity element.
   */
  def identity: A
  
  @tailrec
  final def repeat[A](a: A, n: Int): A = {
    if(n < 1) combine(a, identity)
    else repeat(combine(a, a), n - 1)
  }
}
```

sideeffffect:
@shankarshastri I like this! Especially the part where we reuse the code and thus guarantee consistent behavior of `repeatOption` and `repeat` :+1: 
But there's a catch, if my eyes aren't failing me :smiley: 
The provided implementation of `repeatOption` repreduces the `a` exponentially. If we start with `"a"`, then we will have.
 * `"a"`
 * `"aa"`
 * `"aaaa"`
 * `"aaaaaaaa"`

I think the fix here would be to have a separate variable for the original value to reproduce and another variable for the accumulator.

But you can start your PR with this verbatim and then add tests and then we'll see, either on your machine locally or in CircleCI :wink:

Issue #387, opened 2020-11-19
Add repeatOption method to Associative


sideeffffect:
_Difficulty: Easy_

Add method repeat to `Associative` that will combine input `n` elements `a`.
If n is `<=` 1 it should return `None`.
```scala
def repeatOption(a: A, n: Int): Option[A] = ???
```

shankarshastri:
@sideeffffect , I would like to give it a try for this hackathon.

PR #396, opened 2020-11-20
Adding repeat and repeatOption to Identity And Associative

shankarshastri:
Adding repeat and repeatOption to Identity And Associative

Fixes: #386 , #387


src/test/scala/zio/prelude/CoherentSpec.scala
```diff
@@ -16,15 +17,17 @@ object CoherentSpec extends DefaultRunnableSpec {
       },
       test("IdentityEqual") {
         val instance = implicitly[EqualIdentity[String]]
-
         assert(instance.identity)(isEmptyString) &&
         assert(instance.combine("a", "b"))(equalTo("ab")) &&
+        assert(instance.repeat("a", 2))(equalTo("aa")) &&
```
sideeffffect, line 22: Instead of 2, could we test with a higher number, like 5, so that we defend against the bug we've encountered earlier?

sideeffffect (dismissed):
Nice job @shankarshastri. I think this looks good.
@adamgfraser what do you think?

src/main/scala/zio/prelude/Associative.scala
```diff
@@ -31,6 +33,15 @@ import zio.{ Chunk, NonEmptyChunk }
  */
 trait Associative[A] {
   def combine(l: => A, r: => A): A
+
+  final def repeatOption(a: A, n: Int): Option[A] = {
```
adamgfraser, line 37: Should the semantics here just be that if you give us a number less than or equal to one we return the original value unchanged? That way we wouldn't need the option.

sideeffffect: Maybe it would be best to start with the behavior of `repeat`. I would like the following property:

Assuming `n` and `m` are natural numbers,
```
repeat(n) <> repeat(m)  ===  repeat(n + m)
```

I don't think that property would hold, if `repeat(a, 0)` === `a`.

And `repeatOption` should behave analogously.

What do you think about it? :thinking:

lemastero: We do not have support for positive numbers. If we have, we could:
`repeat(Positive(n)) <> repeat(Positive(m))  ===  repeat(Positive(n) + Positive(m))`
Currently we could have 
`repeat(max(1,n)) <> repeat(max(1,m))  ===  repeat(max(1,n) + max(1,m))`
to keep semantics that @adamgfraser proposed. Also since that was my idea on the start. So maybe this is more intuitive even though law is a bit more complex.

I do not mind sorting this out in next PR :)

sideeffffect: I don't have strong feelings about this either way. But if we take into consideration not Associative, not Identity, but `Inverse` specifically, we wouldn't need to restrict ourselves to Naturals. We could generalize this even further.

For any Integer, this holds (or could hold):
```
repeat(n) <> repeat(m)  ===  repeat(n + m)
```
where `repeat(a, 0)` === `id` and `repeat(a, -2)` === `inverse(inverse(id, a), a)`, etc...

If we make a step back, what we're trying to do here with `repeat` is multiplication.
And only `Inverse` elements can be "multiplied" by a negative factor.
`Identity` elements cannot do that, but can at least be multiplied by 0.
`Associative` elements cannot do even that and have to be multiplied only by > 0.

If we wanted to keep with such semantics, we would need to have `repeatOption` on both `Associative` and `Identity` and `repeat` only on `Inverse`.

Do you think that would make sense?

lemastero (approved):

src/main/scala/zio/prelude/Associative.scala
lemastero: I like the idea, but I would love to see this in separate PR as this is going into the design not ment for beginner PR.

sideeffffect (approved):

sideeffffect:
@shankarshastri could you please fix the merge conflicts?

src/main/scala/zio/prelude/Associative.scala
adamgfraser: I guess I am coming at it from a little bit of a different perspective. I am thinking about methods like `repeat` operators on ZIO where we have for example:

```scala
  final def repeatN(n: Int): ZIO[R, E, A] =
    self.flatMap(a => if (n <= 0) ZIO.succeedNow(a) else ZIO.yieldNow *> repeatN(n - 1))
```

So this is basically saying if you ask us to repeat the effect zero or fewer times we are just going to ignore you and give you back the original value you gave us. I also see this as a case where calling this operator is almost a case of caller error, like doing `foreachPar` with a negative number, and something that if we were in the effectual world would warrant a `Die` cause. So returning a pretty logical value back seems like a very benign response to that. I also think having the `Option` in the return type creates significant user friction in working with this operator if they want to actually do something else with the value.

If the user really cares about guarding for the negative case they can always check the value before calling `repeat` and handle it however they wish.

sideeffffect: OK, I think for `repeat` (or should we rename it to `repeatN`?) we should implement it as in ZIO, as @adamgfraser says.
We'll deal with `multiply` in another ticket.
Sounds good?

adamgfraser: Sounds good!

sideeffffect: Here's the new ticket https://github.com/zio/zio-prelude/issues/417

sideeffffect (approved):

sideeffffect (changes requested):
@shankarshastri sorry for the changes in the assignment during PR review. Do you think you understand the design @adamgfraser proposed?

shankarshastri:
@sideeffffect , Sorry for the late reply Indian time zone :) . So according to @adamgfraser , we want to keep the implementation similar to repeatN from ZIO right? And then I see another ticket using which we can deal for multiply.
Correct me if I'm wrong

adamgfraser:
@shankarshastri No worries! Yes, I think we just want to implement `repeat` on `Associative` with this signature:

```scala
final def repeat(a: A, n: Int): A
```

I think we might want to curry it so it:

```scala
final def repeat(n: Int)(a: A): A
```

Then we can cover everything else in a subsequent PR once we have it scoped out.

shankarshastri:
@adamgfraser , @sideeffffect , @lemastero Please review !!

adamgfraser (approved):
Looks good!

sideeffffect (approved):
Nice!

Issue #375, opened 2020-11-14
Add methods to ZPure inspired by ZIO


sideeffffect:

arnoldlacko:
I would like to contribute to this on Saturday. I would start with `some`, `someOrElse`, `someOrElseM`, `someOrFail` and `someOrFailException`.

bmarinovic:
Hi, I would like to take:
* tap
* tapBoth
* tapCause
* tapError
* tapErrorTrace

(at least those that make sense to ZPure)

bmarinovic:
If `tap` proves to be useless in this context, I'd like to take:
* orElseFail
* orElseOptional
* orElseSucceed

atrianac:
I can take `right`, `rightOrFail`, `rightOrFailException`, `rightOrFailWith`

zach-albia:
I would like to take the `left` family methods.

bmarinovic:
I took:

*  fromEither[S, L, R](either: Either[L, R]): ZPure[S, S, Any, L, R]
*  fromOption[S, A](option: Option[A]): ZPure[S, S, Any, Unit, A]
*  suspend[S1, S2, R, E, A](pure: => ZPure[S1, S2, R, E, A]): ZPure[S1, S2, R, E, A]
*  head

landlockedsurfer:
Hi!
I work(ed) on `reject` and `rejectM`.

DamianReeves:
Do we have a list of what's left here?

bmarinovic:
#### For someone compiling the list of leftovers, this is what I implemented that was pushed to master:

* fromEither[S, L, R](either: Either[L, R]): ZPure[S, S, Any, L, R]
* fromOption[S, A](option: Option[A]): ZPure[S, S, Any, Unit, A]
* suspend[S1, S2, R, E, A](pure: => ZPure[S1, S2, R, E, A]): ZPure[S1, S2, R, E, A]
* fromEffect[S, A](effect: => A): ZPure[S, S, Any, Throwable, A]
* asState
* orElseFail
* orElseOptional
* orElseSucceed (althought better implementation is hanging in PR, waiting for someone to find type inference that will work with Dotty)

#### I have implementation that I'm working on:
* head

#### These were decided not to be implemented:

*  tap
* tapBoth
* tapCause
* tapError
* tapErrorTrace

arnoldlacko:
[...]

PR #405, opened 2020-11-21
ZIO inspired ZPure methods - orElse* family (#375)

bmarinovic:
#375


sideeffffect (approved):
Terrific work :+1: Thanks so much for this @bmarinovic :pray:

Issue #401, opened 2020-11-21
Use CanFail and NeedsEnv For ZPure


adamgfraser:
In ZIO we created `CanFail` and `NeedsEnv` data types that can be added as implicit parameters to error handling or environment providing methods to trigger a warning to the user if they are user an error handling operator on an effect that can't fail or providing an environment to an effect that doesn't need one. We should add the same thing to `ZPure` operators that handle errors or provide the environment.

spoltier:
I'd like to give this one a try.

adamgfraser:
@spoltier Great!

spoltier:
I have a PR open (some checks still running): #406

PR #406, opened 2020-11-21
Use CanFail and NeedsEnv For ZPure (#401)

spoltier:
Let me know if this makes sense, I may have forgotten some method.

Happy to do another pass later on sunday / monday once more ZPure methods (from issue #375) are there.

### DIFF
--- a/src/main/scala/zio/prelude/fx/ZPure.scala
+++ b/src/main/scala/zio/prelude/fx/ZPure.scala
@@ -279,7 +279,7 @@ sealed trait ZPure[-S1, +S2, -R, +E, +A] { self =>
    * Executes this computation and returns its value, if it succeeds, but
    * otherwise fails with the specified error.
    */
-  final def orElseFail[E1](e1: => E1): ZPure[S1, S2, R, E1, A] =
+  final def orElseFail[E1](e1: => E1)(implicit ev: CanFail[E]): ZPure[S1, S2, R, E1, A] =
     orElse(fail(e1))
 
   /**
@@ -296,14 +296,16 @@ sealed trait ZPure[-S1, +S2, -R, +E, +A] { self =>
    * Executes this computation and returns its value, if it succeeds, but
    * otherwise succeeds with the specified value.
    */
-  final def orElseSucceed[A1 >: A](a1: => A1): ZPure[S1, Any, R, Nothing, A1] =
+  final def orElseSucceed[A1 >: A](a1: => A1)(implicit ev: CanFail[E]): ZPure[S1, Any, R, Nothing, A1] =
     orElse(succeed(a1))
 
   /**
    * Executes this computation and returns its value, if it succeeds, but
    * otherwise fallbacks to the new state with the specified value.
    */
-  final def orElseFallback[A1 >: A, S3 >: S2](a1: => A1, s3: => S3): ZPure[S1, S3, R, Nothing, A1] =
+  final def orElseFallback[A1 >: A, S3 >: S2](a1: => A1, s3: => S3)(implicit
+    ev: CanFail[E]
+  ): ZPure[S1, S3, R, Nothing, A1] =
     orElse(succeed(a1).mapState(_ => s3))
 
   /**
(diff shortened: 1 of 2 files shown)


src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -91,13 +92,13 @@ sealed trait ZPure[-S1, +S2, -R, +E, +A] { self =>
     self andThen that
 
   /**
-   * Runs this computatation if the provided environment is a `Left` or else
+   * Runs this computation if the provided environment is a `Left` or else
```
sideeffffect, line 95: :smile_cat:

src/test/scala/zio/prelude/fx/ZPureSpec.scala
```diff
@@ -187,7 +188,7 @@ object ZPureSpec extends DefaultRunnableSpec {
           },
           testM("orElseSucceed (Success case)") {
             check(genInt, genInt, genInt) { (s1, v, v1) =>
-              val (_, a) = ZPure.succeed(v).orElseSucceed(v1).run(s1)
+              val (_, a) = ZPure.succeed(v).orElseSucceed(v1)(CanFail).run(s1)
```
sideeffffect, line 191: I'm not sure we should specify this explicitly. On the opposite, shouldn't this test verify, that when actual users will use this in their code, they won't have to specify it either?

sideeffffect (commented):
Thanks for this PR @spoltier
I have a question regarding the explicit use of `CanFail`.
Unfortunately I'm no ZPure expert, so we will need to rely on @adamgfraser 's expertise.

src/test/scala/zio/prelude/fx/ZPureSpec.scala
adamgfraser: The user shouldn't ever have to specify it. For testing in some cases we may need ignore the `CanFail` warnings. For example here I think the intent is to test that when you call `orElseSucceed` on a successful value you just get the original success back, which makes sense but would normally cause an error.

A slightly nicer way to do this is to do the following within the scope of your test:

```scala
implicit val canFail = CanFail
```

This will silence the warnings for that scope.

spoltier: @adamgfraser Ok, I'll switch to the implicit declaration

src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -35,13 +36,13 @@ sealed trait ZPure[-S1, +S2, -R, +E, +A] { self =>
...
    * runs that computation if the provided environment is a `Right`, returning
    * the result in an `Either`.
    */
   final def +++[S0 <: S1, S3 >: S2, R1, B, E1 >: E](
     that: ZPure[S0, S3, R1, E1, B]
-  ): ZPure[S0, S3, Either[R, R1], E1, Either[A, B]] =
+  )(implicit evL: NeedsEnv[R], evR: NeedsEnv[R1]): ZPure[S0, S3, Either[R, R1], E1, Either[A, B]] =
```
adamgfraser, line 45: Don't think this is correct. Even if the environment type is any this is well defined. Just if we provide `Left` with anything we get one value and if we provide `Right` with anything we get the other value.

src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -91,13 +92,13 @@ sealed trait ZPure[-S1, +S2, -R, +E, +A] { self =>
...
   /**
-   * Runs this computatation if the provided environment is a `Left` or else
+   * Runs this computation if the provided environment is a `Left` or else
    * runs that computation if the provided environment is a `Right`, unifying
    * the result to a common supertype.
    */
   final def |||[S0 <: S1, S3 >: S2, R1, B, E1 >: E, A1 >: A](
     that: ZPure[S0, S3, R1, E1, A1]
-  ): ZPure[S0, S3, Either[R, R1], E1, A1] =
+  )(implicit evL: NeedsEnv[R], evR: NeedsEnv[R1]): ZPure[S0, S3, Either[R, R1], E1, A1] =
```
adamgfraser, line 101: Same comment as above regarding `+++`.

adamgfraser (approved):
Looks good! We have a separate PR in progress to improve the signature of `orElseSucceed`.

sideeffffect:
Cand we do somehting about `license/cla` staying yellow @adamgfraser ?

spoltier:
> Can we do something about `license/cla` staying yellow @adamgfraser ?

Looks related to cla-assistant/cla-assistant#124 - some CLA assistant panel accessible to github org members / admins ?

sideeffffect:
So closing and re-opening helped... you can tell, that GitHub is now operated by Microsoft :joy: 
Thanks again @spoltier good work :muscle:

PR #413, opened 2020-11-21
Better orElseSucceed method (#375)

bmarinovic:


sideeffffect:
Nice. The only thing remaining is to make Dotty with the code. Try this locally on your computer by changing `scalaVersion in ThisBuild` to `ScalaDotty` in `project/BuildHelper.scala` :wink:

sideeffffect:
@bmarinovic could you please fix the merge conflicts? :pray:

bmarinovic:
I will on a next commit when we find a better way to type inference to work on Dotty. Currently, it doesn't make sense to merge this.

sideeffffect:
Any luck moving this forward? :smiley:

bmarinovic:
It is stuck at Dotty's type inference problems.
Until we find concise way to do it, I believe no.
@adamgfraser can double check it

src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -460,6 +446,18 @@ sealed trait ZPure[-S1, +S2, -R, +E, +A] { self =>
...
 }
 
+trait ZPureSyntax {
+  final implicit class ZPureOps[S1, S2, R, E, A](private val self: ZPure[S1, S2, R, E, A]) {
+
+    /**
+     * Executes this computation and returns its value, if it succeeds, but
+     * otherwise succeeds with the specified value.
+     */
+    def orElseSucceed[A1 >: A](a1: => A1)(implicit ev: S1 <:< S2): ZPure[S1, S2, R, Nothing, A1] =
```
jdegoes, line 456: If you are going to require implicit evidence here that `S1 <:< S2`, then you may as well just have one type parameter `S` on the `ZPureOps`. It's the same thing only will infer better.

adamgfraser: Yes. That is the approach I took with `UnifiedSyntax` in #461.

jdegoes:
@bmarinovic I will close this for now. If you can fix conflicts and simplify the encoding a bit (which should get dotty compiling too, hopefully), then we can merge it in right away. Thanks for your work on this!

Issue #371, opened 2020-11-14
Add method bothPar to the type class CommutativeBoth


sideeffffect:
_Difficulty: Easy, familiarity with the Type class design pattern is an advantage_

After adding `bothPar` to `CommutativeBoth`, implement it for all the types implementing the Type class. The signature is like this:

```scala
  /**
   * Combines two values of types `F[A]` and `F[B]` in a parallel manner to produce an `F[(A, B)]`.
   */
  def bothPar[A, B](fa: => F[A], fb: => F[B]): F[(A, B)]
```

Make sure, that the other helper and extension methods (like `zipPar`) in `CommutativeBoth` use `bothPar` instead of just `both`. Note, that for some types, `bothPar` can be the same implementation as `both`.

As a stretch goal, move the `CommutativeBoth` instances (all that can be, which are all in this case) to the supre-classes' `AssociativeBoth` companion object. Read about the reasoning where to put Type class instance and the _Implicit scope_ in this article https://meta.plasm.us/posts/2019/09/30/implicit-scope-and-cats/

As a stretch goal # 2, add tests for missing instances to `CommutativeBothSpec` (tests for some basic instances, like `Option` or `List` are already present).

visar:
I would like to tackle this.

sideeffffect:
closing as per https://github.com/zio/zio-prelude/issues/372#issuecomment-744067514

Issue #372, opened 2020-11-14
Add method eitherPar to the type class CommutativeEither


sideeffffect:
_Difficulty: Easy, familiarity with the Type class design pattern is an advantage_

After adding `eitherPar` to `CommutativeEither`, implement it for all the types implementing the Type class. The signature is like this.
```scala
  /**
   * Combines two values of types `F[A]` and `F[B]` in a parallel manner to produce an `F[Either[A, B]]`.
   */
  def eitherPar[A, B](fa: => F[A], fb: => F[B]): F[Either[A, B]]
```

Make sure, that the other helper and extension methods (like `orElseEitherPar`) in `CommutativeEither` use `eitherPar` instead of just `either`. Note, that for some types, `eitherPar` can be the same implementation as `either`.

As a stretch goal, move the `CommutativeEither` instances (all that can be, which are all in this case) to the supre-classes' `AssociativeEither` companion object. Read about the reasoning where to put Type class instance and the _Implicit scope_ in this article https://meta.plasm.us/posts/2019/09/30/implicit-scope-and-cats/

As a stretch goal # 2, add tests for missing instances to `CommutativeEitherSpec` (tests for some basic instances, like `Option` or `List` are already present).

lbdlbdlbdl:
I would like to work on this one. :)

sideeffffect:
By this ticket, we would have an `CommutativeEither` with both `either` and `eitherPar`, where `either` would express the sequential merging while `eitherPar` would express the "parallel" one (concurrent would be more apt, though :D ).

@adamgfraser in we merged https://github.com/zio/zio-prelude/pull/437, we would loose the ability to merge certain types in a sequential manner (like ZIO, Future, etc), which would be unfortunate.

But if you really wanted `either` on `CommutativeEither` to be the parallel/concurrent variant, we could "do it the other way around" and instead of adding `eitherPar`, we could have an `eitherSeq` member on `CommutativeEither` which would still enable use to the sequential way.

Or we could drop `CommutativeEither` for mere ZIOs and Futures etc. and introduce `Sequential` and `Concurrent` subtypes, where we could implement different strategies for `Sequential[ZIO]` and `Concurrent[ZIO]` and so on for the other. Similarly to the trick how we have both additive and multiplicative Magmas for integers, i.e. `Sum[Int]` and `Prod[Int]`

So I see 4 options:

 * `0`: loose the ability to merge ZIO/Future/... sequentially, but nobody wants that, I guess
[...]

PR #421, opened 2020-11-22
Add method bothPar to the type class CommutativeBoth

visar:
#371

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/CommutativeBoth.scala
+++ b/core/shared/src/main/scala/zio/prelude/CommutativeBoth.scala
@@ -56,17 +56,17 @@ object CommutativeBoth extends LawfulF.Invariant[CommutativeBothDeriveEqualInvar
     new CommutativeBoth[Chunk] {
       def both[A, B](fa: => Chunk[A], fb: => Chunk[B]): Chunk[(A, B)] = fa zip fb
 
-      def bothPar[A, B](fa: => Chunk[A], fb: => Chunk[B]): Chunk[(A, B)] = fa zip fb
+      def bothPar[A, B](fa: => Chunk[A], fb: => Chunk[B]): Chunk[(A, B)] = both(fa, fb)
     }
 
   /**
    * The `CommutativeBoth` instance for `Exit`.
    */
-  implicit def ExitCommutativeBoth[E]: CommutativeBoth[Exit[E, +*]] =
-    new CommutativeBoth[Exit[E, +*]] {
+  implicit def ExitCommutativeBoth[E]: CommutativeBoth[({ type lambda[+a] = Exit[E, a] })#lambda] =
+    new CommutativeBoth[({ type lambda[+a] = Exit[E, a] })#lambda] {
       def both[A, B](fa: => Exit[E, A], fb: => Exit[E, B]): Exit[E, (A, B)] = fa zipPar fb
 
-      def bothPar[A, B](fa: => Exit[E, A], fb: => Exit[E, B]): Exit[E, (A, B)] = fa zipPar fb
+      def bothPar[A, B](fa: => Exit[E, A], fb: => Exit[E, B]): Exit[E, (A, B)] = both(fa, fb)
     }
 
   /**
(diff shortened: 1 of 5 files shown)


src/main/scala/zio/prelude/CommutativeBoth.scala
```diff
@@ -49,14 +55,18 @@ object CommutativeBoth extends LawfulF.Invariant[CommutativeBothDeriveEqualInvar
...
+      def bothPar[A, B](fa: => Chunk[A], fb: => Chunk[B]): Chunk[(A, B)] = fa zip fb
     }
 
   /**
    * The `CommutativeBoth` instance for `Exit`.
    */
-  implicit def ExitCommutativeBoth[E]: CommutativeBoth[({ type lambda[+a] = Exit[E, a] })#lambda] =
-    new CommutativeBoth[({ type lambda[+a] = Exit[E, a] })#lambda] {
+  implicit def ExitCommutativeBoth[E]: CommutativeBoth[Exit[E, +*]] =
+    new CommutativeBoth[Exit[E, +*]] {
```
sideeffffect, line 66: Unfortunately this won't work in Dotty, that's why we have to use the `lambda` trick, sad, I know :crying_cat_face: .
Or try it and then we'll see for sure.

src/main/scala/zio/prelude/CommutativeBoth.scala
```diff
@@ -49,14 +55,18 @@ object CommutativeBoth extends LawfulF.Invariant[CommutativeBothDeriveEqualInvar
   implicit def ChunkCommutativeBoth: CommutativeBoth[Chunk] =
     new CommutativeBoth[Chunk] {
       def both[A, B](fa: => Chunk[A], fb: => Chunk[B]): Chunk[(A, B)] = fa zip fb
+
+      def bothPar[A, B](fa: => Chunk[A], fb: => Chunk[B]): Chunk[(A, B)] = fa zip fb
```
sideeffffect, line 59: In cases like this, I'm thinking whether it wouldn't be for the best if we reused `both` method to emphasize that the implementation is literally the same, like
```scala
       def bothPar[A, B](fa: => Chunk[A], fb: => Chunk[B]): Chunk[(A, B)] = both(fa, fb) 
```

src/main/scala/zio/prelude/CommutativeBoth.scala
```diff
@@ -94,29 +111,44 @@ object CommutativeBoth extends LawfulF.Invariant[CommutativeBothDeriveEqualInvar
...
+        }
     }
 
   /**
    * The `CommutativeBoth` instance for And `Schedule`.
    */
-  implicit def ScheduleAndCommutativeBoth[R, E]
-    : CommutativeBoth[({ type lambda[+a] = AndF[Schedule[R, E, a]] })#lambda] =
-    new CommutativeBoth[({ type lambda[+a] = AndF[Schedule[R, E, a]] })#lambda] {
+  implicit def ScheduleAndCommutativeBoth[R, E]: CommutativeBoth[Lambda[`+a` => AndF[Schedule[R, E, a]]]] =
```
sideeffffect, line 125: I'm afraid that if the `+a` is in between the back-ticks, it won't be taken as covariant -- are you sure this will work?
It's probably for the best, if we stick to the `lambda` trick, because it works across all Scala versions.

sideeffffect (changes requested):
Hello @visar, thanks for working on this thing!
I have a few comment and suggestions, but we're on the right track :rocket:

sideeffffect (approved):
This is great! Thank you.

sideeffffect:
As a stretch goal, could you please move the instances that can be moved (which I think are all of them) from the `CommutativeBoth` companion object to the `AssociativeEither` :pray: 
They should have been there from the start, so it would be great if you could help us clean this up.

If you'd like to understand the theoretical reasons for why such instances need to be in the "most parent"'s companion object, have a look at this article: https://meta.plasm.us/posts/2019/09/30/implicit-scope-and-cats/

sideeffffect:
For the record, merging this kinda hinges on how we resolve this question https://github.com/zio/zio-prelude/issues/372#issuecomment-734500103
/cc @adamgfraser

sideeffffect:
As per https://github.com/zio/zio-prelude/issues/372#issuecomment-744067514, we will not be merging this, we decided to take a different path.
Thank you @visar for the nice work you did on this, I'm sorry that it will not get into `master`, but if you would still like to help us with ZIO Prelude, please check our issue tracker, there's still a lot of interesting work to do, especially the [`#ziohackathon`](https://github.com/zio/zio-prelude/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%23ziohackathon) label

Issue #370, opened 2020-11-13
Port ZIO Prelude to Scala Native


sideeffffect:
_Difficulty: Intermediate, prior experience with (porting to) Scala Native is an advantage_

Add sbt setup for building, testing and publishing ZIO Prelude for Scala Native `0.4.0-M2` for Scala 2.11.
Have a look at ZIO's https://github.com/zio/zio/blob/master/build.sbt#L133 (`coreNative`, for example) and https://github.com/zio/zio/blob/master/.circleci/config.yml to see how it's done. (Other files of interest are `project/BuildHelper.scala` and `project/plugins.sbt`, but the necessary changes to the should be already in place.)
You can also look at some of the PRs that added Scala Native support to ZIO: https://github.com/zio/zio/pull/4307 , https://github.com/zio/zio/pull/4252 , https://github.com/zio/zio/pull/4260

Once the PR is merged, we should see `zio-prelude_native0.4.0-M2_2.11` at
https://oss.sonatype.org/content/repositories/snapshots/dev/zio/

vigoo:
Seems like `zio_native0.4.0-M2_2.11` depends on `zio-stacktracer_native0.4.0-M2_2.11` (https://search.maven.org/artifact/dev.zio/zio_native0.4.0-M2_2.11/1.0.3/jar) and it's not published at all. So first this dependency should be published from `zio`.

sideeffffect:
Yeah, stacktraces is only published as a snapshot
https://oss.sonatype.org/content/repositories/snapshots/dev/zio/zio-stacktracer_native0.4.0-M2_2.11/

That could help you at least with making the chages to the build setting -- even though we will most likely need to wait for ZIO 1.0.4 to merge your pull request. But having such a prepared PR would still help us immensely, basically all the work will be done :bow:

PR #423, opened 2020-11-22
Initial support for scala-native

vigoo:
Adds scala-native builds (https://github.com/zio/zio-prelude/issues/370), only works with latest snapshot because for `1.0.3` the `zio-stacktracer` dependency was not published.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,7 +18,7 @@ inThisBuild(
     pgpSecretRing := file("/tmp/secret.asc"),
     scmInfo := Some(
       ScmInfo(url("https://github.com/zio/zio-prelude/"), "scm:git:git@github.com:zio/zio-prelude.git")
-    ),
+    )
   )
 )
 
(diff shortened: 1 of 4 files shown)


vigoo:
Also there seems to be no sbt test runner for native yet so the CircleCI job only runs `compileNative`

sideeffffect (approved):
This is sooo awesome! Thank you @vigoo for this. We've got Scala.js yesterday and now Native too. I hope many Scala Native devs will find it useful :tada:

build.sbt
```diff
@@ -73,11 +73,30 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
...
 
-lazy val coreJS = core.js
+lazy val coreJS     = core.js
   .settings(jsSettings)
+  .settings(libraryDependencies += "dev.zio" %%% "zio-test-sbt" % zioVersion)
 
-lazy val coreJVM = core.jvm
+lazy val coreJVM    = core.jvm
   .settings(dottySettings)
+  .settings(libraryDependencies += "dev.zio" %%% "zio-test-sbt" % zioVersion)
```
sideeffffect, line 82: Is it supposed to have 3 `%%%` even for JVM?

sideeffffect:
> no sbt test runner
Yes. But actually running tests doesn't work anyway, due to limitations in Scala Native 0.4.0-M2. But we could at least start publishing the artifact.

Do you think you could (we could call it a stretch goal) amend the configuration in ZIO itself, so that it publishes `zio-test-sbt` for Scala Native too?

build.sbt
vigoo: No I think it can be `%%`, just copied it from the shared place. Although I'm not completely aware of how `CrossProject` modifies the behavior of `libraryDependencies`.

vigoo:
I will check why `zio-test-sbt` is not available for native

PR #425, opened 2020-11-22
Add zpure companion obj methods

bmarinovic:
#375

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
+++ b/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
@@ -2,6 +2,7 @@ package zio.prelude.fx
 
 import scala.annotation.switch
 import scala.util.Try
+import scala.util.control.NonFatal
 
 import zio.internal.Stack
 import zio.prelude._
@@ -591,8 +592,8 @@ object ZPure {
     suspend {
       try ZPure.succeed(effect)
       catch {
-        case t: VirtualMachineError => throw t
-        case t: Throwable           => ZPure.fail(t)
+        case NonFatal(e)  => ZPure.fail(e)
+        case t: Throwable => throw t
       }
     }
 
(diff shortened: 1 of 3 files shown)


sideeffffect (approved):
Great!

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -544,12 +545,48 @@ object ZPure {
...
     fromFunction(_._1)
 
+  /**
+   * Constructs a computation from an effect that may throw.
+   */
+  def fromEffect[S, A](effect: => A): ZPure[S, S, Any, Throwable, A] =
+    suspend {
+      try ZPure.succeed(effect)
+      catch {
+        case t: VirtualMachineError => throw t
```
sideeffffect, line 555: I'd consider using `NonFatal` here

sideeffffect (changes requested):
After we use `NonFatal` in catch, I think we're good to go :+1:

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -583,12 +585,48 @@ object ZPure {
...
 
+  /**
+   * Constructs a computation from an effect that may throw.
+   */
+  def fromEffect[S, A](effect: => A): ZPure[S, S, Any, Throwable, A] =
+    suspend {
+      try ZPure.succeed(effect)
+      catch {
+        case NonFatal(e)  => ZPure.fail(e)
+        case t: Throwable => throw t
```
sideeffffect, line 596: this line doesn't even need to be here :wink:

sideeffffect (approved):
Great job, @bmarinovic . Thank you for your work and your patience with this. It's been a long journey, but we're there :tada:

sideeffffect (approved):

sideeffffect (approved):
It's happening!!! :tada: We're merging.
Thank you for your perseverance through this ordeal @bmarinovic

PR #430, opened 2020-11-22
Use anyInt instead of anyString as much as possible, JavaScript seems slow with it

sideeffffect:
Interesting thing is, that it doesn't seem to be as big of a problem with Scala 2.11, only on 2.12 and 2.13... :thinking: 
This has serious consequences, builds not take 10 mins instead of 30 mins and don't timeout in CircleCI on _no output for 10 mins_.


jdegoes (approved):

jdegoes:
🙏

sideeffffect:
Thank you John! It might still be worth investigating _why_ has `anyString` caused such a performance regression -- only on Scala.js and more so on Scala 2.12 and 2.13 and less so on 2.11.
So far I was wondering if it's caused by the fact that HotSpot is reusing identical strings internally and so it might not create such a pressure on GC, but that's a wild guess...

PR #434, opened 2020-11-24
Add Experimental Module

### LINKED ISSUE #412: Add submodule zio-prelude-ext
Difficulty: easy

Idea is to have a place where one can put abstractions that will be changed later on or can be throw away if they don't pass the test of time.

adamgfraser:
Resolves #412.

Adds `experimental` module under namespace `zio.prelude.experimental`. We have a lot of really interesting work going on regarding zivariant, semiring, and lattice structures among others. I think we are being hampered a bit because these things are pushing the boundaries in various directions so we don't want to merge them but that also means it is harder for people to try them out or do further work on top of them. This way we can be more liberal in getting things into the experimental package and then if we feel good about them or get traction we can potentially move to the main package.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -122,7 +122,7 @@ lazy val experimentalNative = core.native
   .settings(crossScalaVersions := Seq(scalaVersion.value))
   .settings(skip in Test := true)
   .settings(skip in doc := true)
-  .settings(       // Exclude from Intellij because Scala Native projects break it - https://github.com/scala-native/scala-native/issues/1007#issuecomment-370402092
+  .settings( // Exclude from Intellij because Scala Native projects break it - https://github.com/scala-native/scala-native/issues/1007#issuecomment-370402092
     SettingKey[Boolean]("ide-skip-project") := true
   )
   .settings(sources in (Compile, doc) := Seq.empty)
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -102,6 +102,35 @@ lazy val coreNative = core.native
...
   )
 
+lazy val experimental = crossProject(JSPlatform, JVMPlatform, NativePlatform)
+  .in(file("experimental"))
+  .dependsOn(core)
+  .settings(stdSettings("zio-prelude-experimental"))
+  .settings(crossProjectSettings)
+  .settings(buildInfoSettings("zio.prelude.experimental"))
+
+lazy val experimentalJVM    = core.jvm
```
sideeffffect, line 112: I believe `experimentalJVM`, `experimentalJS` and `experimentalNative` should be aggregated in `root`.
Also, `experimentalJVM/test` should be added to the `testJVM` alias and analogously for `*JS` and `*Native`.

sideeffffect (changes requested):
Thank you Adam! Only small details to wire it all up :pray:

build.sbt
adamgfraser: @sideeffffect Fixed now!

adamgfraser:
Thanks! Fixed now.

sideeffffect (approved):
Great, this should be good to go :+1:

PR #435, opened 2020-11-24
make environment in ZPure scope correctly

felher:
Currently, `provide` on `ZPure` behaves more like `setState`. I.e., providing some value changes the environment for younger siblings and not only for children. Here is an example:

```scala
val zpure = for {
 _ <- doSomething.provide(1)
 i <- ZPure.environment[Any, Int]
} yield i

zpure.provide(0).runResult(())
```

The code above will return 1, not 0.

This PR provides a test for that and tries to fix it.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
+++ b/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
@@ -414,6 +414,8 @@ sealed trait ZPure[-S1, +S2, -R, +E, +A] { self =>
             unwinding = false
           case null                                =>
             unwinding = false
+          case value: ProvideEnd[_, _, _, _, _]    =>
+            r = value.r
           case _                                   =>
         }
     }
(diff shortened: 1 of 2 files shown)


jdegoes:
@felher That is a problem. Internally the interpreter for ZPure needs to use a stack of environment values, and push / pop.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -460,18 +460,25 @@ sealed trait ZPure[-S1, +S2, -R, +E, +A] { self =>
...
           stack.push(zPure)
-        case Tags.Access  =>
+        case Tags.Access     =>
           val zPure = curZPure.asInstanceOf[Access[Any, Any, Any, Any, Any]]
           curZPure = zPure.access(r)
-        case Tags.Provide =>
+        case Tags.Provide    =>
           val zPure = curZPure.asInstanceOf[Provide[Any, Any, Any, Any, Any]]
+          val old   = r
+          stack.push(_ => ZPure.ProvideEnd(old))
```
jdegoes, line 473: I think this should work correctly. 👍

adamgfraser (approved):
Good catch!

PR #436, opened 2020-11-24
Simplify Algebra of ZPure

adamgfraser:
We don't need a new term in the algebra for popping the environment from the stack.


sideeffffect (approved):
:+1: from what I can tell

sideeffffect:
@adamgfraser I think we can merge this. Or do you plan to do further work on this PR?

Issue #373, opened 2020-11-14
Define laws for Traversable


sideeffffect:

zach-albia:
I'd like to take this.

zach-albia:
I'm hitting a roadblock writing the naturality law for `Traversable`. Looking at the implementation of the laws in Cats & ScalaZ, it looks like we need to have an implementation of natural transformation. Do we need to implement it in zio-prelude to write this law or am I missing something?

sideeffffect:
I'm no category theorist, so please excuse me if I say something nonsensical.

Natural transformations are morphisms between functors, right? In that case we don't have them in ZIO Prelude (yet :wink:).

But let me ask "from the other side". Are we able to write some interesting laws/properties for Traversable with the current state of things? If yes, could you please open a PR with that? We'll figure something out about _naturality_ in the meantime :+1:

Badmoonz:
@zach-albia I think we don't need so Generic approach for generating Applicatives used by foreach, testing for several hardcoded instances is sufficient, event using only Id Applicative is enough imho

Badmoonz:
@zach-albia 

I think it will be nice to add some tests to check that laws are really working:
```
        {
          implicit val badTraversable = new Traversable[Option] {
            // breaks broken Traversable Identity law
            override def foreach[G[+_]: IdentityBoth: Covariant, A, B](fa: Option[A])(f: A => G[B]): G[Option[B]] =
              (None: Option[B]).succeed
            override def map[A, B](f: A => B): Option[A] => Option[B] = _.map(f)
          }

          testM("Traversable.Identity law violation captured")(checkAllLaws(Traversable)(GenF.option, Gen.anyInt).map(x => assert(x.isFailure)(Assertion.isTrue)))
        },
        {
          implicit val badTraversable = new Traversable[Option] {
            override def foreach[G[+_]: IdentityBoth: Covariant, A, B](fa: Option[A])(f: A => G[B]): G[Option[B]] = fa match {
              case Some(x) => f(x).map(Some(_): Option[B])
              case None    => (None: Option[B]).succeed
            }

            // breaks Covariant Identity law
            override def map[A, B](f: A => B): Option[A] => Option[B] = _ => None

          }
          testM("Covariant.Identity law violation captured")(checkAllLaws(Traversable)(GenF.option, Gen.anyInt).map(x => assert(x.isFailure)(Assertion.isTrue)))
        }

```

[...]

PR #442, opened 2020-11-25
Define laws for Traversable

zalbia:
#373


core/shared/src/main/scala/zio/prelude/Traversable.scala
```diff
@@ -269,11 +270,25 @@ trait Traversable[F[+_]] extends Covariant[F] {
 
 object Traversable extends LawfulF.Covariant[DeriveEqualTraversable, Equal] {
 
+// need to implement natural transformations for this property?
+// val naturalityLaw: LawsF.Covariant[DeriveEqualTraversable, Equal] = ???
+
+  /**
+   * Traversing by `Id` is equivalent to mapping.
+   */
+  val identityLaw: LawsF.Covariant[DeriveEqualTraversable, Equal] = {
```
adamgfraser, line 279: You can implement your own law until we add one to ZIO Test.

core/shared/src/main/scala/zio/prelude/Traversable.scala
```diff
@@ -269,11 +270,25 @@ trait Traversable[F[+_]] extends Covariant[F] {
 
 object Traversable extends LawfulF.Covariant[DeriveEqualTraversable, Equal] {
 
+// need to implement natural transformations for this property?
+// val naturalityLaw: LawsF.Covariant[DeriveEqualTraversable, Equal] = ???
```
adamgfraser, line 274: I think we can restate this as a fusion law that if we have `F[G[H[A]]]` where `F` is traversable and `G` and `H` have `IdentityBoth` and `Covariant` then we can individually `flip` twice or we can create a composed instance and `flip` the whole thing once.

zach-albia: Thanks for the big tip!

zach-albia:
@adamgfraser can you check if I'm going in the right direction so far? AFAICT, there's still lots of work to be done, like making it so the `Traversable` laws run with `Covariant`'s, as well as adding some more instances to check maybe?

core/shared/src/main/scala/zio/prelude/newtypes/package.scala
```diff
@@ -91,4 +91,30 @@ package object newtypes {
...
   type FailureOut[+A] = FailureOut.Type[A]
+
+  /**
+   * A newtype representing Right-to-left composition of functors.
+   *
+   * If F[_] and G[_] are both Covariant, then Nested[F, G, *] is also a Covariant.
+   *
+   * If F[_] and G[_] are both IdentityBoth, then Nested[F, G, *] is also an IdentityBoth.
+   *
+   * If F[_] and G[_] are both Traversable, then Nested[F, G, *] is also a Traversable.
```
Badmoonz, line 102: I think it would be nice to add tests for that at CovariantSpec, IdentityBothSpec, TraversableSpec

can copy it from my branch:
https://github.com/Badmoonz/zio-prelude/commit/6df1301ec35870736f92a15e4c4f1f58d51a4fd1#diff-82e9bed52abfcc1cabd8fb86c811dcab611fbd08d2ff81124a90af2fb0a78834R21

zach-albia: So I tried to add the ones for `CovariantSpec` & `IdentityBothSpec`, and they led to the same build errors for 2.11 in #449. Any idea on how to make them work?

Badmoonz: After some research , I've found solution : 

```{
        type T[+A] = Nested[Vector, Option]
        testM("Nested[vector,option]")(checkAllLaws(Traversable)(GenFs.nested(GenF.vector, GenF.option) : GenF[T, Random with Sized] , Gen.anyInt))
       }
```

It looks like scala 2.11 compiler cannot derive type `CapF[_[_[_]]]` from first argument GenFs.nested(GenF.vector, GenF.option) which has type `GenF[(lambda[+A] = Nested[Vector, Option, A)#lambda, Random with Sized]`.
I supposed that's because it has lambda type param, but CovariantSpec has tuple, either gens that also return GenF with lambda type param, and it works well

jdegoes:
@adamgfraser @zalbia @sideeffffect This one ready to merge (aside from merge conflicts)???

sideeffffect:
@jdegoes it may turn out we may need just one law for Traversable https://github.com/zio/zio-prelude/issues/373#issuecomment-739153221

adamgfraser:
@Badmoonz @zalbia You have both done good work in this and I believe some of the work by @zalbia on this has been incorporated into the PR from @Badmoonz. How would you like to move forward to get this over the finish line?

zalbia:
We should merge #449 once the PR passes checks.

adamgfraser:
@zalbia 👍

Badmoonz:
ok, i will update und fix  #449  today

PR #445, opened 2020-11-26
new ZPure constructor 'modifyEither'

arnoldlacko:
I would like to propose this new `ZPure` constructor for constructing computations that may fail.


sideeffffect (approved):
Nice, thank you!

arnoldlacko:
> Nice, thank you!

Thanks for the review!

PR #452, opened 2020-12-02
Add Traversable.compose

joroKr21:
Also:
 - rename `composeInvariant` to just `compose`
 - make `compose` parameters implicit so that we can call `.compose[G]`

I didn't find any tests.

And btw, `Applicative` functors compose too, but by splitting them into `Covariant` and `IdentityBoth` we lose that ability 😞

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/Contravariant.scala
+++ b/core/shared/src/main/scala/zio/prelude/Contravariant.scala
@@ -46,7 +46,7 @@ trait Contravariant[F[-_]] extends ContravariantSubset[F, AnyType] with Invarian
   /**
    * Compose two contravariant functors.
    */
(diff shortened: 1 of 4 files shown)


adamgfraser (approved):
Looks good. I think we could do the same thing with `IdentityBoth` by requiring implicit evidence that a `Covariant` instance also exists.

joroKr21:
> I think we could do the same thing with IdentityBoth by requiring implicit evidence that a Covariant instance also exists.

The problem is that we have `F[_]` (invariant) in `IdentityBoth` and `F[+_]` (covariant) in `Convariant`

adamgfraser:
Moving to the companion object should resolve that.

joroKr21:
You mean like this? #454

PR #454, opened 2020-12-03
Add AssociativeBoth.compose

joroKr21:
To the companion object because we need a covariant F

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/AssociativeBoth.scala
+++ b/core/shared/src/main/scala/zio/prelude/AssociativeBoth.scala
@@ -65,6 +65,13 @@ object AssociativeBoth extends LawfulF.Invariant[AssociativeBothDeriveEqualInvar
       override def both[A, B](fa: => F[A], fb: => F[B]): F[(A, B)] = fa.map(a => fb.map(b => (a, b))).flatten
     }
 
+  def compose[F[+_]: AssociativeBoth: Covariant, G[_]: AssociativeBoth]
+    : AssociativeBoth[({ type lambda[A] = F[G[A]] })#lambda] =
+    new AssociativeBoth[({ type lambda[A] = F[G[A]] })#lambda] {
+      override def both[A, B](fa: => F[G[A]], fb: => F[G[B]]) =
+        fa.zipWith(fb)(_ zip _)
+    }
+
   /**
    * Combines 2 `F` values using the provided function `f`.
    */
--- a/core/shared/src/main/scala/zio/prelude/AssociativeBoth.scala
+++ b/core/shared/src/main/scala/zio/prelude/AssociativeBoth.scala
@@ -65,11 +65,13 @@ object AssociativeBoth extends LawfulF.Invariant[AssociativeBothDeriveEqualInvar
       override def both[A, B](fa: => F[A], fb: => F[B]): F[(A, B)] = fa.map(a => fb.map(b => (a, b))).flatten
     }
 
-  def compose[F[+_]: AssociativeBoth: Covariant, G[_]: AssociativeBoth]
-    : AssociativeBoth[({ type lambda[A] = F[G[A]] })#lambda] =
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/zio/prelude/AssociativeBoth.scala
```diff
@@ -65,6 +65,13 @@ object AssociativeBoth extends LawfulF.Invariant[AssociativeBothDeriveEqualInvar
       override def both[A, B](fa: => F[A], fb: => F[B]): F[(A, B)] = fa.map(a => fb.map(b => (a, b))).flatten
     }
 
+  def compose[F[+_]: AssociativeBoth: Covariant, G[_]: AssociativeBoth]
```
adamgfraser, line 68: I wonder if this should require that `G` by `Covariant` as well and then we can return a composed type that is both an `AssociativeBoth` and `Covariant`. They are both valid derivations but I wonder if that would be more useful. In particular that would give you a type that you could further compose if you wanted to work with even more deeply nested structures.

adamgfraser (approved):
Looks great. Just one question.

core/shared/src/main/scala/zio/prelude/AssociativeBoth.scala
joroKr21: I was wondering the same - I guess why not

joroKr21: Although it will not be possible to do `AssociativeBoth[Chunk].compose[List].compose[Option]`.
The intermediate `Covariant` will have to be implicit.

joroKr21: Done

sideeffffect (approved):
Nice!

core/shared/src/test/scala/zio/prelude/CovariantSpec.scala
```diff
@@ -18,7 +23,8 @@ object CovariantSpec extends DefaultRunnableSpec {
         testM("cause")(checkAllLaws(Covariant)(GenFs.cause, Gen.anyInt)),
         testM("chunk")(checkAllLaws(Covariant)(GenF.chunk, Gen.anyInt)),
         testM("exit")(checkAllLaws(Covariant)(GenFs.exit(Gen.causes(Gen.anyInt, Gen.throwable)), Gen.anyInt)),
-        testM("nonEmptyChunk")(checkAllLaws(Covariant)(GenFs.nonEmptyChunk, Gen.anyInt))
+        testM("nonEmptyChunk")(checkAllLaws(Covariant)(GenFs.nonEmptyChunk, Gen.anyInt)),
+        testM("chunk . option")(checkAllLaws(Covariant)(chunkOptionGenF, Gen.anyInt))
```
sideeffffect, line 27: This is already tested in `TraversableSpec`, so it doesn't make sense to test it again here, no?

sideeffffect (changes requested):
Maybe we don't need the redundant test.

core/shared/src/test/scala/zio/prelude/CovariantSpec.scala
joroKr21: Well, it's a different piece of code. But the implementation is obviously the same.

sideeffffect: `TraversableSpec` literally tests (also) the Covariant laws, so we shouldn't need to test them again here. Or am I missing something? What do you mean by
> different piece of code

:thinking:

joroKr21: It's a different `compose` method

sideeffffect: Oh, of course! :facepalm:  Thanks for the explanation :smile_cat:

sideeffffect (approved):

adamgfraser (approved):
Looks great!

PR #459, opened 2020-12-09
Order sbt plugins

sideeffffect:


mijicd (approved):

mijicd:
In fact, I think some of the plugins are unused and can be either used or removed (e.g. explicit deps).

sideeffffect:
@mijicd That's a good point, but my aim here was mostly to bring the file in line with ZIO and `zio-project-seed`

mijicd (approved):

PR #461, opened 2020-12-10
Further Generalize ZPure

adamgfraser:
This PR explores further generalizing ZPure in two ways:

1. Adding an additional type parameter `W` representing a log that can be written to in the course of executing the computation
2. Adding a concept of being able to accumulate multiple errors through the use of a `zipWithPar` operator that would not do actual concurrency but would perform both computations, even if the first one failed, and accumulate all errors in a `Cause` data structure.

The idea is that with these changes we would be able to delete the `Validation` data type and it would just become:

```scala
type Validation[+E, +A]      = ZPure[Nothing, Unit, Unit, Any, E, A]
type ZValidation[+W, +E, +A] = ZPure[W, Unit, Unit, Any, E, A]
```

`ZPure` would also be able to model the concept of non-fatal errors through the `W` type, potentially allowing `These` to be a more unbiased data representing the concept of either or both versus explicitly modeling non-fatal errors.


core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -2,227 +2,244 @@ package zio.prelude.fx
...
 import scala.util.control.NonFatal
 
 /**
- * `ZPure[S1, S2, R, E, A]` is a purely functional description of a computation
+ * `ZPure[S1, S2, W, R, E, A]` is a purely functional description of a computation
  * that requires an environment `R` and an initial state `S1` and may either
  * fail with an `E` or succeed with an updated state `S2` and an `A`. Because
  * of its polymorphism `ZPure` can be used to model a variety of effects
  * including context, state, and failure.
  */
```
arnoldlacko, line 18: I would like to suggest this small change to the description :)
```suggestion
 * `ZPure[W, S1, S2, R, E, A]` is a purely functional description of a computation
 * that requires an environment `R` and an initial state `S1`, writes to a log `W` and may either
 * fail with an `E` or succeed with an updated state `S2` and an `A`. Because
 * of its polymorphism `ZPure` can be used to model a variety of effects
 * including context, state, and failure.
 */
```

adamgfraser: @arnoldlacko Good catch! Fixed now.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -554,9 +586,16 @@ sealed trait ZPure[-S1, +S2, -R, +E, +A] { self =>
           a = updated._2
           val nextInstr = stack.pop()
           if (nextInstr eq null) curZPure = null else curZPure = nextInstr(a)
+        case Tags.Log     =>
+          val zPure     = curZPure.asInstanceOf[Log[Any, Any]]
+          log = log :+ zPure.log
```
jdegoes, line 591: I would use a mutable array builder here.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -464,38 +492,42 @@ sealed trait ZPure[-S1, +S2, -R, +E, +A] { self =>
...
    */
-  final def runEither(s: S1)(implicit ev: Any <:< R, ev1: CanFail[E]): Either[E, (S2, A)] = {
-    val _                                                   = ev
-    val stack: Stack[Any => ZPure[Any, Any, Any, Any, Any]] = Stack()
-    val environments: Stack[AnyRef]                         = Stack()
-    var s0: Any                                             = s
-    var a: Any                                              = null
-    var failed                                              = false
-    var curZPure: ZPure[Any, Any, Any, Any, Any]            = self.asInstanceOf[ZPure[Any, Any, Any, Any, Any]]
+  final def runEitherCause(s: S1)(implicit ev: Any <:< R, ev1: CanFail[E]): Either[Cause[E], (S2, A)] = {
```
jdegoes, line 502: The log is "eaten", right? That's ok if we expose "get log". In fact, maybe through "get state", we could further simplify this to just `Either[Cause[E], A]`, so if you want the state or the log, you have to get it and bundle it into the result.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
jdegoes: If we change this to array builder, then "get log" needs to do a copy, of course.

adamgfraser: Good idea!

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
adamgfraser: I like that.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -970,38 +1070,62 @@ object ZPure {
...
+  private final case class Provide[W, S1, S2, R, E, A](r: R, continue: ZPure[W, S1, S2, R, E, A])
+      extends ZPure[W, S1, S2, Any, E, A]          {
     override def tag: Int = Tags.Provide
   }
+
+  private final case class Log[S, +W](log: W) extends ZPure[W, S, S, Any, Nothing, Unit] {
+    override def tag: Int = Tags.Log
+  }
+
+  private final case class GetLog[W, S]() extends ZPure[W, S, S, Any, Nothing, Chunk[W]] {
```
adamgfraser, line 1115: I'm not sure this is the optimal encoding of getting the log.

adamgfraser:
@jdegoes This is ready for another review.

core/shared/src/main/scala/zio/prelude/fx/Cause.scala
```diff
@@ -0,0 +1,187 @@
...
+
+import zio.prelude._
+
+import scala.annotation.tailrec
+
+/**
+ * `Cause` is a data type that represents the potentially multiple ways that a
+ * computation can fail.
+ */
+sealed trait Cause[+E] { self =>
```
jdegoes, line 11: This is more general than just failure. Especially if we add an empty, it becomes a free semiring. It maybe should live in `zio.prelude` directly and not speak about errors, just "events" or something, in parallel and sequence. It can be used for errors but could be used for tests and other things. Anywhere you need a tree of parallel / sequential operations.

adamgfraser: Yes, we could definitely generalize it and then just have `Cause` be a type alias for it.

adamgfraser:
@jdegoes I generalized `Cause` to `Semiring`. There is definitely more work I can do there to build that out but I think it may make sense to do that separately as `Semiring` now has all the functionality currently needed by `ZPure` and this PR is already quite large.

adamgfraser:
@jdegoes I added a second type parameter to `Semiring` to represent the empty type. Would appreciate your feedback when you have the chance. It does let us abstract over emptiness but there are a couple of places where the results are not ideal.

core/shared/src/main/scala/zio/prelude/Semiring.scala
```diff
@@ -0,0 +1,398 @@
...
+ * parallel and some of which must be executed sequentially.
+ */
+sealed trait Semiring[+Z <: Unit, +A] { self =>
+
+  /**
+   * Combines this collection of events with that collection of events to
+   * return a new collection of events that represents this collection of
+   * events in parallel with that collection of events.
+   */
+  final def &&[Z1 >: Z <: Unit, A1 >: A](that: Semiring[Z1, A1]): Semiring[Z1, A1] =
```
adamgfraser, line 21: We're losing information right now that combining an empty collection with a nonempty collection results in a nonempty collection. Potentially we could could use overloading or type level programming to recover this.

core/shared/src/main/scala/zio/prelude/Semiring.scala
```diff
@@ -0,0 +1,398 @@
...
+      case Semiring.Then(left, _) => left.first
+      case Semiring.Single(value) => value
+    }
+
+  /**
+   * Constructs a new collection of events for each event in this collection of
+   * events, collecting them back into a single collection of events.
+   */
+  final def flatMap[Z1 >: Z <: Unit, B](f: A => Semiring[Z1, B]): Semiring[Z1, B] =
+    fold(Semiring.empty, f)(_ ++ _, _ && _).asInstanceOf[Semiring[Z1, B]]
```
adamgfraser, line 75: We know that the resulting collection can only be empty if `Z` or `Z1` is `Unit` but we still have to handle the case where the initial collection is empty so we have to cast here.

core/shared/src/main/scala/zio/prelude/Semiring.scala
```diff
@@ -0,0 +1,398 @@
...
+    loop(List(self), List.empty).head
+  }
+
+  /**
+   * Performs the specified effectual function for each event in this
+   * collection of events, collecting them back into a single collection of
+   * events.
+   */
+  final def foreach[F[+_]: IdentityBoth: Covariant, B](f: A => F[B]): F[Semiring[Z, B]] =
+    fold[F[Semiring[Unit, B]]](Semiring.empty.succeed, a => f(a).map(Semiring.single))(
```
adamgfraser, line 116: Similar issue as above.

core/shared/src/main/scala/zio/prelude/Semiring.scala
```diff
@@ -0,0 +1,398 @@
...
+    private def equal(that: Semiring[Unit, Any]): Boolean                           =
+      (self, that) match {
+        case (br: Both[Unit, Any], bl: Both[Unit, Any]) =>
+          br.left == bl.left && br.right == bl.right
+        case _                                          =>
+          false
+      }
+  }
+
+  case object Empty extends Semiring[Unit, Nothing]
```
adamgfraser, line 210: Pattern matches against `Empty` don't work well because it is not known to be a valid subtype if `Z` is abstract, since `Z` could be `Nothing`. Maybe if we go down this route we should make the constructors private and only expose them through folding and the operators we provide?

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -516,38 +541,42 @@ sealed trait ZPure[-S1, +S2, -R, +E, +A] { self =>
   final def run(s: S1)(implicit ev1: Any <:< R, ev2: E <:< Nothing): (S2, A) =
```
adamgfraser, line 541: With the additional functionality we are adding there is a risk of having a ton of different ways to `run` a `ZPure` value. Also, I would like to improve the ergonomics in the many cases where there is no state type.

I was thinking about trying to make the following the primary method to run `ZPure` values:

```scala
final def run(implicit ev1: Unit <:< S1, ev2: Any <:< R, ev3: E <:< Nothing): A
```

Then we could have a `runLog` variant that exposes the value and the log. Users would provide the input state through a `provideState` operator if they had not already set an initial state, similar to how you provide the environment to a `ZIO` effect before running it, and would normally handle their errors before running it, even by just doing `zPure.either`. We could perhaps have a `runEither` variant and maybe a `runState` variant for when you are just doing state but I would like to avoid users having to always provide a `Unit` state type when running computations and having every possible variant of state, logging, error, and value outputs.

core/shared/src/main/scala/zio/prelude/Semiring.scala
```diff
@@ -0,0 +1,398 @@
...
+   * Combines this collection of events with that collection of events to
+   * return the Cartesian product of events using the specified function.
+   */
+  final def zipWith[Z1 >: Z <: Unit, B, C](that: Semiring[Z1, B])(f: (A, B) => C): Semiring[Z1, C] =
+    self.flatMap(a => that.map(b => f(a, b)))
+}
+
+object Semiring {
+
+  final case class Both[+Z <: Unit, +A](left: Semiring[Z, A], right: Semiring[Z, A]) extends Semiring[Z, A] { self =>
```
adamgfraser, line 162: We could potentially make this:

```scala
final case class Both[+A](left: Semiring[Nothing, A], right: Semiring[Nothing, A]) extends Semiring[Nothing, A]
```

Then in `++` and `&&` we could just ignore empty values in composition. Basically say that if there is an empty element it has to occur at the root, there can't be empty elements at arbitrary nodes in the tree.

jdegoes:
Ready to merge, although let's change the name `Semiring` to `ParSeq` or something similar. We can figure out a better name later.

jdegoes (approved):

PR #466, opened 2020-12-13
Remove unnecessary methods from *Both and *Either hierarchy

sideeffffect:
Somewhat related to https://github.com/zio/zio-prelude/issues/372#issuecomment-742935796

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/IdentityEither.scala
+++ b/core/shared/src/main/scala/zio/prelude/IdentityEither.scala
@@ -63,10 +63,11 @@ object IdentityEither extends LawfulF.Invariant[DeriveEqualIdentityEitherInvaria
 
 trait IdentityEitherSyntax {
 
-  implicit class IdentityEitherAnyOps[A](nothing: => Nothing) {
+  implicit class IdentityEitherAnyOps(a: Any) {
 
-    def fail[F[-_]](implicit either: IdentityEither[F], contravariant: Contravariant[F]): F[A] =
-      contravariant.contramap[Nothing, A](_ => nothing)(either.none)
+    /** Ignores its argument and returns a "failed" `F` */
+    def fail[F[_]](implicit either: IdentityEither[F]): F[Nothing] =
+      either.none
   }
 
 }
--- a/core/shared/src/main/scala/zio/prelude/IdentityEither.scala
+++ b/core/shared/src/main/scala/zio/prelude/IdentityEither.scala
@@ -63,45 +63,10 @@ object IdentityEither extends LawfulF.Invariant[DeriveEqualIdentityEitherInvaria
 
 trait IdentityEitherSyntax {
 
-  /**
-   * Provides infix syntax for identity operations for invariant types.
-   */
-  implicit class IdentityEitherOps[F[_], A](fa: => F[A]) {
+  implicit class IdentityEitherAnyOps[A](nothing: => Nothing) {
 
-    /**
-     * Combines two values of types `F[A]` and `F[B]` to produce an
-     * `F[Either[A, B]]`.
-     */
-    def orElseEitherIdentity[B](fb: => F[B])(implicit either: IdentityEither[F]): F[Either[A, B]] =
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/zio/prelude/IdentityEither.scala
```diff
@@ -63,45 +63,10 @@ object IdentityEither extends LawfulF.Invariant[DeriveEqualIdentityEitherInvaria
...
-  implicit class IdentityEitherOps[F[_], A](fa: => F[A]) {
+  implicit class IdentityEitherAnyOps[A](nothing: => Nothing) {
 
-    /**
-     * Combines two values of types `F[A]` and `F[B]` to produce an
-     * `F[Either[A, B]]`.
-     */
-    def orElseEitherIdentity[B](fb: => F[B])(implicit either: IdentityEither[F]): F[Either[A, B]] =
-      either.either(fa, fb)
+    def fail[F[-_]](implicit either: IdentityEither[F], contravariant: Contravariant[F]): F[A] =
```
sideeffffect, line 68: Since we have `succeed` for `IdentityBoth`, I was thinking, whether we could have something analogous for `IdentityEither`.
Do you thing it's meaningful to have this?

jdegoes: `IdentityBoth` has "succeed", while `IdentityEither` has "fail".

jdegoes: I would add two methods, fail and (xyz)map + fail, and not just on contravariant but also covariant and divariant.

adamgfraser (changes requested):
Nice cleanup! Looks good other than revising the implicit syntax.

core/shared/src/main/scala/zio/prelude/IdentityEither.scala
sideeffffect: The more I think about it, the less I'm sure I understand what you want.
Is this what you had in mind?
```scala
  implicit class IdentityEitherNothingOps(nothing: => Nothing) {

    def fail[F[-_], A](implicit either: IdentityEither[F], contravariant: Contravariant[F]): F[A] =
      contravariant.contramap[Nothing, A](_ => nothing)(either.none)

    def fail[F[+_], A](implicit either: IdentityEither[F], covariant: Covariant[F]): F[A] =
      covariant.map[Nothing, A](_ => nothing)(either.none)

    def fail[:=>[-_, +_], B, C](implicit
      either: IdentityEither[({ type lambda[x] = x :=> B })#lambda],
      divariant: Divariant[:=>]
    ): C :=> B =
      divariant.dimap[Nothing, B, C, B](_ => nothing, identity[B])(either.none)
  }
```
What would the _(xyz)map + fail_ method do?

adamgfraser: I think we can do:

```scala
implicit class IdentityAnyOps(a: Any) {
  def fail[F[_](implicit either: identityEither[F]): F[Nothing] =
    either.none
}
```

It is not very exciting but it just gives you infix syntax for creating the failed value, throwing away whatever the original value was.

I'm not sure the `map` variants make much sense. We could do something like `either.none.map(_ => ???)` but since we have an `F[Nothing]` there can never be a `Nothing` value so all it is really doing is evidencing that for a covariant type we can widen it to anything. Similarly for a contravariant type we should never be able to produce a `Nothing` so we shouldn't be able to implement that at all.

sideeffffect: done :heavy_check_mark:

jdegoes: True, so maybe not so useful after all.

jdegoes:
Let me know when ready for a final review!

sideeffffect:
Ready to be reviewed :pray:

adamgfraser (approved):

PR #470, opened 2020-12-16
Adopt the new IntelliJ import order

sideeffffect:
Intellij IDEA has recently changed the order of import blocks. For the good of our contributors, we should follow it.


adamgfraser:
@sideeffffect Good to go when tests are passing!

adamgfraser (approved):

PR #486, opened 2021-01-06
Sketch of NewtypeSmartF and SubtypeSmartF

sideeffffect:
fixes https://github.com/zio/zio-prelude/issues/483

Is this what you had in mind, @jdegoes ?

On a second thought, shouldn't these be called rather `NewtypeSmartK` and `SubtypeSmartK`?

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/Invariant.scala
+++ b/core/shared/src/main/scala/zio/prelude/Invariant.scala
@@ -605,6 +605,10 @@ object Invariant extends LowPriorityInvariantImplicits with InvariantVersionSpec
     new NonEmptyTraversable[Id] {
       override def foreach1[G[+_]: AssociativeBoth: Covariant, A, B](fa: Id[A])(f: A => G[B]): G[Id[B]] =
         f(Id.unwrap(fa)).map(Id(_))
+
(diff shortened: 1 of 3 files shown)


adamgfraser (changes requested):
I think this looks good. Can we add some tests so we can see the functionality in action?

sideeffffect:
Hi, this should be ready for merging :wink:

adamgfraser (approved):
Awesome!

sideeffffect:
Btw, are you guys @adamgfraser @jdegoes comfortable with the name `NewtypeSmartF`, `SubtypeSmartF` (so not `NewtypeSmartK` and `SubtypeSmartK`) and `AssertionK`?

adamgfraser:
I think we use `F` elsewhere so we should stick with that,

sideeffffect:
Should I rename `AssertionK` to `AssertionF`? What is the distinction between `F` and `K`? (AFAIK, `XyzK` communicates that it is like `Xyz`, but on a "kind level", like `Function` is about `A => B` and `FunctionK` is about `A[_] => B[_]`) :thinking:

adamgfraser:
I thought we were just kind of using `F` instead of `K` so I think we should align everything around that.

PR #500, opened 2021-01-17
Test floating point instances with tolerance

sideeffffect:


core/shared/src/main/scala/zio/prelude/Equal.scala
```diff
@@ -246,6 +246,14 @@ object Equal extends Lawful[Equal] {
   implicit def DeriveEqual[F[_]: DeriveEqual, A: Equal]: Equal[F[A]] =
     Derive[F, Equal].derive(Equal[A])
 
+  /**
+   * `Equal` instance for `Double` for imprecise equality with configurable tolerance.
+   */
+  def DoubleEqualWithEpsilon(epsilon: Double = 1.0d / 1024 * 1024 * 1024 * 1024): Equal[Double] = {
```
adamgfraser, line 252: I would move these to test files. There is value in having some tests for the corresponding instances since we decided to have them but this is kind of specific to testing them and these equality instances are not lawful in the sense that for example you could have `a` and `b` be equal within tolerance and `b` and `c` be equal within tolerance but not `a` and `c` being equal within tolerance.

core/shared/src/main/scala/zio/prelude/Equal.scala
```diff
@@ -266,6 +274,13 @@ object Equal extends Lawful[Equal] {
   implicit def EitherEqual[A: Equal, B: Equal]: Equal[Either[A, B]] =
     Equal[A] either Equal[B]
 
+  /**
+   * `Equal` instance for `Float` for imprecise equality with configurable tolerance.
+   */
+  def FloatEqualWithEpsilon(epsilon: Float = 1.0f / 1024 * 1024): Equal[Float] = { (l: Float, r: Float) =>
```
adamgfraser, line 280: I think there is an order of operations issue with the default values here and above. Don't you want this to be `1.0f / (1024 * 1024)`? Otherwise the default value is just `1.0f` which seems like way to high a tolerance.

adamgfraser (changes requested):
Couple of specific comments but good to add tests for these instances.

core/shared/src/main/scala/zio/prelude/Equal.scala
sideeffffect: of course! thank you :+1:

core/shared/src/main/scala/zio/prelude/Equal.scala
sideeffffect: You're right that the instance created by `DoubleEqualWithEpsilon` wouldn't be lawful (unless `epsilon` = 0).
But these kinds of instances will still be practically useful. Not just for us testing various other Type classes, but even more importantly to our users. In practice, when you're working with floating point numbers -- geometry and alike come to mind, but this is a widespread phenomenon -- you very often don't want 100% precise comparison. You need to compare floats within a small "epsilon" tolerance.

That's why `DoubleEqualWithEpsilon` needs to stay in `main`, so that it's accessible to our users.

adamgfraser: That's a good point. I wonder if it makes sense to think of this as a separate abstraction, `TestEqual` or `ApproximatelyEqual`, but let's get this in and we can discuss that.

adamgfraser (approved):

PR #504, opened 2021-01-18
Rename name collisions

### LINKED ISSUE #312: Traversable.foreach name collision
@felher brought this up in the Discord and I thought we should create an issue to track it.

Since most collections already have `foreach` on them, this can result in name collisions. A different name would prevent having to write stuff like `Traversable[Vector].foreach(vec)(function)`

@jdegoes also agreed that it needs to be renamed because of this.

It's not a problem in ZIO itself, but given how providing syntax/ops in Scala works, it's an issue in zio-prelude.

Also, it might be a good idea to give other methods a quick scan to see if there are other common name collisions. Better to catch them early.

kitlangton:
closes #312 as well as a collision for `reduce` in NonEmptyTraversable.

I left untouched anything that didn't collide, including `foreach1` and `foreach_`. Perhaps we want to change these as well for internal consistency? 

UPDATE: As discussed below, I've actually undone this ^^^ and instead added `reduce0` and `foreach0` to _only_ the implicit classes.

### DIFF
--- a/benchmarks/src/main/scala/zio/prelude/StateBenchmarks.scala
+++ b/benchmarks/src/main/scala/zio/prelude/StateBenchmarks.scala
@@ -116,7 +116,7 @@ class StateBenchmarks {
   @Benchmark
   def zioEffectfulTraversal(): Int =
     State
-      .foreach(list) { el =>
+      .forEach(list) { el =>
         State.get[Int].flatMap(s => State.set(s + el))
       }
       .runState(0)
--- a/core/shared/src/main/scala-2.11-2.12/zio/prelude/InvariantVersionSpecific.scala
+++ b/core/shared/src/main/scala-2.11-2.12/zio/prelude/InvariantVersionSpecific.scala
@@ -5,11 +5,11 @@ import scala.collection.generic.CanBuildFrom
 trait InvariantVersionSpecific {
 
   /**
-   * Derives a `Traverable[F]` from an `Iterable[F]`.
+   * Derives a `ForEach[F]` from an `Iterable[F]`.
    */
-  implicit def IterableTraversable[F[+a] <: Iterable[a]](implicit derive: DeriveCanBuildFrom[F]): Traversable[F] =
-    new Traversable[F] {
-      def foreach[G[+_]: IdentityBoth: Covariant, A, B](fa: F[A])(f: A => G[B]): G[F[B]] =
+  implicit def IterableForEach[F[+a] <: Iterable[a]](implicit derive: DeriveCanBuildFrom[F]): ForEach[F] =
+    new ForEach[F] {
+      def forEach[G[+_]: IdentityBoth: Covariant, A, B](fa: F[A])(f: A => G[B]): G[F[B]] =
         fa.foldLeft(derive.derive[B](fa).succeed)((bs, a) => bs.zipWith(f(a))(_ += _)).map(_.result())
     }
 
--- a/core/shared/src/main/scala-2.13+/zio/prelude/InvariantVersionSpecific.scala
+++ b/core/shared/src/main/scala-2.13+/zio/prelude/InvariantVersionSpecific.scala
@@ -5,11 +5,11 @@ import scala.collection.BuildFrom
 trait InvariantVersionSpecific {
 
   /**
-   * Derives a `Traverable[F]` from an `Iterable[F]`.
+   * Derives a `ForEach[F]` from an `Iterable[F]`.
    */
-  implicit def IterableTraversable[F[+a] <: Iterable[a]](implicit derive: DeriveBuildFrom[F]): Traversable[F] =
-    new Traversable[F] {
-      def foreach[G[+_]: IdentityBoth: Covariant, A, B](fa: F[A])(f: A => G[B]): G[F[B]] =
+  implicit def IterableForEach[F[+a] <: Iterable[a]](implicit derive: DeriveBuildFrom[F]): ForEach[F] =
+    new ForEach[F] {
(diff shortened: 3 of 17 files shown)


sideeffffect (commented):
Thanks for rolling up your sleeves and doing the work on this! :muscle: 

What do you guys think about keeping the `foreach` name and just adding `traverse` as yet another extension method (an alias) for cases where there would be conflict with the cannonical, which would still be `foreach`?

(For the record, I would be even for renaming `Traverse` to `Foreach` :pirate_flag: absolute mad lad, I know 🤘)

kitlangton:
Oh no! I've failed my lint! I'll make sure to run all the magic commands. But first I'll wait for a decision re: the final syntax ;)

sideeffffect:
Having `0` suffixed aliases to commonly conflicting methods could also work :+1:

sideeffffect:
btw the only magic command you'll ever need is `fix` :point_left: it fixes things :sunglasses:

kitlangton:
> Having `0` suffixed aliases to commonly conflicting methods could also work 👍

I'll try this for now. I like this because you don't have to memorize a special case for each instance:
"Oh, foreach is forEach, reduce is reDuce/reduceAll, I forget!"😜

kitlangton:
Okie dokie! @sideeffffect 
An extremely simple PR :P Only added a couple of methods to escape the shadowing. I didn't even remove the originals, in case you use explicitly wrap the argument in the Ops constructor. Happy to delete them if that's better though.

core/shared/src/main/scala/zio/prelude/NonEmptyTraversable.scala
```diff
@@ -168,6 +168,9 @@ trait NonEmptyTraversableSyntax {
       F.foreach1_(self)(f)
     def reduce(f: (A, A) => A)(implicit F: NonEmptyTraversable[F]): A                                             =
       F.reduce(self)(f)
+    // Another version to prevent name collision with the built-in `reduce`
```
sideeffffect, line 171: let's not be shy, a comment like this should be `/**`, not `//`, so that users in IDE with auto-completion can know right away what's going on.

sideeffffect (changes requested):
Using doc comments will enhance the UX
Otherwise :+1: from me.
Let's wait for @adamgfraser / @jdegoes though, because this is about establishing a general pattern how to deal with conflicting extension methods, not just these two new functions.

core/shared/src/main/scala/zio/prelude/NonEmptyTraversable.scala
kitlangton: Gotcha! Upgraded to a bona fide doc string. And agreed re: waiting for other opinions :) Thanks, @sideeffffect!

sideeffffect (dismissed):
:+1: 
Thanks!

adamgfraser:
@kitlangton Thanks for working on this! Good to finally get this taken care of.

`reduceAll` seems like a great name.

I'm a little torn on the replacement for `foreach`. `traverse` does have some logic to it in that something that is `Traversable` is something you can `traverse`. However, the core market for this library is people who are already using ZIO so they are going to be familiar with `foreach` and both that name and consistency have pedagogical benefits. Maybe we can distinguish based  on this being the true "returns the same collection" type version?

On the other hand having different capitalization is not great either.

@jdegoes What do you think?

kitlangton:
@adamgfraser Oh! Sorry, I should've updated the initial message (just did that). I actually changed the implementation based on the discussion w/ @sideeffffect . I ended up taking this approach:

> UPDATE: As discussed below, I've actually undone this ^^^ and instead added reduce0 and foreach0 to only the implicit classes.
>  In fact, for consistency's sake, it might be nice to just add an ugly postfix, like foreach0, just for those syntax cases that get clobbered. Then we'd be able to use whatever names make the most sense everywhere, and if something's clobbered in an extension, you know exactly what to reach for.

jdegoes:
I would suggest one of the following two possibilities:

 - `reduceAll` (avoids the `0` suffix, which I've found people don't like)
 - `traverse` (for the version that returns the same structure)
 - `forEach` (for the version that only works on collections)
 - `forEach_` (for a version that does not build any specific value)

OR:

 - `reduceAll` (avoids the `0` suffix, which I've found people don't like)
 - `forEach` (for the version that returns the same structure)
 - `forEach_` (for a version that does not build any specific value)

This is somewhat close to ZIO, enough to be familiar to ZIO users. We could deprecate `ZIO.foreach` and upgrade to `ZIO.forEach` in the next major release (2.0).

kitlangton:
> * `reduceAll` (avoids the `0` suffix, which I've found people don't like)
> * `forEach` (for the version that returns the same structure)
> * `forEach_` (for a version that does not build any specific value)

I'm into this one. I'll give this a whirl 👍

jdegoes:
For the 2nd possibility, I think it would be useful to, as @sideeffffect suggest, change the name of the type class to `ForEach`.

kitlangton:
Going all in on ForEach 😎

sideeffffect:
Don't forget `NonEmptyForEach` :wink:

kitlangton:
And of course `DeriveEqualNonEmptyForEach` @sideeffffect  ;)

Well. Here it is. Lots of renames. I did keep `reduce` on the concrete instances (NonEmptyList, etc). I just changed it on NonEmptyTr— I mean, `NonEmptyForEach`

sideeffffect (approved):

jdegoes (approved):

jdegoes:
I have no objections if everyone else is happy!

jdegoes:
@kitlangton Thank you! 🙏

PR #509, opened 2021-01-25
Add some instances for Parallel collections

sideeffffect:
This PR adds instances for the Parallel collections (ParMap, ParSeq and ParSet) for the Equal and Invariant hierarchy. I was able to find nice, law-abiding instances for ForEach, so I could remove the ad-hoc extension methods.

I would like make the case, why it makes sense to have these in ZIO Prelude core:
 * Parallel collections are part of standard library for 2.11 and 2.12
 * We will support them (or 2.12 at least) for a long time
 * That means we can afford to treat Parallel collections as if they were standard library classes
 * For 2.13+, the `scala-parallel-collections` JAR is added as an _optional_ dependency, so it won't burden people who don't actually use it
 * Instances can be in Type classes' companion objects, i.e. the implicit scope
 * Users will be able to work with these instances without any imports => better user experience

/cc @adamgfraser @jdegoes


sideeffffect:
We're green :tada:

adamgfraser:
I'm 👍 on including this as a separate module.

Without getting into any merits of the library it is not part of the Scala standard library on 2.13 or Scala 3. There are a lot of libraries that it would be good to provide instances for. I would argue that providing instances for data types defined in other ZIO ecosystem libraries is at least as important as this. But I don't think it is sustainable to support these directly in the core library.

We are already going in a lot of directions between instances for standard library and ZIO data types, `ZPure`, additional abstractions in the arrow, lattice, and ring hierarchies, bifunctor, trifunctor, and constrained functor types, and others, which is great. This seems like a really easy area for us to be a little more modular with a separate module in this repository for these data types.

Scala 2.13 and Scala 3 users are already having to do a separate import to get these collections at all so I don't see the need to do something like `import zio.prelude.parcollections._` as much of an issue.

sideeffffect:
Any opinion on this @jdegoes ?

I proposed keeping the instances in the same module, because I'm avoiding requiring our users to do special implicit imports (ala `cats.syntax.all._`) as much as possible. Those kinds of imports have been the bane of Scala and have made implicits and FP in Scala a bad name.

But @adamgfraser 's arguments also make sense -- the parallel collections aren't in Scala 2.13 and 3 after all.

If you don't have any strong desire to avoid implicits imports, I'll go with @adamgfraser 's suggestion and put the code into separate module.

adamgfraser:
@sideeffffect I hear you. I just feel like users already have to do additional imports. If you are on Scala 2.13 or Scala 3, which is what I think we have got to be optimizing for, you already have to add a separate dependency and do a separate import.

```scala
libraryDependencies +=
  "org.scala-lang.modules" %% "scala-parallel-collections" % "1.0.0"
```

```scala
import scala.collection.parallel.CollectionConverters._
```

So if you are doing that I think it is reasonable that you are also going to have to do an import if you want implicit instances for those data types. This is just the same as if you wanted to use third party libraries with ZIO where you need to import both the third party library and an interior package for ZIO and the third party library. This is going to be the case for third party libraries in general unless they are in the ZIO ecosystem and want to depend on ZIO Prelude and implement their own instances in the companion object for their data types.

sideeffffect:
@adamgfraser I've moved all the `scala-parallel-collections` instances to a dedicated module. I hope users will find it easy to work with. We can always revisit this decision if it turns out not to be the case.

adamgfraser (approved):
Let's get this merged!

PR #510, opened 2021-01-25
Add Variants For Running ZPure Values

adamgfraser:
With the addition of the `W` type parameter to `ZPure` we have a little bit of an issue with a combinatorial explosion of different ways to run `ZPure` values:

- Do you want the log, the result, or both?
- Do you want the state, the result, or both?
- Do you want the first error, the full cause, or to require that errors were already handled?
- Do you want to provide an input state or require that no input state is provided?

To address this, this PR adds a set of more composable operators to improve ergonomics of simple use cases as well as allow flexibility for any combination of these choices the user wants.

The fundamental operator is `runAll`, which requires an initial state and returns the full result of the computation including the log and either all errors that occurred or the final result and updated state. It is expected that users will rarely if ever need to call this but it is available if necessary and provides maximum flexibility.

There are then a set of convenience operators for common cases included `run`, `runEither`, `runState`, and `runResult` to allow users working with more monomorphic versions of `ZPure` without having to interact with additional type parameters.

Finally, there are a set of operators, some of which already existed, to allow flexibility in what is returned:

- `sandbox` / `unsandbox` to expose or submerge the cause of failure
- `either` / `absolve` to surface errors into the value channel or submerge them
- `getState` / `provideState` to surface the output state into the value channel or provide the input state

With these the user has full flexibility in a composable way of what they want returned.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
+++ b/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
@@ -3,7 +3,7 @@ package zio.prelude.fx
 import zio.internal.Stack
 import zio.prelude._
 import zio.test.Assertion
-import zio.{CanFail, ChunkBuilder, NeedsEnv}
+import zio.{CanFail, Chunk, ChunkBuilder, NeedsEnv}
 
 import scala.annotation.{implicitNotFound, switch}
 import scala.util.Try
@@ -295,6 +295,12 @@ sealed trait ZPure[+W, -S1, +S2, -R, +E, +A] { self =>
     foldCauseM((cause: Cause[E]) => failure(cause.first), success)
 
   /**
+   * Exposes the output state into the value channel.
+   */
+  final def getState: ZPure[W, S1, S2, R, E, (S2, A)] =
+    flatMap(a => get.map(s => (s, a)))
+
+  /**
    * Returns a successful computation with the head of the list if the list is
    * non-empty or fails with the error `None` if the list is empty.
    */
@@ -460,6 +466,12 @@ sealed trait ZPure[+W, -S1, +S2, -R, +E, +A] { self =>
     ZPure.accessM(r0 => self.provide(f(r0)))
 
   /**
+   * Provides this computation with its initial state.
+   */
+  final def provideState(s: S1): ZPure[W, Any, S2, R, E, A] =
+    set(s) *> self
+
+  /**
    * Fail with the returned value if the `PartialFunction` matches, otherwise
    * continue with our held value.
    */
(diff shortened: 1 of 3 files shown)


felher:
Awesome! Thank you :rocket: 

Just one thing: Could it be that the implicit evidence for "no initial state needed" should actually be `Any <:< S1` instead of `S1 <:< Unit`? I think this would also be in line with other signatures like `ZIO.fail`, which gives you a `ZPure` with `S1 = Any, S2 = Nothing`.

adamgfraser:
@felher No because we want to be able to run a `ZPure` computation if `S1` is either `Any` or `Unit`, since in either case we can provide the value. Note that it is `Unit <:< S1`, not `S1 <:< Unit`, so if `S1` is equal to `Unit` or `Any` the requirement is satisfied.

felher:
Oh, yeah, you are completely right. I read that the wrong way around!

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -541,21 +553,25 @@ sealed trait ZPure[+W, -S1, +S2, -R, +E, +A] { self =>
...
 
   /**
-   * Runs this computation with the specified initial state, returning either a
-   * failure or the updated state and the result
+   * Runs this computation with the specified initial state, returning both the
+   * log and either all the failures that occurred or the updated state and the
+   * result.
    */
-  final def runEitherCause(s: S1)(implicit ev: Any <:< R, ev1: CanFail[E]): Either[Cause[E], (S2, A)] = {
+  final def runAll(s: S1)(implicit ev: Any <:< R): (Chunk[W], Either[Cause[E], (S2, A)]) = {
```
sideeffffect, line 574: I find it counter-intuitive that `Chunk[W]` is returned _even if_ the computation finished successfully.

I thought that `ZPure` is now also a replacement for `Validated` (or was it named `Validation`?) which aggregates multiple errors (instead of bailing on the first one).

What should _ZIO Prelude_'s story be for something like this?

```scala
def parseInt(s: String): Validation[ParseError, Int] = ???

val list = List("1", "2", "hello", "3", "world")
val parsed = list.traverse(parseInt)

List(ParseError(...), ParseError(...))
```

---

I'm also wondering, whether we need the distinction between `W` and `E`. Could we have just one error type `E` and then have two methods on ZPure:
 * one would accumulate all the errors and gave you `Either[Chunk[E], A]
 * the other would bail on first encountered error and returned `Either[E, A]`

:thinking:

adamgfraser: If multiple errors occur they are captured in the `Cause` which reflects the full structure of parallel and sequential errors and can be exposed using operators like `Sandbox`. I don't think we can combine `W` and `E` because `W` is a log that we want include in the success case as well, for example we might just be logging what we are doing so we can go back and look at the steps we took later.

adamgfraser: You can get the two different behaviors you are talking about by using `foreach` (fails fast) versus `foreachPar` (accumulates errors) and then using one of the operators that exposes all the errors that occurred.

sideeffffect: Oh, I've misunderstood the purpose of `W` then. Thanks for the explanation :+1:

sideeffffect (approved):

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
adamgfraser: Yeah, I realize there was a lot that went into that one PR. Originally there was `Validation` and then we wanted to augment it to be `ZValidation` with the log parameter and that got me thinking about unifying with `ZPure`. But yeah `W` is a writer monad effect which could be used for logging with different interpretations of what you have done or potentially a list of warnings that were triggered but didn't prevent the computation from continuing, and then `E` through the `Cause` structure handles the accumulation of errors and more ability to retain the structure of what order they occurred in than a traditional non-empty collection.

PR #513, opened 2021-01-28
Initial Work To Support Recursion Schemes

adamgfraser:


regiskuckaertz:
This is great! I forgot to say that I explored the divariant encoding. It went well, I had both covariant, contravariant and divariant recursion schemes working ... but only on invariant type constructors. The problem is that in practice, divariant recursion is not very useful, except is some edge case like HOAS and some such. It _may_ be of use for zio-codec though. But there are major caveats explained [here](http://comonad.com/reader/2008/rotten-bananas/).

core/shared/src/main/scala/zio/prelude/recursive/Coalgebra.scala
```diff
@@ -0,0 +1,27 @@
+package zio.prelude.recursive
+
+import zio.prelude._
+
+sealed trait Coalgebra[Case[+_], A] extends (A => Case[A]) { self =>
+
+  def zipWith[B](that: Coalgebra[Case, B])(implicit both: AssociativeBoth[Case]): Coalgebra[Case, (A, B)] =
```
sideeffffect, line 7: Else where we use `zipWith` for zipping other data structure _with some function_. Could we name this `contrazip`, `cozip` or something like that? :thinking: 

I'm not really sure, just thinking out loud.

sideeffffect (dismissed):
A small clashing of names, but I don't have a clearly better name myself :shrug:

core/shared/src/main/scala/zio/prelude/recursive/Coalgebra.scala
adamgfraser: Oh I think we should just rename it `zip`.

sideeffffect: We won't have an object which is an instance of both algebra and co-algebra, so there's no risk of a clash, right?

adamgfraser: Right.

sideeffffect (approved):

core/shared/src/main/scala/zio/prelude/recursive/Algebra.scala
```diff
@@ -0,0 +1,27 @@
+package zio.prelude.recursive
+
+import zio.prelude._
+
+sealed trait Algebra[Case[+_], A] extends (Case[A] => A) { self =>
```
jdegoes, line 5: I'd suggest we call this `Folder`, because while it is an algebra, it is only an algebra in the sense of "algebraic data type", e.g. sum and product. Whereas, if we call it `Folder`, we can describe it as something that folds all the cases to an `A`.

core/shared/src/main/scala/zio/prelude/recursive/Coalgebra.scala
```diff
@@ -0,0 +1,27 @@
+package zio.prelude.recursive
+
+import zio.prelude._
+
+sealed trait Coalgebra[Case[+_], A] extends (A => Case[A]) { self =>
```
jdegoes, line 5: `Unfolder`

core/shared/src/main/scala/zio/prelude/recursive/Recursive.scala
```diff
@@ -0,0 +1,33 @@
+package zio.prelude.recursive
+
+import zio.prelude._
+
+final case class Recursive[Case[+_]](caseValue: Case[Recursive[Case]]) {
```
jdegoes, line 5: Very nicely done!

adamgfraser:
@jdegoes I like it! So much easier to understand!

jdegoes (approved):

PR #516, opened 2021-01-30
Replace CircleCI with Github actions

mijicd:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,97 @@
...
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.ivy2/cache
+            ~/.sbt
+            ~/.m2
+            ~/.cache
+          key: sbt-cache-${{ hashFiles('**/*.sbt') }}-${{ hashFiles('project/build.properties') }}
+      - name: Setup GPG
+        uses: olafurpg/setup-gpg@v3
```
sideeffffect, line 90: Thankfully, `setup-gpg` is no longer needed since https://github.com/olafurpg/sbt-ci-release/releases/tag/v1.5.5

.github/workflows/ci.yml
```diff
@@ -0,0 +1,97 @@
...
+      fail-fast: false
+    steps:
+      - name: Checkout current branch
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 0
+      - name: Setup Scala and Java
+        uses: olafurpg/setup-scala@v10
+      - name: Cache scala dependencies
+        uses: actions/cache@v2
```
sideeffffect, line 25: I would just use https://github.com/coursier/cache-action and be done with it :smile:

sideeffffect (changes requested):
It's amazing news that we're moving to GitHub Actions :tada: 

Since we're doing this (almost) from the scratch, I think it would be best for us, if we started with [`sbt-github-actions`](https://github.com/djspiewak/sbt-github-actions) right from the start :wink: 

It will be the easiest for us to maintain, no loose duplicate fields (like versions of Scala) and so on.

mijicd:
@sideeffffect Regarding Daniel's plugin: we aren't starting from scratch :). Many projects are already migrated to Github actions, so I'd prefer to keep the setup aligned among them.

.github/workflows/ci.yml
mijicd: Oh this is cool.

.github/workflows/ci.yml
mijicd: Neat, updated.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,74 @@
+name: CI
+
+on:
+  pull_request:
+    branches: ['*']
```
sideeffffect, line 5:
```suggestion
```
AFAIK we don't even need the line here

sideeffffect (dismissed):

.github/workflows/clean.yml
```diff
@@ -0,0 +1,50 @@
+name: Clean
```
sideeffffect, line 1: If you pardon my curiosity, what is the purpose of this?

mijicd: This is actually auto-generated stuff from Daniel's plugin, intended to clean artifacts produced by build.

sideeffffect: Yeah, but why would you want to do that? The whole CI environment disappears after it finishes, so why bother deleting individual files?

mijicd: We can clean it up.

PR #523, opened 2021-02-02
Add Higher Arity Variants Of ZPure#zipWith and Pure#zipWithPar Operators

adamgfraser:


jdegoes (approved):

jdegoes:
Compilation error in Dotty JVM.

sideeffffect:
Heroic work!
Sorry to be such a bother, but it would be nice to have these `mapN`/`mapParN`/`tupled`/`tupledPar` also as extension methods on tuples (like we do for other cases).
But I can do that in a new PR, if you don't have the time to do it :muscle:

sideeffffect:
> Shouldn't that be picked up by the existing syntax methods on the type classes?

Ha! Of course they should, good point :+1: . However, for the `*Par` cases, `ZPure` would need to have an `CommutativeBoth` instance (so far, it only has `IdentityBoth`).
Could you also test this? No need for all 22 variants, maybe for just, let's say, a tuple of 3?

adamgfraser:
It actually does have a low priority `CommutativeBoth` instance. Trying to sort out some Scala 3 issues and will add a test for that.

nicolasdasriaux:
> > Shouldn't that be picked up by the existing syntax methods on the type classes?
> 
> Ha! Of course they should, good point 👍 . However, for the `*Par` cases, `ZPure` would need to have an `CommutativeBoth` instance (so far, it only has `IdentityBoth`).
> Could you also test this? No need for all 22 variants, maybe for just, let's say, a tuple of 3?

To be able to implement smart constructors for case classes, I think we need to have tuples up to 22. Please do not restrict it to 3.

adamgfraser:
@nicolasdasriaux I think the suggestion was just that we don't need to add tests for the implicit syntax working correctly for all sizes of tuples. Tuples up to size 22 will be supported.

nicolasdasriaux:
> @nicolasdasriaux I think the suggestion was just that we don't need to add tests for the implicit syntax working correctly for all sizes of tuples. Tuples up to size 22 will be supported.

Oh yes ! Sometimes I read too quickly.

adamgfraser:
Fixed the Dotty compilation issue, added a test for the `CommutativeBoth` syntax, and fixed the naming of the `CommutativeBoth` extension methods.

Also opened an issue for what I believe is a regression in implicit search in Dotty at lampepfl/dotty#11294.

sideeffffect (approved):
Good job. I'm especially happy that this standardizes the naming :+1:

PR #525, opened 2021-02-03
Add URIO instances for the AssociativeCompose family

sideeffffect:
/cc @lemastero @tues @jorge-vasquez-2301


experimental/shared/src/main/scala/zio/prelude/experimental/EitherCompose.scala
```diff
@@ -44,6 +46,24 @@ object EitherCompose {
...
+
+    def toLeft[A]: URIO[A, Either[A, Nothing]] = URIO.access(Left(_))
+
+    def toRight[B]: URIO[B, Either[Nothing, B]] = URIO.access(Right(_))
+
+    def fromEither[A, B, C](a2c: => URIO[A, C])(b2c: => URIO[B, C]): URIO[Either[A, B], C] =
+      URIO.accessM[Either[A, B]] {
+        case Left(a)  => a2c.provide(a)
+        case Right(b) => b2c.provide(b)
+      }
```
adamgfraser, line 62:
```suggestion
      a2c ||| b2c
```

experimental/shared/src/main/scala/zio/prelude/experimental/BothCompose.scala
```diff
@@ -57,6 +59,32 @@ object BothCompose {
...
 
   }
+
+  implicit val URIOApplicationCompose: ApplicationCompose[URIO] = new ApplicationCompose[URIO] {
+
+    type -->:[-t, +r] = URIO[t, r]
+
+    def application[A, B]: URIO[(URIO[A, B], A), B] = URIO.accessM[(URIO[A, B], A)] { case (a2b, a) => a2b.provide(a) }
+
+    def curry[A, B, C](f: URIO[(A, B), C]): URIO[A, URIO[B, C]] =
```
adamgfraser, line 69: I'm not sure that `curry` and `uncurry` is really an identity here in the presence of effects. I'm not sure if that is a guarantee you are expecting to hold.

sideeffffect: I know that when side effects are involved, anything goes, but I was thinking that these still might be useful instances anyway.

adamgfraser: Sure there is a general "anything goes" caveat but we still have a lot of properties we state hold about properties like interruption.

My concern is more specific. If you start with an effect `nested: URIO[A, URIO[B, C]]` the outer effect may perform some work, say incrementing a `ref1` and the inner effect may perform some other work, say increment `ref2`. In `f` those are separate effects and we can observe that by running the outer effect, checking the `Ref` values, and then running the inner effect. If we call `uncurry` on that we collapse it into one effect. Calling `curry` let's us pull the environment out, but we get a new nested effect where if we run the outer effect etither `Ref` will be updated and the inner effect will update both of them. So these operations aren't value preserving in a way that is pretty directly observable.

More conceptually, if an `AssociativeCompose` is a function with "something extra", `curry` and `uncurry` says we can split and recombine these function like things in a value preserving way but we actually can't do that with the "something extra" that makes `URIO[A, B]` different from `A => B`.

sideeffffect: Yeah, I think I see what you're getting at. `uncurry` smashes the effects together and `curry` can't "unsmash" them, it can only pull out the environment again.

Perhaps that's a good reason for not making this instance.

I'll make it just `BothCompose[URIO]`. If the need to make it anything more arises, we can add it.

core/shared/src/main/scala/zio/prelude/AssociativeCompose.scala
```diff
@@ -25,6 +27,12 @@ object AssociativeCompose {
       bc.compose(ab)
 
   }
+
+  implicit val URIOIdentityCompose: IdentityCompose[URIO] = new IdentityCompose[URIO] {
+    def identity[A]: URIO[A, A] = URIO.environment
+
+    def compose[A, B, C](bc: URIO[B, C], ab: URIO[A, B]): URIO[A, C] = ab.flatMap(bc.provide)
```
adamgfraser, line 34:
```suggestion
    def compose[A, B, C](bc: URIO[B, C], ab: URIO[A, B]): URIO[A, C] = ab >>> bc
```

experimental/shared/src/main/scala/zio/prelude/experimental/BothCompose.scala
```diff
@@ -57,6 +59,22 @@ object BothCompose {
     }
 
   }
+
+  implicit val URIOApplicationCompose: BothCompose[URIO] = new BothCompose[URIO] {
```
adamgfraser, line 63: Shouldn't we just define this instance in `AssociativeCompose` based on our usual organization strategy?

experimental/shared/src/main/scala/zio/prelude/experimental/BothCompose.scala
```diff
@@ -57,6 +59,22 @@ object BothCompose {
...
+
+    type :*:[+f, +s] = Tuple2[f, s]
+
+    def fromFirst[A]: URIO[(A, Any), A] = URIO.access[(A, Any)](_._1)
+
+    def fromSecond[B]: URIO[(Any, B), B] = URIO.access[(Any, B)](_._2)
+
+    def toBoth[A, B, C](a2b: URIO[A, B])(a2c: URIO[A, C]): URIO[A, (B, C)] =
+      URIO.accessM[A](a2b.zip(a2c).provide)
```
adamgfraser, line 72:
```suggestion
      a2b &&& a2c
```

experimental/shared/src/main/scala/zio/prelude/experimental/EitherCompose.scala
```diff
@@ -44,6 +46,21 @@ object EitherCompose {
       case Right(b) => b2c(b)
     }
   }
+
+  implicit val URIOEitherCompose: EitherCompose[URIO] = new EitherCompose[URIO] {
```
adamgfraser, line 50: Same thing here. Can't all these instances be collapsed into one instance at the top of the hierarchy and then we don't have to keep reimplementing `compose`?

experimental/shared/src/main/scala/zio/prelude/experimental/BothCompose.scala
sideeffffect: We can't currently do that, because `AssociativeCompose` is in _core_, while `BothCompose` is in _experimental_.

Or we could move `BothCompose` to _core_ and do it there, I'm all for it :wink:

adamgfraser: Ah I see.

adamgfraser (approved):

PR #526, opened 2021-02-03
Build website as part of CI

sideeffffect:


.github/workflows/ci.yml
```diff
@@ -11,8 +11,6 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-20.04
-    strategy:
-      fail-fast: false
```
sideeffffect, line 15: no point having this config if not in `matrix` mode

sideeffffect:
Question also is, whether shouldn't run Website (i.e. `docs/docusaurusPublishGhpages`) only after `publish` (`ci-release`) has successfully finished
/cc @mijicd

.github/workflows/ci.yml
```diff
@@ -51,9 +49,23 @@ jobs:
         if: ${{ matrix.scala == '3.0.0-M3' && matrix.platform == 'JVM' }}
         run: sbt ++${{ matrix.scala }}! testJVM
 
+  buildWebsite:
```
softinio, line 52: this step is pointless as it doesn't get published anywhere. This just results in the build running slower. I would not add this.

softinio:
> Question also is, whether shouldn't run Website (i.e. `docs/docusaurusPublishGhpages`) only after `publish` (`ci-release`) has successfully finished
> /cc @mijicd

The goal of a successful release is for both the artifacts to be published to maven successfully and the docs to be published successfully to github pages. So no need to have one depend on the other, they can run in parallel to save time and release deemed only a success if both successful. 

In a future improvement to our pipeline will try to cater for recovering from one or the other failing without having to do a new release but that is further down the road on the roadmap.

softinio:
@sideeffffect @mijicd @adamgfraser Made a few comments for you to review on this PR. Aside from those comments note that the changes in the PR are deviating from the build flow we are trying to standardize on. Deviating from any standardization is fine but at some point it will cause delays in problem resolution. I would not proceed with the changes proposed in this PR.

adamgfraser:
@softinio Thank you! I think this was just an attempt to get things working but I think there is no need for us to deviate from the standardized build as long as we can figure out how to get release process working reliably.

.github/workflows/ci.yml
sideeffffect: The point here is not to publish the website, but verify that it is buildable. Why shouldn't it be part of CI?

sideeffffect: @softinio Why shouln't building the website be part of CI? Or is it a process which is basically guaranteed to succeed? Or am I missing something?

sideeffffect:
@mijicd @softinio Maybe I'm missing something, but why wouldn't we want to build website as part of CI? It seems like a reasonable thing to do, not only for ZIO Prelude, but also for other ZIO Projects.

PR #531, opened 2021-02-03
add timeout to ci

softinio:


adamgfraser (approved):

mijicd (approved):
Should it be propagated?

adamgfraser:
@mijicd I think so. I think this would definitely be useful for other projects.

softinio:
> Should it be propagated?

yes

PR #533, opened 2021-02-04
Implement ZPure#mapErrorCause

adamgfraser:
This is a useful operator for transforming the full cause of failure that we did not have on `ZPure`.


sideeffffect (approved):
Nice :+1:

nicolasdasriaux:
Nice 👍. Love the name. And much more efficient than the implementation I had initially thought about.

PR #544, opened 2021-02-17
Implement Const

adamgfraser:


sideeffffect (approved):
Wow, that was quick :+1:

sideeffffect:
Can we also have an Associative/Identity-Flatten instance for Const?

adamgfraser:
@sideeffffect I just added an `AssociativeFlatten` instance. We can't implement an `IdentityFlatten` instance because the `B` type is a phantom type so if we map a new `Const` value inside and `flatten` we don't get it back. In essence the associative operation for flattening with respect to `Const` is just taking the first value from the outermost `Const`, because that is the only one that actually exists, so `first` is associative but does not have an identity.

sideeffffect (approved):

PR #545, opened 2021-02-17
Fix Type Signature in ConstForEach Implementation

adamgfraser:


sideeffffect (approved):
good catch! I totally missed it :smile:

jdegoes (approved):

PR #553, opened 2021-03-02
Add ZPure#toZio

sideeffffect:


core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -777,6 +777,13 @@ sealed trait ZPure[+W, -S1, +S2, -R, +E, +A] { self =>
 
   def tag: Int
 
+  /**
+   * Transforms ZPure to ZIO that either succeeds with `A` or fails on the first error `E`.
+   */
+  def toZio(implicit ev: Unit <:< S1): zio.ZIO[R, E, A] = zio.ZIO.accessM[R] { r =>
+    zio.ZIO.fromEither(provide(r).runEither)
```
sideeffffect, line 784: ZPure can fail with a whole `fx.Cause[E]`. Is there a way to "forward" this tree of errors to ZIO's tree of errors (i.e. `zio.Cause`)?

jdegoes: I think this should accept a `ZRef` with `S1` / `S2`, and be called `toZIO` or maybe `toEffect`. Then as a special case, we can provide a variant that works with `Unit` state type.

In addition, we should probably provide access to the write log.

sideeffffect: Did you have something like this in mind?

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -777,6 +777,53 @@ sealed trait ZPure[+W, -S1, +S2, -R, +E, +A] { self =>
 
   def tag: Int
 
+  /**
+   * Transforms ZPure to ZIO that either succeeds with `A` or fails with the first error `E`.
+   * State is read and written from/to a `ZRef`, log is written to another `ZRef`.
+   */
+  def toZIO[E1 >: E](s: ZRef[E1, E1, S2, S1], log: ZRef[E1, E1, Chunk[W], Any]): zio.ZIO[R, E1, A] =
```
jdegoes, line 784: The ref for state is perfect. However, I would make this one return a tuple with the log in it. And maybe call it `toZIOWithLog` or `toZIOLogged`.

The other ones are perfect. 👍 ❤️

sideeffffect: But where would the log go? ZPure returns a log regardless if it fails or succeeds (`runAll: (Chunk[W], Either[Cause[E], (S2, A)])`).

adamgfraser: So can't we return the log in both the error and value channels since in either case there will be a log?

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
adamgfraser: Should it accept a `ZRef` or just return an `S1 => ZIO[R, E, (S2, A)`? Feels like the?

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
sideeffffect: What I wanted to say is that returning it via writing it to a Ref felt cleaner.
Alternative would be to return something like `ZIO[R, (Chunk[W], E1), (Chunk[W], A)]`.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
sideeffffect: `S1 => ZIO[R, E, (S2, A)]` could work. But I think that the current way aligns better with the (my?) intuition behind ZPure (or a state monad in general), which is that it operates "on an mutable memory cell".
That may be a reason why @jdegoes has suggested that in the first place.

adamgfraser: Yeah. I guess I would think of `S1 => ZIO[R, E, (S2, A)]` as the definition of an effectual state transition function, which is exactly what `ZPure` is when lifted into `ZIO` and so it reflecting a very deep structure.

The `ZRef` also feels a little weird to me because we're not really using it as a mutable memory cell. We're just getting its result at the beginning, which we could easily get as an input, and getting its result at the end, which we could easily provide as an output. The real mutable memory cell lives inside the implementation of the `ZPure` run method which allocates the mutable memory and then uses it for potentially multiple state transitions.

The "get a value from a `ZRef` and set a value in a `ZRef`" can also be easily implemented in terms of that as `ref.get.flatMap(zPure.toZIO).tap { case (s, _) => ref.set(s) }` so the state transition function implementation seems more fundamental in some sense. What if you want to get the value from a queue and write it to a queue, or complete a promise with it, or throw it away?

adamgfraser: Also feels very procedural ("I'll provide a value to this function which it will mutate"). `ZPure` is embedding an imperative program in a functional context but feels unnecessarily so when we can just return the value and the caller can do whatever they want with it including writing it somewhere. The "mutate in place" pattern feels more like something I would see in very performance sensitive code.

adamgfraser: Also think ergonomics are going to be better. Otherwise you always have to create a `ZRef` to convert a `ZPure` value into a `ZIO` which you may not otherwise need.

adamgfraser: There is actually a further issue here. `S1` is the `get` type of the `ZRef` and `S2` is the `set` type. So if we lift a `ZPure` value into a `ZIO` and run it we now set the `ZRef` with the `S2` type but the `get` type is `S1` again. So as the user we now don't have any way to get the `S2` value returns by the effect! Furthermore to even build the polymorphic `ZRef` in the first place we need to provide a function `S2 => S1` which we never use and would be completely arbitrary or possibly not even have a meaningful implementation.

sideeffffect: @adamgfraser I'm convinced, `S1 => ...` seems like the way to go. But in that case, I think, we will have to discard the log. If somebody wants the log, she can run ZPure manually.

adamgfraser: We could have a variant `toZIOLogged` that returns the log in both the error and value channels.

sideeffffect: @jdegoes I've made the change. Could you also have a look at it?

adamgfraser (approved):
Looks good to me other than potentially adding a variant that exposes the log.

sideeffffect:
Variant with log added.
@jdegoes if you want to stick to ZRef, you still have a chance :smile:

adamgfraser (dismissed):

adamgfraser (approved):

PR #554, opened 2021-03-03
Add ZPure#refineOrDie

sideeffffect:
we can drop the custom NotNull once this will get released in ZIO: https://github.com/zio/zio/pull/4761


core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -877,6 +878,23 @@ object ZPure extends ZPureLowPriorityImplicits with ZPureArities {
   def accessM[R]: AccessMPartiallyApplied[R] =
     new AccessMPartiallyApplied
 
+  def fromEffectCatching[E >: Null <: Throwable]: FromEffectCatchingPartiallyApplied[E] =
```
adamgfraser, line 881: I think that implementing `refineOrDie` on `ZPure` so that we could do `fromEffect(???).refineOrDie[NumberFormatException]` would be a more compositional way of addressing this. This would also be consistent with the way this use case would be handled with ZIO.

sideeffffect: I tried to implement it, but we're not in ZIO. ZPure doesn't have a "die" error channel, so the only option would be to literally `throw` and that doesn't seem to me like a good idea :shrug:

adamgfraser: Maybe it should have a "die" channel??? There is an issue that we're still on a particular platform so let's say we are interpreting a `ZPure` value with an error type of `String` and we evaluate a continuation that throws an `IllegalArgumentException`. Right now we have nowhere to put that so the only thing we can do no is throw it, which isn't the end of the world but throws away any information we have about other errors that occurred as well as ability of higher levels of the application to handle it other than by explicitly using `try` and `catch`.

sideeffffect: :thinking: That's an interesting idea!

I haven't given it much thought, but I think ZPure could do well without "die channel". ZIO does need it, but ZPure isn't _"ZIO with more type parameters"_, so it could (should?) be fine without it. Maybe the best ZPure can do is to pretend "unchecked" exceptions don't exist. And if one is raised, which should be in a rare, truly serious and exceptional case, it will blow right through ZPure. This approach could make the design and intended use of ZPure more focused, which could be a good thing.

But from me, this is still just a hunch; no strong feeling either way. So if your guts tell you to go that way, go ahead. We can try it see how it works and perhaps even get some feedback from users :+1:

sideeffffect:
@adamgfraser :pray:

adamgfraser:
Why don't we have `attempt` that suspends a by name value that can fail and catches any `Throwable`, setting the error type to `Throwable`? Then we can have a `refineOrDie` variant that can throw at least until we figure out if `ZPure` should have a channel for that. That's the same behavior as what this is currently doing. That way we can avoid needing a `ClassTag` or using the partial application trick.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
jdegoes: I would err on the side of no channel for now. It's true this can result in the loss of the "primary error" (the `E`). Whereas, ZIO itself would hold onto the `E` and the `Throwable` via `Cause`. But ZPure is more likely to be used for short, pure computations, where exhaustive error reporting has less value; which is why I don't think adding stack traces, etc., is absolutely essential for ZPure (unlike ZIO).

jdegoes:
> Why don't we have `attempt` that suspends a by name value that can fail and catches any `Throwable`, setting the error type to `Throwable`? Then we can have a `refineOrDie` variant that can throw at least until we figure out if `ZPure` should have a channel for that. That's the same behavior as what this is currently doing. That way we can avoid needing a `ClassTag` or using the partial application trick.

Yes, this is great. `ZPure.attempt(...).refineToOrDie[IllegalStateException]`.

Also consistent with ZIO (2.0).

sideeffffect:
Awesome, thanks for the suggestion. Now ZPure has `refineToOrDie` in the style of ZIO.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -1277,3 +1297,16 @@ trait ZPureLowPriorityImplicits {
         ZPure.tupledPar(fa, fb)
     }
 }
+
+trait ZPureSyntax {
+
+  implicit final class ZPureRefineToOrDieOps[W, S1, S2, R, E <: Throwable, A](self: ZPure[W, S1, S2, R, E, A]) {
```
sideeffffect, line 1303: Could this be moved to ZPure trait as a regular method?

adamgfraser: I don't think so because we want `E` to appear as the upper bound in `E1 <: E` which is inconsistent with a covariant type parameter.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -495,6 +497,23 @@ sealed trait ZPure[+W, -S1, +S2, -R, +E, +A] { self =>
   final def provideState(s: S1): ZPure[W, Any, S2, R, E, A] =
     set(s) *> self
 
+  /**
+   * Keeps some of the errors, and terminates the fiber with the rest
```
adamgfraser, line 501: Maybe update the documentation here and below? This isn't necessarily even being run on a fiber and we aren't terminating it, just throwing the exception.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -936,6 +955,17 @@ object ZPure extends ZPureLowPriorityImplicits with ZPureArities {
...
     new AccessMPartiallyApplied
 
+  /**
+   * Constructs a computation, catching any `Throwable` that is thrown.
+   */
+  def attempt[S, A](a: => A): ZPure[Nothing, S, S, Any, Throwable, A] =
+    suspend {
+      try ZPure.succeed(a)
+      catch {
+        case NonFatal(e) => ZPure.fail(e)
```
adamgfraser, line 965: We would normally only throw if it is an instance of `VirtualMachineError`.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
adamgfraser: Why not just define in companion object instead of trait so will always automatically be in implicit scope?

core/shared/src/main/scala/zio/prelude/package.scala
```diff
@@ -48,7 +48,8 @@ package object prelude
     with OrdSyntax
     with PartialOrdSyntax
     with ForEachSyntax
-    with ZivariantSyntax {
+    with ZivariantSyntax
+    with fx.ZPureSyntax {
```
adamgfraser, line 52: Don't think we need this.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -1277,3 +1297,16 @@ trait ZPureLowPriorityImplicits {
...
 }
+
+trait ZPureSyntax {
+
+  implicit final class ZPureRefineToOrDieOps[W, S1, S2, R, E <: Throwable, A](self: ZPure[W, S1, S2, R, E, A]) {
+
+    /**
+     * Keeps some of the errors, and terminates the fiber with the rest.
+     */
+    def refineToOrDie[E1 <: E: ClassTag: CanFail](implicit ev: CanFail[E]): ZPure[W, S1, S2, R, E1, A] =
```
adamgfraser, line 1308: I don't think we are actually preventing any mistakes here.

sideeffffect: This is what happens if I don't supply the type parameter when using `refineToOrDie`
```
[error] /home/ondra/Projects/zio-prelude/core/shared/src/test/scala/zio/prelude/fx/ZPureSpec.scala:793:54: This error handling operation assumes your effect can fail. However, your effect has Nothing for the error type, which means it cannot fail, so there is no need to handle the failure. To find out which method you can use instead of this operation, please see the reference chart at: https://zio.dev/docs/can_fail
[error]               assert(ZPure.attempt(a.toString.toInt).refineToOrDie.runEither)(
[error]                                                      ^
```

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
sideeffffect: Oh, of course! Good idea :+1:

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
sideeffffect: Silly me... forgot to update comments after copy-paste :smile:

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
sideeffffect: Done. But why??

core/shared/src/main/scala/zio/prelude/package.scala
sideeffffect: removed :+1:

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
adamgfraser: I think minor disagreements over what should be treated as fatal and a desire to do as little work as possible when the system is already in an undefined state.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
adamgfraser: Sorry, I meant by requiring this I don't think we are preventing any user errors. I haven't seen any cases of users getting confused by not supplying an argument to this method, and if they didn't it would be equivalent to `orDie` which is perfectly well defined. So I feel like we're not actually preventing user errors and we are preventing you from refining the error type to `Nothing` which, while there is a more specific operator for it, seems like something you should be able to do with the more general operator as well.

sideeffffect: IMHO it would be better to have a dedicated `orDie` and use of `refineToOrDie` with mandatory type argument, but if you're sure this is better :+1:

adamgfraser (approved):
Let's get this merged!

PR #555, opened 2021-03-03
Bring back Validation

sideeffffect:
Validation was removed in this PR and replaced by a generalized ZPure
https://github.com/zio/zio-prelude/pull/461
However, I think we should bring `Validation` back, because it still has some advantages over ZPure.

Validation can be checked for equality, hashed, serialized and sent over network or stored in a DB. It is mere data and has all the advantages that come with it. On the other hand, ZPure is not data. And that's OK, it should not be -- it's an computation. But that means it just cannot be used for certain things.

We can have a `runValidation` method on ZPure, though, which will return all the errors (in contrast to `runEither`).

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/Assertions.scala
+++ b/core/shared/src/main/scala/zio/prelude/Assertions.scala
@@ -36,11 +36,9 @@ trait Assertions {
    * specified assertion.
    */
   def isFailureV[E](assertion: Assertion[E]): Assertion[Validation[E, Any]] =
-    Assertion.assertionRec("isFailureV")(param(assertion))(assertion) { validation =>
-      validation.either.run match {
-        case Left(e) => Some(e)
-        case _       => None
-      }
+    Assertion.assertionRec("isFailureV")(param(assertion))(assertion) {
+      case Validation.Failure(e) => Some(e.head)
+      case _                     => None
     }
 
   /**
@@ -76,10 +74,8 @@ trait Assertions {
    * specified assertion.
    */
   def isSuccessV[A](assertion: Assertion[A]): Assertion[Validation[Any, A]] =
-    Assertion.assertionRec("isSuccessV")(param(assertion))(assertion) { validation =>
-      validation.either.run match {
-        case Right(a) => Some(a)
-        case _        => None
-      }
+    Assertion.assertionRec("isSuccessV")(param(assertion))(assertion) {
+      case Validation.Success(a) => Some(a)
+      case _                     => None
     }
 }
--- a/core/shared/src/main/scala/zio/prelude/Gens.scala
+++ b/core/shared/src/main/scala/zio/prelude/Gens.scala
@@ -69,8 +69,8 @@ object Gens {
    * A generator of `Validation` values.
    */
   def validation[R <: Random with Sized, E, A](e: Gen[R, E], a: Gen[R, A]): Gen[R, Validation[E, A]] =
-    Gen.either(Gens.parSeq(Gen.empty, e), a).map {
-      case Left(es) => Validation.halt(es)
+    Gen.either(Gen.mapOf1(e, Gen.int(1, 9)), a).map {
+      case Left(es) => Validation.Failure(NonEmptyMultiSet.fromMapOption(es).get)
       case Right(a) => Validation.succeed(a)
     }
 }
--- a/core/shared/src/main/scala/zio/prelude/ParSeq.scala
+++ b/core/shared/src/main/scala/zio/prelude/ParSeq.scala
@@ -142,6 +142,12 @@ sealed trait ParSeq[+Z <: Unit, +A] { self =>
     flatMap(a => ParSeq.single(f(a)))
 
   /**
+   * Transforms a parSeq to a flat chunk of events.
+   */
+  final def toChunk[Z <: Unit, A](parSeq: ParSeq[Z, A]): zio.Chunk[A] =
+    parSeq.fold(zio.Chunk.empty, zio.Chunk.single)(_ ++ _, _ ++ _)
+
+  /**
    * Combines this collection of events with that collection of events to
    * return the Cartesian product of events, combining the elements into a
    * tuple.
--- a/core/shared/src/main/scala/zio/prelude/ParSeq.scala
+++ b/core/shared/src/main/scala/zio/prelude/ParSeq.scala
@@ -144,8 +144,8 @@ sealed trait ParSeq[+Z <: Unit, +A] { self =>
   /**
    * Transforms a parSeq to a flat chunk of events.
    */
-  final def toChunk[Z <: Unit, A](parSeq: ParSeq[Z, A]): zio.Chunk[A] =
-    parSeq.fold(zio.Chunk.empty, zio.Chunk.single)(_ ++ _, _ ++ _)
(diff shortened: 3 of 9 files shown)


adamgfraser:
I think it is a good question and pushes us to define more what `ZPure` is and is not.

There are definitely some pain points with the current encoding of `Validation` as a subtype of `ZPure` right now. I would actually put type inference with the state types higher on the list than whether it is data or not. So there would clearly be some immediate benefit in implementing it separately.

At the same time, this is a significant lack of unification and results in a lot of duplication, as we can just see by the number of lines of coded added by this PR. It probably also requires us to do more work in the future, for example if we do this I probably need to dust off my PR to upgrade `Validation` to `ZValidation` which we get for free with the current encoding.

So I think we should ask what the issues are and whether we think they are solvable or not.

I don't actually think whether it is data or not is a problem. If the `R` and `S1` types are `Unit` then we can treat it as data. And in fact we can serialize and deserializes data structures like this. For example the following data type passes ZIO's serialize and back test:

```scala
final case class Lazy[+A] private (run: () => A)

object Lazy {
  def make[A](a: => A): Lazy[A] =
    Lazy(() => a)
}
```

This does require that we not embed arbitrary side effecting functions in `ZPure`, which we have been a bit vague on to date, but I think that is a price worth paying, so I think I would delete the `fromEffect` constructors on `ZPure`.

The type inference issue is I think very real and we have made some progress towards this but have not solved it yet. And we need to solve it regardless for other uses cases such as `State`.

So I think I would say this is premature and we should clarify the contract around embedding effects in `ZPure` and work to address the type inference issues. If the type inference issues turn out to be unsolvable then we will have to revisit this but in that case we will have to do the same thing for `State` and I think we are pretty far away from that point.

sideeffffect:
I'm all for improving inference and prohibiting side effects in ZPure seems like a worthwhile idea. :+1: Let's suppose that we've solved the issue with type inference and there were no side effects involved.

Still, how would we check two ZPure for equality or hash them? ZPure is a reified computation, a state monad. Arbitrary computations, potentially time consuming, can be embedded in it. Would we have to run them every time we call `===`/`.hash`? How does a ZPure look in the debugger? ZPure is also complex -- it can do a lot, but we pay for it by it not being simple.
Validation on the other hand is just a value, a dead piece of data. And it's simple, easy to use and fits well into most of our user's needs. For anything advanced, people can go to ZPure (why would we need `ZValidation`, if people can use `ZPure`?). In the sense of _"simple things being simple and complex possible_".
IMHO we should distinguish between a computation and its result, a value. The distinction between a computation and a value is good and users will find it intuitive.

The code size comes mainly from the repetitive implementations of `mapParN` and `tupledPar`. We could avoid it by automation (like Magnolia or Shapeless). Or even better, using our own Type Classes. We shouldn't need those kinds of methods, since they're already implemented for `CommutativeBoth`. I think we should remove them. What do you think?

Btw, I think we will still need a function with `fromEffect`'s signature, so that users can wrap throwing code, like `"asf".toInt`, kind of like cat's `Validated` `catchNonFatal` and `catchOnly` do.

adamgfraser:
So I think if we went down this road we would say that `Equal` and `Hash` are only defined if `S1` is `Unit` and `R` is `Any`. Then `equal` and `hash` would amount to just evaluating a thunk, which is done for example in existing "data" types like `LazyList`. If the computations embedded are pure then that is always sound and at worse inefficient in some cases, and we could offer an operator like `materialize` on `Chunk` that evaluates the `ZPure` and stores its result to address that.

It is not the end of the world if we have `Validation` as a separate data type, we already wrote the code for it, but I think `ZValidation` kind of raises the problem that there really is significant duplication. If you're already using `Validation` to collect multiple errors you could easily also want to accumulate different non-fatal warnings. Imagine you are traversing a dependency graph and some things cause it to fail but other things are just warnings that you want to report to the user and can occur even if the overall traversal is successful. So now we need to build in a type for that, which we can do, in fact I already had a PR that added it, but it is something else that we get "for free" with `ZPure`.

Overall I see this as sparking a healthy discussion of what we want `ZPure` to be. I am proposing that `ZPure` model pure computations and for us to see if we can abstract over computations and data with it. There is definitely another way we can go where we make more of a distinction between data and computations as you are suggesting and that would potentially push us in the direction of `ZPure` also modeling synchronous effects, at least based on user preference, and being something more like "pure or synchronous computations".

Definitely trade-offs with each so worth thinking about.

sideeffffect:
@adamgfraser this is really making me think! :smile: 

My take:

## ZPure
 * A reified computation. A state monad on steroids. Something that may sometime in a future produce a value.
 * Only for pure computations.
 * Suitable for complex algorithms, typically in business logic, that would otherwise require mutating variable while imperatively operating
   * Creating a big ZPure computation, and then running it to get ZValidation that will be worked with later could be a common workflow
 * Doesn't need Equal/Hash/Debug/.toString implementation
 * I would be worried that memoizing `materialize` could do more harm than good (even though the computations are designed to be pure).
 * `fromEffect` renamed to something that suggests it's for catching exceptions, but not for IO. `catchNonFatal`?
 * `runAll` can still return `(Chunk[W], Either[Cause[E], (S2, A)])`
 * `runValidation` will return ZValidation

## ZValidation
 * Just Data, an already computed value.
 * (One of the possible) result of a ZPure
 * On its own, suitable for simpler tasks, for example input data validation/decoding at the outer layer of an application.
 * Has nice Equal/Hash/Debug/.toString and so is easy to manipulate and view in debugger.

```scala
sealed abstract case class ZValidation[+W, +E, +A](log: Chunk[W])

object ZValidation {
  final case class Failure[+W, +E](override val log: Chunk[W], errors: NonEmptyMultiSet[E]) extends ZValidation[W, E, Nothing](log)
  final case class Success[+W, +A](override val log: Chunk[W], value: A) extends ZValidation[W, Nothing, A](log)
}

type Validation[+E, +A] = ZValidation[Nothing, E, A]
val Validation: ZValidation.type = ZValidation
```

We fight duplication and code repetition and boilerplate with relying on our Type Classes and extension methods as much as possible.

I'll try to thing about this again tomorrow.

adamgfraser:
I think that is a very reasonable approach to take. I think we could also go the other way as outlined above but I agree having it be "lazy" unless memoized is not ideal. I would like to get some more views on this but in the meantime I will resubmit my `ZValidation` PR and get it ready to go so if we do go in direction we're not at risk for any loss of generality versus the current version.

poslegm:
I am strongly agree with @sideeffffect. 

Today I tried to validate simple case class fields with current `ZValidation` and it was **very** hard. `ZPure` generic and complex API is _too rich_ for simple validation cases. I spent a lot of time to find methods for composition with errors accumulation and running it right way to get my `Either[Cause[String], T]`. 

I'll never make my team to validate data if they will need to use `ZPure` :smile:

Now `ZValidation` forces programmer to understand the difference between monadic and applicative compositions. But I think that it should be simple and validation-specific DSL without anything unnecessary.

adamgfraser:
@poslegm  I am coming to the same conclusion.

jdegoes:
Agreed on bringing back ZValidation. Lack of duplication (for library authors) cannot trump usability (for end users).

adamgfraser:
Done in #558 .

PR #558, opened 2021-03-04
Implement ZValidation

adamgfraser:
Updated version of #233 if we decide we want we do not want to have `ZValidation` be a subtype of `ZPure`.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -1,0 +1,1433 @@
+package zio.prelude
+
+import zio.prelude.ZValidation._
+import zio.test.Assertion
+import zio.{Chunk, IO, ZIO}
+
+import scala.util.Try
+
+/**
+ * `ZValidation` represents either a success of type `A` or a collection of one
+ * or more errors of type `E` along with in either case a log with entries of
+ * type `W`. Unlike `Either`, `ZValidation` does not "short circuit" on
+ * failures and instead allows accumulating multiple errors. This can be
+ * particularly useful in validating data, where we want to attempt to validate
+ * all of the data and retain information about all errors that arose, rather
+ * than failing at the first error.
+ */
+sealed trait ZValidation[+W, +E, +A] { self =>
+
+  /**
+   * A symbolic alias for `zipParLeft`.
+   */
+  final def <&[W1 >: W, E1 >: E, B](that: ZValidation[W1, E1, B]): ZValidation[W1, E1, A] =
+    zipParLeft(that)
+
+  /**
+   * A symbolic alias for `zipParRight`.
+   */
+  final def &>[W1 >: W, E1 >: E, B](that: ZValidation[W1, E1, B]): ZValidation[W1, E1, B] =
+    zipParRight(that)
+
+  /**
+   * A symbolic alias for `zipPar`.
+   */
+  final def <&>[W1 >: W, E1 >: E, B](that: ZValidation[W1, E1, B]): ZValidation[W1, E1, (A, B)] =
+    zipPar(that)
+
+  /**
+   * A symbolic alias for `log`.
+   */
+  final def ??[W1 >: W](w1: W1): ZValidation[W1, E, A] =
+    log(w1)
+
+  /**
+   * Returns whether this `ZValidation` and the specified `ZValidation` are
+   * equal to each other.
+   */
+  override final def equals(that: Any): Boolean =
+    (self, that) match {
+      case (Failure(_, e), Failure(_, e1)) => e == e1
+      case (Success(_, a), Success(_, a1)) => a == a1
+      case _                               => false
+    }
+
+  /**
+   * Transforms the value of this `ZValidation` with the specified validation
+   * function if it is a success or returns the value unchanged otherwise.
+   */
+  final def flatMap[W1 >: W, E1 >: E, B](f: A => ZValidation[W1, E1, B]): ZValidation[W1, E1, B] =
+    self match {
+      case Failure(w, e) => Failure(w, e)
+      case Success(w, a) =>
+        f(a) match {
+          case Failure(w1, e) => Failure(w ++ w1, e)
+          case Success(w1, b) => Success(w ++ w1, b)
+        }
+    }
+
+  /**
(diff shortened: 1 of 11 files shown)


core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -0,0 +1,1433 @@
...
+/**
+ * `ZValidation` represents either a success of type `A` or a collection of one
+ * or more errors of type `E` along with in either case a log with entries of
+ * type `W`. Unlike `Either`, `ZValidation` does not "short circuit" on
+ * failures and instead allows accumulating multiple errors. This can be
+ * particularly useful in validating data, where we want to attempt to validate
+ * all of the data and retain information about all errors that arose, rather
+ * than failing at the first error.
+ */
+sealed trait ZValidation[+W, +E, +A] { self =>
```
sideeffffect, line 18: I could be worthwhile to put `log` on `ZValidation`, since it's supposed to be available for any case.
```suggestion
sealed abstract case class ZValidation[+W, +E, +A](log: Chunk[W]) { self =>
```

core/shared/src/test/scala/zio/prelude/fx/ZPureSpec.scala
```diff
@@ -842,35 +842,7 @@ object ZPureSpec extends DefaultRunnableSpec {
               )
             }
           )
-        ),
-        test("parallel errors example") {
```
sideeffffect, line 846: Do these tests even need to be removed? I'm fine with keeping the current functionality in ZPure :wink:

sideeffffect (dismissed):
I've had a few thoughts, but looks great! :+1:

core/shared/src/main/scala/zio/prelude/ZValidation.scala
adamgfraser: I added a `getLog` operator since we already use `log` for appending to the log.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -85,6 +85,15 @@ sealed trait ZValidation[+W, +E, +A] { self =>
       case Success(_, a) => success(a)
     }
 
+  /**
+   * Returns the value of the log.
+   */
+  final def getLog: Chunk[W] =
```
sideeffffect, line 91: `log`? :smile_cat:

core/shared/src/test/scala/zio/prelude/fx/ZPureSpec.scala
adamgfraser: Good point! Added them back.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
adamgfraser: We already use `log` for writing to the log. We could overload it. I thought that might be a little confusing but I definitely thought about it.

adamgfraser: Well those would be options so I think we already have `toOption` which does the same thing got the value channel and then you could fold for the error case or maybe we could have `errorsOption` or something.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
sideeffffect: Oh, OK then :+1:

core/shared/src/main/scala/zio/prelude/ZValidation.scala
sideeffffect: Scratch that :point_up: I hadn't read your comment above.

sideeffffect (approved):

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -0,0 +1,1442 @@
...
+      case Failure(w, e) => (w, Left(e))
+      case Success(w, a) => (w, Right(a))
+    }
+
+  /**
+   * Transforms this `ZValidation` to an `Either`, discarding the log.
+   */
+  final def toEither[E1 >: E]: Either[NonEmptyMultiSet[E1], A] =
+    fold(Left(_), Right(_))
+
```
poslegm, line 152: When use `cats.data.Validated` I often accumulate errors into one exception at the end of validation. Because work with `Either[MyDataValidationException, MyData]` is often simpler than with `Either[NonEmptyMultiSet[String], Mydata]` or `Either[NonEmptyChain[String], Mydata]`.

What about `squash` method?

```scala
def squash[E1](f: String => E1)(implicit D: Debug[E]): Either[E1, A] =
  fold(errors => Left(f(errors.toSet.map(D.debug).mkString("; "))), Right(_))

final case class ValidationErrors(errors: String) extends Exception(s"validation failed with [$errors]")
def squash(implicit D: Debug[E]): Either[ValidationErrors, A] =
  sqash(ValidationErrors(_))
```

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -0,0 +1,1442 @@
...
+  /**
+   * Constructs a `ZValidation` from a value and an assertion about that value.
+   * The resulting `ZValidation` will be a success if the value satisfies the
+   * assertion or else will contain a string rendering describing how the
+   * value did not satisfy the assertion.
+   */
+  def fromAssert[A](value: A)(assertion: Assertion[A]): Validation[String, A] =
+    if (assertion.test(value)) succeed(value)
+    else fail(s"$value did not satisfy ${assertion.render}")
+
```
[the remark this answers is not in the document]
adamgfraser: @poslegm Added `fromPredicateWith` in #573. Let me think about the best way to address the second use case.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
adamgfraser: Yes. Maybe something more generic that let's you squash the errors with a function. Will work on this.

sideeffffect: I don't if this is what you had in mind, but I've made https://github.com/zio/zio-prelude/pull/574

PR #560, opened 2021-03-05
Work around bad type inference of our extension methods

sideeffffect:
current master would fail with this
```scala
  val x: Either[Int, String] = Either.cond(test = true, "x", 1)
  val y: Right[Int, String]  = Right("y")

  val _ = x <> y // OK
  val _ = y <> x // doesn't work, but will be fixed
  val _ = x <*> y // OK
  val _ = y <*> x // doesn't work, and won't be fixed, see below
```

## Scala 2.11 and 2.12 and Higher kinds

While Scala 2.13 works fine with `F1[x] >: F[x]`, 2.11 and 2.12 don't like it. If I expand `fa.zipWith(fb)(_ zip _)` in `AssociativeBoth.compose`, I get
```scala
        AssociativeBothCovariantOps[F, G[A]](fa).zipWith[F, G[B], G[(A, B)]](fb)({ case (ga, gb) =>
          AssociativeBothOps[G, A](ga).zip[G, B](gb)
        })
```
For 2.12, the compiler complains
```
type arguments [G,B] do not conform to method zip's type parameter bounds [F1[x] >: G[x],B]
           AssociativeBothOps[G, A](ga).zip[G, B](gb)
                                           ^
```
Seems like we'll have to do this fix only for `*` kind types :crying_cat_face:

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/Associative.scala
+++ b/core/shared/src/main/scala/zio/prelude/Associative.scala
@@ -1336,13 +1336,13 @@ trait AssociativeSyntax extends PlatformSpecificAssociativeSyntax {
     /**
      * A symbolic alias for `combine`.
      */
-    def <>(r: => A)(implicit associative: Associative[A]): A =
+    def <>[A1 >: A](r: => A1)(implicit associative: Associative[A1]): A1 =
       associative.combine(l, r)
 
     /**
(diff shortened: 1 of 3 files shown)


adamgfraser (approved):
Nice!

adamgfraser (approved):

PR #565, opened 2021-03-08
Debug interpolator

sideeffffect:


sideeffffect:
@adamgfraser :pray:

core/shared/src/main/scala/zio/prelude/Debug.scala
```diff
@@ -792,4 +795,8 @@ trait DebugSyntax {
   implicit class DebugOps[A](self: A) {
     def debug(implicit debug: Debug[A]): Debug.Repr = debug.debug(self)
   }
+
+  implicit final class DebugInterpolator(_sc: StringContext) {
+    def dbg(args: Debug.Repr*): String = _sc.s(args.map(_.toString): _*)
```
adamgfraser, line 800: Should we make this `d` so it feels more like a native interpolator?

core/shared/src/test/scala/zio/prelude/DebugSpec.scala
```diff
@@ -192,6 +192,16 @@ object DebugSpec extends DefaultRunnableSpec {
         testM("testTrait")(check(genTestTrait) { c =>
           assert(c.debug.render(Renderer.Full))(equalTo(s"DebugSpec.${c.getClass.getSimpleName.init}"))
         })
+      ),
+      suite("DebugInterpolator")(
```
adamgfraser, line 196: Should we have a test whether the `Debug` rendering is different from `toString`?

adamgfraser (approved):
Great idea! Just a couple of questions.

core/shared/src/main/scala/zio/prelude/Debug.scala
sideeffffect: I also considered it... Plain Scala uses `s"..."` for `toString` interpolation and cats use `show"..."` for `show`. To me, `debug"..."` felt way too long and `d"..."` could be ambiguous or unclear to some people ("has this something to do with _decimal_, or what?").
I'm open to changing it, just unsure what would people like most.

core/shared/src/test/scala/zio/prelude/DebugSpec.scala
sideeffffect: `dbg"..."` actually does behave differently from `s"..."` in some cases, because it uses `Renderer.Simple` (and not `Renderer.Scala`).

adamgfraser: Yes definitely. I'm just wondering we should have a test where the `Debug` rendering is different from the `toString` rendering so we can verify that we actually getting the result of `Debug`.

core/shared/src/test/scala/zio/prelude/DebugSpec.scala
```diff
@@ -192,6 +192,16 @@ object DebugSpec extends DefaultRunnableSpec {
...
+        test("String") {
+          val name = "World"
+          assert(dbg"Hello, my name is $name.")(equalTo("Hello, my name is \"World\"."))
```
sideeffffect, line 199: :point_up: this is the test :smile_cat: `s"..."` doesn't add `"`s around strings, for example.

adamgfraser: Ah, too subtle for me! 😃

core/shared/src/main/scala/zio/prelude/Debug.scala
adamgfraser: Did you want to do the `d` here?

sideeffffect: Oh, I thought you're fine with `dbg`. I can change it to `d` if you're sure it's better :+1:

adamgfraser: There is definitely subjectivity here. It seems like people preferred the `d` on Discord.

sideeffffect: I had made a survey, hadn't I :facepalm: 
:smile_cat: 
PR on the way!

PR #566, opened 2021-03-08
Add forEachUnit

sideeffffect:
Useful if you want to make sure you're not throwing away values -- often a mistake when practicing pure FP

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ForEach.scala
+++ b/core/shared/src/main/scala/zio/prelude/ForEach.scala
@@ -75,6 +75,13 @@ trait ForEach[F[+_]] extends Covariant[F] { self =>
     forEach(fa)(identity)
 
   /**
+   * Converts a collection with elements that are in the context of effects to
+   * a collection of elements in the context of an effect.
+   */
+  def flip_[G[+_]: IdentityBoth: Covariant](fa: F[G[Unit]]): G[Unit] =
+    forEach_(fa)(identity)
+
+  /**
    * Folds over the elements of this collection using an associative operation
    * and an identity.
    */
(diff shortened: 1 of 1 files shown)


adamgfraser:
I'm worried we're providing too many ways to do the same thing here between this, `ForEach`, and `ForEach_`. I think we have a pretty clear message right now:

1. If you don't care about the return type use `ForEach_`. This will both allow a more efficient implementation and conveys intent.
2. Otherwise use `ForEach`

I'm not sure what `ForEachUnit` adds to this story and a message with three parts is inherently going to be more complicated than one with only two.

jdegoes:
I agree to keep it simple. Also we could always add implicits to the other ones, if we want to warn about ignoring an effect in there.

sideeffffect:
> too many ways to do the same thing here between this, `forEach`, and `forEach_`

Then I would suggest to change `forEach_`. Currently, `forEach_` does two things at once, which is not compositional and makes reading code (and inferring author's intent) difficult:
 * "discards" the return value from `f` (as if turning `f: A -> G[B]` into `f: A -> G[Unit]`)
 * returns G[Unit] (instead of G[F[Unit]])

I propose that `forEach_` only accepts `f: A -> G[Unit]` and returns `G[Unit]`. If the user has `f: A -> G[X]`, she can pass in `f.ignore` to make the decision to discard the returned values explicit.

Going forward with this design, we're making `forEach_` designated specifically for functions whose only purpose is to perform a side effect. It also has the benefit of unintentionally discarding important values impossible.

adamgfraser:
@sideeffffect I'm sorry maybe I'm not getting it but I'm not seeing the meaningfulness of the distinction between those two things. If I use `foreach_` I am saying I want to discard the return value of the effectual function so at that point it doesn't really matter whether I have a list of discarded values or a single discarded values. If I don't want to discard the values I use `foreach` and get them all back.

To use a concrete example, I think a very reasonable thing to do would be something like:

```scala
ZIO.foreach_(promises)(promise => promise.succeed(value))
```

This is "discarding" results in that `complete` returns a `Boolean` indicating whether the `Promise` was already completed but in many cases I don't care about that so I use `foreach_` because I just want to make sure it is done, I don't care about the result. I get back a `UIO[Unit]` but there wouldn't really be any point in getting a `UIO[List[Unit]]`. I don't think there is really a risk that I accidentally discard these values because this is the operator that is explicitly for discarding values.

If I do want to know whether the promises were already completed I can use `foreach`. I'm not really seeing what the value of this third operator would be or what the thing is we're trying to avoid.

jdegoes:
Like Adam says, there is really only one choice here: whether to ignore the success value of the provided function. If you ignore it, then the result must be ignored, because `List[Any]` / `List[Unit]` isn't useful. So there really are only 2 variants: collect all the results, or ignore them. These correspond to `forEach` and `forEach_`.

PR #567, opened 2021-03-09
Scala 3.0.0-RC1

vigoo:


sideeffffect:
Thanks @vigoo for such a quick update!
Could you also change the Mergify conditions in this PR? https://github.com/zio/zio-prelude/blob/master/.mergify.yml#L11

@adamgfraser / @mijicd we would need to change the _Branch protection settings_ to be able to merge this PR (discard `test (*, 3.0.0-M3, *)` and start requiring `test (*, 3.0.0-RC1, *)`)

But it seems like we've hit a bug in Dotty. Have you seen anything like it? I guess we would have to report it, unless it has been reported before.

adamgfraser:
Yes we can definitely make those changes but the compiler is crashing so it is hard to even know what if anything we need to change. We need to report it and may have to wait for the next release candidate unless there is a workaround. I will open an issue for it.

vigoo:
Yes I just opened the PR anyway to make it easier to show the compiler issue :)

adamgfraser:
Yes definitely good to submit a PR even though it is not working. I just submitted an issue for this.

adamgfraser:
Looks like a fix for this just got merged. May be good to try to compile on the latest nightly to confirm that there are no other issues before the next release candidate.

sideeffffect:
Wouldn't we need to compile with NIGHTLY all the dependencies like izumi-reflect and zio itself?

adamgfraser:
Ah that's a good point.

PR #572, opened 2021-03-14
fromPredicate* arguments switched for better type inference

poslegm:
Now I need to write
```scala
ZValidation.fromPredicateWith(s"ttl [$ttl] must be less than 30 days")((x: FiniteDuration) => x < 30.days)(ttl)
```

but I want to write
```scala
ZValidation.fromPredicateWith(s"ttl [$ttl] must be less than 30 days")(ttl)(_ < 30.days)
```


adamgfraser (approved):
Good idea!

PR #573, opened 2021-03-14
Add Additional Operators To ZValidation

adamgfraser:


sideeffffect (approved):
A missing piece :+1:

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -346,6 +346,19 @@ object ZValidation extends LowPriorityValidationImplicits {
...
+  /**
+   * Constructs a `Validation` from a predicate, failing with None.
+   */
+  def fromPredicate[A](value: A)(f: A => Boolean): Validation[None.type, A] =
+    fromPredicateWith(None)(value)(f)
+
+  /**
+   * Constructs a `Validation` from a predicate, failing with the error provided.
+   */
+  def fromPredicateWith[E, A](error: E)(value: A)(f: A => Boolean): Validation[E, A] =
```
poslegm, line 358: I just noticed that here are used call-by-value parameters. Maybe make them lazy?
```scala
def fromPredicateWith[E, A](error: => E)(value: => A)(f: A => Boolean):
```

sideeffffect: I think that could makes sense. Do you have any specific use case in mind?
Do you think you could open a PR for this (and the other function)?

poslegm: https://github.com/zio/zio-prelude/pull/579

> Do you have any specific use case in mind?

Value always used in `if` condition, but error needed only if condition failed. Error often contains string interpolation. It's minor perfomance improvement, but we get it for free

PR #574, opened 2021-03-14
Add ZValidation#{mapErrorAll, mapLogAll, toEitherWith}

sideeffffect:
You save typing 5 characters. (And maybe get better type inference?)

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -124,6 +124,16 @@ sealed trait ZValidation[+W, +E, +A] { self =>
     }
 
   /**
+   * Transforms all the error values of this `ZValidation` with the specified
+   * function.
+   */
+  final def mapErrorAll[E2](f: NonEmptyMultiSet[E] => NonEmptyMultiSet[E2]): ZValidation[W, E2, A] =
+    self match {
+      case Failure(w, e) => Failure(w, f(e))
+      case Success(w, a) => Success(w, a)
+    }
+
+  /**
    * Transforms the log entries of this `ZValidation` with the specified
    * function.
    */
@@ -131,6 +141,16 @@ sealed trait ZValidation[+W, +E, +A] { self =>
     self match {
       case Failure(w, e) => Failure(w.map(f), e)
       case Success(w, a) => Success(w.map(f), a)
+    }
+
+  /**
+   * Transforms all the log entries of this `ZValidation` with the specified
+   * function.
+   */
+  final def mapLogAll[W2](f: Chunk[W] => Chunk[W2]): ZValidation[W2, E, A] =
+    self match {
+      case Failure(w, e) => Failure(f(w), e)
+      case Success(w, a) => Success(f(w), a)
     }
 
   /**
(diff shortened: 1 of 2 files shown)


adamgfraser (approved):

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -150,6 +170,12 @@ sealed trait ZValidation[+W, +E, +A] { self =>
   final def toEither[E1 >: E]: Either[NonEmptyMultiSet[E1], A] =
     fold(Left(_), Right(_))
 
+  /**
+   * Transforms this `ZValidation` to an `Either`, transforming the accumulated errors and discarding the log.
+   */
+  final def toEitherWith[E1 >: E, E2](f: NonEmptyMultiSet[E1] => E2): Either[E2, A] =
```
sideeffffect, line 176: Btw, is `E1 >: E` the correct thing to do here? I'm never sure with this :laughing:

adamgfraser: No, if `E` just appears as an input into a function like here you can use `E` and it can avoid type inference issues in some cases.

sideeffffect: Fixed. Thanks!

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -235,7 +261,7 @@ object ZValidation extends LowPriorityValidationImplicits {
   /**
    * Derives an `Equal[ZValidation[W, E, A]]` given an `Equal[A]`.
    */
-  implicit def ZValidationEqual[W, E: Equal, A: Equal]: Equal[ZValidation[W, E, A]] =
+  implicit def ZValidationEqual[W, E, A: Equal]: Equal[ZValidation[W, E, A]] =
     Equal.make {
       case (Failure(_, e), Failure(_, e1)) => e == e1
```
adamgfraser, line 266: `===` here, right?

sideeffffect: Welp, we're comparing two `NonEmptyMultiSet`s, which are just a `Map[X, Int]` inside. From that point of view, `===` is more elegant, but unnecessary ceremony possibly even with some performance overhead.

But if you still think, `===` would be better, I'm not against, I can put it there.

adamgfraser: Ah, sorry, I thought you were adding the `Equal` construct there. Yes, it doesn't really make sense since we don't have a way to compare equality on the keys with a custom notion of equality efficiency so no reason to pretend that we do.

adamgfraser (approved):

core/shared/src/main/scala/zio/prelude/ZValidation.scala
sideeffffect: Exactly. Sometimes I wish we just had "our own" ZHashMap and ZOrdMap which would be with proper variance and used our Type classes, like `Hash` and `Ord`...

adamgfraser: Yeah. That would be a significant project. I'm not sure it is worth it since those are complex data structures where it would be easy to introduce bugs or performance regressions and I don't think there is really a benefit outside of this but I have thought the same thing.

PR #575, opened 2021-03-16
Use NonEmptyChunk in ZValidation

adamgfraser:
We want to honor the guarantee from `CommutativeBoth` that:

```scala
ZVallidation.fail("foo") <&> ZValidation.fail("bar") === ZValidation.fail("bar") <&> ZValidation.fail("foo")
```

However, from the user perspective it can be helpful to be able to see the order in which errors occurred for debugging purposes.

This PR addresses that by deferring the conversion of the errors into a multiset until we actually call `equals` or `hashCode`. This approach of "lazily" honoring these guarantees is similar to what we do in other ZIO data structures such as `Cause` in ZIO and `BoolAlgebra` in ZIO Test.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/Assertions.scala
+++ b/core/shared/src/main/scala/zio/prelude/Assertions.scala
@@ -16,6 +16,7 @@
 
 package zio.prelude
 
+import zio.NonEmptyChunk
 import zio.test.Assertion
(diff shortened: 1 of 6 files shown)


jdegoes:
Great idea!

jdegoes (approved):

sideeffffect (approved):

PR #576, opened 2021-03-16
Implement ZSet#forEach

adamgfraser:
Implements `ForEach` for `ZSet`. This is only defined when the `ZSet` is a multiset, as we can only perform an effect an integral number of times.

Doing this in a way that was sound also required implementing a `Natural` new type so that we know whether the `ZSet` contains only non-negative values since a `ZSet` with negative integer values is also well defined but has different semantics.

In the course of doing this I also noticed that the definition of laws and type class instance as `val` could lead to deadlocks due to early initialization issues. I changed all the laws to be `lazy val` and am hopeful that this will improve the flakiness we have seen in CI. I think we should consider doing the same with type class instances.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/Ord.scala
+++ b/core/shared/src/main/scala/zio/prelude/Ord.scala
@@ -128,7 +128,7 @@ object Ord extends Lawful[Ord] {
    * For all values `a1` and `a2`, `a1` is less than or equal to `a2` or `a2`
    * is less than or equal to `a1`.
    */
-  val connexityLaw1: Laws[Ord] =
+  lazy val connexityLaw1: Laws[Ord] =
     new Laws.Law2[Ord]("connexityLaw1") {
       def apply[A: Ord](a1: A, a2: A): TestResult =
         (a1 lessOrEqual a2) || (a2 lessOrEqual a1)
@@ -138,7 +138,7 @@ object Ord extends Lawful[Ord] {
    * For all values `a1` and `a2`, `a1` is greater than or equal to `a2` or
    * `a2` is greater than or equal to `a1`.
    */
-  val connexityLaw2: Laws[Ord] =
+  lazy val connexityLaw2: Laws[Ord] =
     new Laws.Law2[Ord]("connexityLaw2") {
       def apply[A: Ord](a1: A, a2: A): TestResult =
         (a1 greaterOrEqual a2) || (a2 greaterOrEqual a1)
@@ -148,7 +148,7 @@ object Ord extends Lawful[Ord] {
    * For all values `a1` and `a2`, `a1` is less than or equal to `a2` if and
    * only if `a2` is greater than or equal to `a1`.
    */
-  val complementLaw: Laws[Ord] =
+  lazy val complementLaw: Laws[Ord] =
     new Laws.Law2[Ord]("complementLaw") {
       def apply[A: Ord](a1: A, a2: A): TestResult =
         (a1 lessOrEqual a2) <==> (a2 greaterOrEqual a1)
@@ -157,7 +157,7 @@ object Ord extends Lawful[Ord] {
   /**
    * The set of all laws that instances of `Ord` must satisfy.
    */
-  val laws: Laws[Ord] =
+  lazy val laws: Laws[Ord] =
     connexityLaw1 +
       connexityLaw2 +
       complementLaw +
--- a/core/shared/src/main/scala/zio/prelude/newtypes/package.scala
+++ b/core/shared/src/main/scala/zio/prelude/newtypes/package.scala
@@ -119,35 +119,20 @@ package object newtypes {
     def successor(n: Natural): Natural =
       Natural(n + 1)
 
-    /**
-     * The `Commutative` and `Identity` instance for the product of `Natural` values.
-     */
-    implicit val NaturalProdCommutativeIdentity: Commutative[Prod[Natural]] with Identity[Prod[Natural]] =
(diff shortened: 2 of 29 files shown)


core/shared/src/main/scala/zio/prelude/newtypes/package.scala
```diff
@@ -107,4 +107,51 @@ package object newtypes {
   object FailureOut extends NewtypeF
 
   type FailureOut[+A] = FailureOut.Type[A]
+
+  object Natural extends SubtypeSmart[Int](isGreaterThanEqualTo(0)) {
```
sideeffffect, line 111: A type for integers in Scala is simply called `Int`. So why don't we go with this convention and call this `Nat`?
Not a big deal, but the shorter name could nudge more people to use it (which they should!).

core/shared/src/main/scala/zio/prelude/newtypes/package.scala
```diff
@@ -107,4 +107,51 @@ package object newtypes {
...
+    val zero: Natural =
+      Natural(0)
+
+    def successor(n: Natural): Natural =
+      Natural(n + 1)
+
+    /**
+     * The `Commutative` and `Identity` instance for the product of `Natural` values.
+     */
+    implicit val NaturalProdCommutativeIdentity: Commutative[Prod[Natural]] with Identity[Prod[Natural]] =
```
sideeffffect, line 125: Since `Nat` is not a class/trait, it can't have a companion object and so we have to put Type class implementations to the companion objects of the Type classes themselves.

core/shared/src/main/scala/zio/prelude/Ord.scala
```diff
@@ -157,7 +157,7 @@ object Ord extends Lawful[Ord] {
   /**
    * The set of all laws that instances of `Ord` must satisfy.
    */
-  val laws: Laws[Ord] =
+  lazy val laws: Laws[Ord] =
```
sideeffffect, line 160: Could `lazy val laws` be enough? Do we need to make the individual laws lazy too? The individual laws could be private for all we care, no?
`laws` on the other hand have a dependency on each other in the hierarchy, though.
Just thinking out loud...

core/shared/src/test/scala/zio/prelude/ZSetSpec.scala
```diff
@@ -43,11 +49,12 @@ object ZSetSpec extends DefaultRunnableSpec {
             // Scala 2.11 doesn't seem to be able to infer the type parameter for CovariantDeriveEqual.derive
             CovariantDeriveEqual.derive[({ type lambda[+x] = ZSet[x, Int] })#lambda](
               ZSetCovariant(IntSumCommutativeInverse),
-              ZSetDeriveEqual(IntHashOrd)
+              ZSetDeriveEqual(IntHashOrd, Identity[Sum[Int]])
             ),
             IntHashOrd
           )
         ),
+        // testM("foreach")(checkAllLaws(ForEach)(genFZSet(naturals), Gen.anyInt)),
```
sideeffffect, line 57: commented test?

sideeffffect (changes requested):
Terrific job! Haven't realized we can make ForEach for MultiSet. Much :heart: `Nat` (or `Natural` if you prefer). Just a few suggestions

core/shared/src/test/scala/zio/prelude/ZSetSpec.scala
adamgfraser: Good catch! Fixed now. Thanks!

core/shared/src/main/scala/zio/prelude/Ord.scala
adamgfraser: I think for our purposes here yes but contributors or users could potentially want to test an individual law and could run into the same problem. Maybe erring a bit on the side of caution but I think the cost is minimal (if a `val` versus a `lazy val` is relevant to your performance you shouldn't be using this level of abstraction at all) and there can be really hard bugs to track down.

core/shared/src/main/scala/zio/prelude/newtypes/package.scala
adamgfraser: Yes. I thought we had that functionality with our new types but looks like not yet. I moved it now.

core/shared/src/main/scala/zio/prelude/newtypes/package.scala
adamgfraser: I'm fine either way. I think this is what we have used in our examples in the documentation but have also seen lots of sources that use `Nat`. The other consideration here is I think right now unfortunately this is not a great user facing data type because the definitions of the arithmetic operators on `Int` shadow the extension methods we would like to define.

jdegoes (approved):

core/shared/src/main/scala/zio/prelude/Ord.scala
sideeffffect: makes sense :+1:

sideeffffect (approved):
I have to admit that `Nat` feels more _natural_ (:upside_down_face:) to me, but I think this is ready to be merged :+1:

adamgfraser:
Let's get this in now and totally open to changing the name if people feel like it is too long. I think we are going to have to figure out the operators before users can really work with it anyway. Maybe we need to switch to `NewtypeSmart` instead of `SubtypeSmart`? Would be a bit of a shame since this seems like the kind of use case the `Subtype` functionality was designed for.

PR #577, opened 2021-03-16
Add helper methods to NonEmpty* our collections

sideeffffect:


core/shared/src/main/scala/zio/prelude/NonEmptyList.scala
```diff
@@ -78,6 +78,21 @@ sealed trait NonEmptyList[+A] { self =>
       if (seen(a)) (as, seen) else (cons(a, as), seen + a)
     }._1.reverse
 
+  /** Decomposes the `NonEmptyList` into an element and a (possibly empty) `List` */
+  final def destruct: (A, List[A]) = self match {
```
adamgfraser, line 82: Isn't this just pattern matching on the list?

core/shared/src/main/scala/zio/prelude/NonEmptyList.scala
```diff
@@ -78,6 +78,21 @@ sealed trait NonEmptyList[+A] { self =>
...
+    case Single(head)     => (head, List())
+    case Cons(head, tail) => (head, tail.toList)
+  }
+
+  /**
+   * Decomposes the `NonEmptyList` either into an element and the remaining `NonEmptyList`,
+   * or just a single element, if there aren't any other.
+   */
+  final def destructEither: Either[A, (A, NonEmptyList[A])] = self match {
```
adamgfraser, line 91: See above.

core/shared/src/main/scala/zio/prelude/NonEmptyList.scala
```diff
@@ -620,3 +635,23 @@ trait LowPriorityNonEmptyListImplicits {
...
     PartialOrd[List[A]].contramap(_.toList)
 }
+
+trait NonEmptyListSyntax {
+  implicit class NonEmptyListListOps[A](self: List[A]) {
+
+    /**
+     * Converts to a `NonEmptyList` or `None` if empty.
+     */
+    def toNonEmptyList: Option[NonEmptyList[A]] = self match {
```
adamgfraser, line 645: Implementation can just be `fromIterableOption`.

core/shared/src/main/scala/zio/prelude/NonEmptySet.scala
```diff
@@ -83,6 +95,11 @@ final class NonEmptySet[A] private (private val set: Set[A]) { self =>
   /** Removes the `elem` from this `NonEmptySet`. Alias for `-`. */
   def remove(elem: A): Set[A] = set - elem
 
+  /**
+   * Returns the tail of this `NonEmptySet` if it exists or `None` otherwise.
+   */
+  def tailOption: Option[NonEmptySet[A]] = destructEither.toOption.map(_._2)
```
adamgfraser, line 101: Not sure these make sense for a set that is not ordered. This seems to imply that there is a "head" and "tail" which is not the case.

adamgfraser (changes requested):
Definitely good to have helper methods for converting to and from these data types but I think some of these operators imply an ordering and a pattern of linear access that is not an efficient way to work with these data structures and is not provided by the Scala collections library for analogous data structures.

core/shared/src/main/scala/zio/prelude/NonEmptyList.scala
sideeffffect: Nice! Thank you.

core/shared/src/main/scala/zio/prelude/NonEmptyList.scala
sideeffffect: Yep, but it is sometimes handy to have it like a method.

core/shared/src/main/scala/zio/prelude/NonEmptySet.scala
sideeffffect: True, there is no order. But sometimes you need to be able to access _some_ element from the collection, even though there isn't any order and the element is picked "arbitrarily". And sometimes you may need to have access to the rest of the collection without that element. And sometimes you may need both.

adamgfraser (commented):
I think these are definitely good operators to have, I'm just trying to think of names for them that are idiomatic and don't imply an ordering when an ordering doesn't exist.

I don't think I've seen `destruct` anywhere else and I think the `head` and `tail` names are problematic because they imply an ordering.

Here are some ideas:

**NonEmptyList**
- `destruct` - `peel` analogous to `ZStream#peel`?
- `destructEither` - `peelEither`?

**NonEmptySet**
- `destruct` - `peel`?
- `destructEither` - `peelEither`?
- `tailOption` - delete?

**ZNonEmptySet**
- `destruct` - `peel`?
- `destructEither` - `peelEither`?
- `head` - `extract`?
-`tail` - delete?
- `tailOption` - delete?

**ZSet**
Do we need operators here? Understand the value in the other cases because they have this property of being non-empty that normal Scala collections don't have but this is more analogous to a traditional collection that could be empty.

sideeffffect:
:+1: on `peel` and `peelEither`

Regarding `head` and `tail`, I agree that the names are a bit misleading. But these names are already established in Scala, virtually all collections (including Set and Map) have them via IterableOps. So my thinking was that we use them and at least give them "type safe" signatures (e.g. `tail` throws on empty Set, etc).

core/shared/src/main/scala/zio/prelude/ZSet.scala
```diff
@@ -86,6 +86,14 @@ final class ZSet[+A, +B] private (private val map: HashMap[A @uncheckedVariance,
       }
     })
 
+  def destruct[B1 >: B: Equal](implicit Inv: Inverse[Sum[B1]], Idn: Identity[Prod[B1]]): Option[(A, ZSet[A, B1])] =
```
adamgfraser, line 89: I don't think these operators are well defined. If the measure is an integer for example that could be negative I don't think returning one of the values and then a new set with the measure even more negative corresponds to the traditional definition of processing every element of a collection that you are doing for here.

core/shared/src/main/scala/zio/prelude/NonEmptyList.scala
```diff
@@ -78,6 +78,21 @@ sealed trait NonEmptyList[+A] { self =>
...
+  final def peel: (A, List[A]) = self match {
+    case Single(head)     => (head, List())
+    case Cons(head, tail) => (head, tail.toList)
+  }
+
+  /**
+   * Decomposes the `NonEmptyList` either into an element and the remaining `NonEmptyList`,
+   * or just a single element, if there aren't any other.
+   */
+  final def peelEither: Either[A, (A, NonEmptyList[A])] = self match {
```
adamgfraser, line 91: Should this be an `Either` or an `(A, Option[NonEmptyList[A])`. They are isomorphic but often it makes sense to factor out the common term of the sum type so it is really reflecting what is different. You always have an `A` and then you may or may not have another `NonEmptyList[A]`.

adamgfraser:
Fair point on the Scala collections though I'm not sure that them defining `head` for something like a `Set` or a `Map` is a good thing. 😃 

For `NonEmptySet` it is already convertible to a Scala Set so having variants of existing methods makes sense.

For `ZSet` I don't think these operators are well-defined, at least with the existing constraints. The measure could be negative, fractional, or a range for example so trying to extract an element like this doesn't necessarily make sense. Some of these could potentially be well-defined for the case where the measure is `Natural` or `Boolean` but I really think we should be pushing people to use collection operators rather than processing elements individually for a collection that not only isn't ordered but doesn't necessarily have a concept of integral elements at all.

core/shared/src/main/scala/zio/prelude/NonEmptyList.scala
sideeffffect: Good idea :+1:

core/shared/src/main/scala/zio/prelude/ZSet.scala
sideeffffect: Makes sense, I've implemented it only for `*MultiSet`.

core/shared/src/main/scala/zio/prelude/ZSet.scala
```diff
@@ -396,3 +402,32 @@ trait LowPriorityZSetImplicits {
     )
 
 }
+
+trait ZSetSyntax {
+  implicit final class ZSetMapOps[+A](self: Map[A, Natural]) {
+
+    /** Converts a `Map[A, Int]` to a `MultiSet` */
+    def toMultiSet[A1 >: A]: MultiSet[A1] = MultiSet.fromMap(self)
```
sideeffffect, line 410: Btw @adamgfraser are these `[A1 >: A]` necessary for for methods that don't take have any parameters and only return values?

adamgfraser: You don't need it if `A` just appears as an output like here.

core/shared/src/main/scala/zio/prelude/ZNonEmptySet.scala
```diff
@@ -87,6 +87,9 @@ final class ZNonEmptySet[+A, +B] private (private val zset: ZSet[A, B]) { self =
   )(implicit ev1: Commutative[Sum[B1]], ev2: Commutative[Prod[B1]]): ZNonEmptySet[C, B1] =
     new ZNonEmptySet[C, B1](zset.flatMap(f(_).toZSet))
 
+  /** Returns an element */
+  def head: A = zset.toMap.head._1
```
adamgfraser, line 91: This is potentially a little weird in that the `head` could be an element with a measure that is negative or not integral. Somewhat beyond the scope of this PR but I think the concept of what it means to be `NonEmpty` is a little unclear. At the least that it doesn't have any elements with measures that are equal to an identity value? Not sure that is enforced right now.

adamgfraser (dismissed):

adamgfraser (approved):

PR #578, opened 2021-03-16
ZValidation#toZIO via Cause

sideeffffect:
I've been thinking that even though that `ZValidation` has `NonEmptyChunk` internally, it might be "cleaner" to convert it to `IO[E, A]` (and not `IO[NonEmptyChunk[E], A]`).
What do you think?

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -195,7 +195,7 @@ sealed trait ZValidation[+W, +E, +A] { self =>
    */
   final def toZIO: IO[E, A] =
     self.fold(
-      nec => ZIO.halt(nec.foldLeft[zio.Cause[E]](zio.Cause.empty)((c, e) => zio.Cause.Both(c, zio.Cause.fail(e)))),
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -193,8 +193,11 @@ sealed trait ZValidation[+W, +E, +A] { self =>
   /**
    * Converts this `ZValidation` into a `ZIO` effect, discarding the log.
    */
-  final def toZIO: IO[NonEmptyChunk[E], A] =
-    ZIO.fromEither(self.toEither)
+  final def toZIO: IO[E, A] =
+    self.fold(
+      nec => ZIO.halt(nec.foldLeft[zio.Cause[E]](zio.Cause.empty)((c, e) => zio.Cause.Both(c, zio.Cause.fail(e)))),
```
adamgfraser, line 198: Can we use `reduceMapLeft` here so we don't have to handle the empty case?

adamgfraser (changes requested):
Great idea!

core/shared/src/main/scala/zio/prelude/ZValidation.scala
sideeffffect: Nice!

adamgfraser (approved):
Looks great!

PR #579, opened 2021-03-17
Lazy error for fromPredicateWith methods

poslegm:
Example:

```scala
val x = 42
ZValidation.fromPredicateWith(s"error in cond for value $x")(x)(_ > 0)
```

String interpolation will be avoided at runtime.


adamgfraser (approved):
Thank you!

PR #580, opened 2021-03-17
Use Domain Specific Names For Validation Operators

adamgfraser:
`mapParN` goes to `validateWith` and `tupledPar` goes to `validate`.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -402,10 +402,488 @@ object ZValidation extends LowPriorityValidationImplicits {
     Success(Chunk(w), ())
 
   /**
+   * Constructs a `Validation` that succeeds with the specified value.
+   */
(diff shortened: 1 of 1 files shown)


sideeffffect (approved):
I'm wandering whether applying a function named `validate` on multiple values of type `Validation` could feel a bit weird ("haven' they been validated already?"), but it's better than `mapParN` I guess :shrug:

adamgfraser:
I'm totally open to other names. I think the logic would be you are going through a bunch of values that represent either validation successes and failures and checking them to determine whether the whole thing is valid. But maybe there is something better.

sideeffffect:
It's late, so the only word that comes to mind is `confirm` -- as in "_confirm_ that these are successful"...
But I have to admit it's still not great. :shrug: 
Out of ideas :smile: 

I think the PR is fine to merge -- if we come up with a better name, we can still rename.

adamgfraser:
👍 I agree it is a big improvement from the current names.

PR #582, opened 2021-03-19
Add ZValidation#toEitherUnordered (returning NonEmptyMultiSet)

sideeffffect:
Since we want `ZValidation` to behave as if the errors were accumulated in `NonEmptyMultiSet`, I've been thinking that we should be nudging our users to use methods that work with `NonEmptyMultiSet`.
Methods that work with the actual `NonEmptyChunk` should not be in the spotlight as much and people should use them only if explicitly wanted (`toEitherOrdered` or `toEitherWith`).

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -88,7 +88,7 @@ sealed trait ZValidation[+W, +E, +A] { self =>
     }
 
   /**
-   * Folds over the error and success values of this `ZValidation`.
+   * Folds over the ordered errors and success values of this `ZValidation`.
    */
   final def foldOrdered[B](failure: NonEmptyChunk[E] => B, success: A => B): B =
     self match {
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -167,13 +167,13 @@ sealed trait ZValidation[+W, +E, +A] { self =>
     }
 
   /**
-   * Transforms this `ZValidation` to an `Either` with errors in the order they occurred, discarding the log.
+   * Transforms this `ZValidation` to an `Either`, discarding the log.
    */
   final def toEither[E1 >: E]: Either[NonEmptyChunk[E1], A] =
     fold(Left(_), Right(_))
 
   /**
-   * Transforms this `ZValidation` to an `Either`, discarding the log.
+   * Transforms this `ZValidation` to an `Either`, discarding the order in which the errors ocurred and discarding the log.
    */
   final def toEitherUnordered[E1 >: E]: Either[NonEmptyMultiSet[E], A] =
     self match {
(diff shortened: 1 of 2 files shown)


adamgfraser (commented):
Hmmm, I thought we were going in a different direction on this. We previously got a complaint about the fact that the order was lost in the multiset.

From a correctness perspective the multiset is definitely useful to compare two collections for equality without regard for order but taking into account how many times elements appear. However, I'm not sure it is the best user facing data type. `Chunk` is our flagship data type from ZIO that is baked in to almost all ZIO and ZIO Stream operators that return a collection, preserves information about order of errors if users care, and has a very familiar API.

Multiset is definitely a "fancier" data type and most of the time you don't have multiple identical errors and just want to get all the errors.

So I would tend to think the most user friendly API is to return a `NonEmptyChunk`. If users don't care about the order they don't have to use that information. This also avoids the need to have additional operators which fits with our goal of simplifying and avoiding multiple ways to do the same thing.

sideeffffect:
You're right. `NonEmptyMultiSet` may not be the most user friendly and well known collection.

I have reworked the PR to only add `toEitherUnordered`. That method is still immensely useful for defining much of ZValidation's behavior, including `hashCode`, which has been broken till now (or inconsistent with the Hash instance at least).

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -170,6 +172,15 @@ sealed trait ZValidation[+W, +E, +A] { self =>
   final def toEither[E1 >: E]: Either[NonEmptyChunk[E1], A] =
     fold(Left(_), Right(_))
 
+  /**
+   * Transforms this `ZValidation` to an `Either`, discarding the order in which the errors occurred and discarding the log.
+   */
+  final def toEitherUnordered[E1 >: E]: Either[NonEmptyMultiSet[E], A] =
```
adamgfraser, line 178: `ToEitherMultiSet`? We technically aren't guaranteeing any ordering for the errors with the `Chunk` variant, we just actually observe them in some order and happen to return them in the same order.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -46,12 +46,14 @@ sealed trait ZValidation[+W, +E, +A] { self =>
...
-      case (Failure(_, e), Failure(_, e1)) => e == e1
-      case (Success(_, a), Success(_, a1)) => a == a1
-      case _                               => false
+    that match {
+      case that: AnyRef if self.eq(that) => true
+      case that: ZValidation[_, _, _]    => self.toEitherUnordered == that.toEitherUnordered
+      case _                             => false
     }
 
+  override final def hashCode(): Int = toEitherUnordered.hashCode()
```
adamgfraser, line 55: Good fix!

adamgfraser (dismissed):
Looks good! One suggestion on naming but fine either way!

core/shared/src/main/scala/zio/prelude/ZValidation.scala
sideeffffect: Renamed :+1:

adamgfraser (approved):

PR #584, opened 2021-03-21
provide ZValidation.flatten

phderome:


adamgfraser (approved):
Looks great! Thanks for adding!

sideeffffect (dismissed):
Awesome, thank you!

adamgfraser:
@phderome Can you format?

sideeffffect:
@phderome you may want to run `fix` before submitting a PR to make sure `lint` won't fail :wink:

phderome:
> @phderome you may want to run `fix` before submitting a PR to make sure `lint` won't fail 😉

Will do. Consider adding that to https://zio.github.io/zio-prelude/docs/about/about_contributing as I didn't see it there being mentioned.

adamgfraser:
@phderome I think we should align the commands with the ones in ZIO. The contributor's guide actually does mention it here but it is the wrong command.

image

sideeffffect:
`fix` works on both ZIO and ZIO Prelude. It runs _both_ Scalafix and Scalafmt to ensure all what can be automatically fixed gets fixed :wink:

adamgfraser:
Sorry, I'm thinking of `prepare`. I think we have that on ZIO but not on ZIO Prelude.

adamgfraser:
https://github.com/zio/zio/blob/9a199aff8bd0b649f1573a4d2097f062795d8ff9/build.sbt#L27

adamgfraser (approved):

PR #585, opened 2021-03-21
Update Contributing.md (replace fmt with fix)

phderome:
as headline says

See https://github.com/zio/zio-prelude/pull/584#issuecomment-803642226

### DIFF
--- a/docs/about/contributing.md
+++ b/docs/about/contributing.md
@@ -150,9 +150,9 @@ sbt test
 If all the tests are passing, then you can format your code:
 
 ```bash
-# If you are already in a SBT session you can type only 'fmt'
+# If you are already in a SBT session you can type only 'fix'
 
-sbt fmt
+sbt fix
 ```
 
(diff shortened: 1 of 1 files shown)


adamgfraser:
Love aligning this but would actually go the other way and just rename `fix` to `fmt` so we have the same command for users who are contributing to both ZIO and ZIO Prelude.

phderome:
> Love aligning this but would actually go the other way and just rename `fix` to `fmt` so we have the same command for users who are contributing to both ZIO and ZIO Prelude.

Now I "get" it.

sideeffffect (approved):
`fix` is actually the correct command. It works on both ZIO and ZIO Prelude. It runs both Scalafix and Scalafmt to ensure all what can be automatically fixed gets fixed :wink:

phderome:
closing until there is absolute clarity. Perhaps it's the ZIO project that needs updating.

PR #591, opened 2021-04-01
Remove Error by putting into Warning Log

nbenns:
Add "or" to Validation which if the first fails will append the warnings and errors from the first into the second, returning the second result.

The usage is when you just want to use the error type to verify you have handled all failure scenarios (want Nothing for the error type) and use the warnings to log the rest.

eg.

```
val a: ZValidation["one", "bad", Nothing] = ZValidation.fail("bad": "bad").log("one": "one")
val b: ZValidation["two", Nothing, Int] = ZValidation.succeed(1).log("two": "two")

val c: ZValidation[String, Nothing, Int] = a or b
```

which gives the same as
```
val res = ZValidation.succeed(1).log("bad").log("one").log("two")
```


core/shared/src/main/scala/zio/prelude/NotSubType.scala
```diff
@@ -0,0 +1,11 @@
+package zio.prelude
+
+object NotSubType {
+  // Encoding for "A is not a subtype of B"
+  trait <:!<[A, B]
+
+  // Uses ambiguity to rule out the cases we're trying to exclude
+  implicit def nsub[A, B]: A <:!< B            = null
+  implicit def nsubAmbig1[A, B >: A]: A <:!< B = null
+  implicit def nsubAmbig2[A, B >: A]: A <:!< B = null
```
sideeffffect, line 10: In ZIO, we have a convention of implementing this "trick".
https://github.com/zio/zio/blob/master/core/shared/src/main/scala-2.12-2.13/zio/CanFail.scala
Do you think you could do it more in line with that? :pray:

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -163,6 +163,20 @@ sealed trait ZValidation[+W, +E, +A] { self =>
...
     }
 
+  /**
+   * Returns a ZValidation that if succeeds is the result of the first or if fails is
+   * the result of the second.  The second will combine the warnings and errors of the first
+   * into the warnings of the second, then append any new warnings from the second.
+   */
+  final def or[W1 >: W, E1, E2 >: E, W2, B >: A](
+    that: ZValidation[W1, E1, B]
+  )(implicit lub: Lub[E2, W1, W2]): ZValidation[W2, E1, B] =
```
sideeffffect, line 173:
```scala
  final def or[W1 >: W, E1 >: E, A1 >: A](
    that: ZValidation[W1, E1, A1]
  )(implicit ev: E <:< W1): ZValidation[W1, E1, A1] =
```
I'm wondering whether we could go with just this :thinking: 

Not as powerful and flexible as `Lub`, but we wouldn't need `Lub`...

sideeffffect (changes requested):
Thanks for the contribution! :+1: 
I have a few suggestions.
Could you also add a test for this? :pray:

core/shared/src/main/scala/zio/prelude/ZValidation.scala
adamgfraser: Definitely agree about trying to avoid adding additional abstractions.

I'm also not sure that this is the most natural way to deal with an error in the alternative validation. It seems like there are several possible semantics:

1. Just fail with the new error. We "recovered" from the original error so it is no longer a cause of the failure.
2. Fail with both the original error and the new error, combining them with an `Associative` instance defined for `E`. This preserves the most information but arguably goes against the intent of us "recovering" from this error.
3. Push the original error into the log type. This is to some extent an intermediate between the first two but requires unifying the log and error types which is fairly constraining. It also doesn't really work at all if you are using `Validation` rather than `ZValidation` since it doesn't return a `Validation` anymore.

I would tend to think that (1) is the most intuitive, though I think you could debate between (1) and (2). You could call that `orElse`. Then I think (3) becomes something like `orElseLog` which not only recovers from the error but also adds the error to the log and has this additional constraint.

sideeffffect: How about we have
```scala
  final def orWith[W1 >: W, E1 >: E, A1 >: A](convert: E => W1)(that: ZValidation[W1, E1, A1]): ZValidation[W1, E1, A1] = ???
```
or
```scala
  final def orWith[W1 >: W, E1 >: E, A1 >: A](convert: Chunk[E] => Chunk[W1])(that: ZValidation[W1, E1, A1]): ZValidation[W1, E1, A1] = ???
```
?

nbenns: The thought behind keeping the original error around would be to know for example if a default was applied.
For example if an external service was unavailable during the time period of execution.  The data might still be valid, however perhaps a backup method is not as accurate and we may wish to schedule some further data process at a later time when it comes back online.

I suppose there is probably other ways to do this, such as always add warning log to the default that gets applied.

Something like
```
verifyItem or (default.log("default applied"))
```

Let me tinker around with the `orWith` version and see if that fits the use-cases I'm trying to solve.  I like that idea of the conversion as its even more general than expecting `E` and `W` to inherit from the same root object.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -163,6 +163,15 @@ sealed trait ZValidation[+W, +E, +A] { self =>
       case Success(w, a) => Success(f(w), a)
     }
 
+  final def or[E1, W1 >: W, A1 >: A](that: ZValidation[W1, E1, A1])(implicit ev: E <:< W1): ZValidation[W1, E1, A1] =
```
adamgfraser, line 166: This is still embedding the idea of moving the errors from the first effect into the log channel which is perfectly reasonable but also not necessarily what you want to do, especially since it requires going from a `Validation` to a `ZValidation` which you may not be using. I would suggest renaming `or` to `orElseLog` and `orWith` to `orElseLogWith`.

```scala
final def orElse[W1

File: core/shared/src/main/scala/zio/prelude/ZValidation.scala [hunk 4]
```diff
@@ -163,6 +163,15 @@ sealed trait ZValidation[+W, +E, +A] { self =>
...
+    orWith[W1, E1, A1](_.map(ev))(that)
+
+  final def orWith[W1 >: W, E1, A1 >: A](convert: Chunk[E] => Chunk[W1])(that: ZValidation[W1, E1, A1]): ZValidation[W1, E1, A1] =
+    self match {
+      case Failure(log, errors) => that.mapLogAll(log ++ convert(errors) ++ _)
```
adamgfraser, line 171: The semantics here are not very well defined. We now have three fairly different things that we are putting into the same collection here: (1) the log from the original value that failed, (2) the errors from the original value that failed, and (3) the log from the new value that failed.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
nbenns: 👍 I've also been thinking about what other combinators would be reasonable.  I do like the idea of not specifically requiring the log gets moved over.

nbenns:
@adamgfraser I think these are probably the simplest combinators that perform the action I was originally intending.
I start with `orElseLog` as `orElse(that).mapLogAll(_ ++ errors)` however I felt that it's more logical that `errors` gets ordered before `that.log` so that everything from `self` comes first and then everything from `that`.

Let me know if you want the commits squashed or any other changes you'd want.

adamgfraser (approved):
Thank you! 🙏

PR #599, opened 2021-04-08
Renovate: rebase PR only on conflict

sideeffffect:


sideeffffect:
@mijicd I think it would be good to have it here too :pray:

mijicd (approved):

PR #603, opened 2021-04-12
Add ZValidation#toException

sideeffffect:


core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -261,6 +271,21 @@ object ZValidation extends LowPriorityValidationImplicits {
...
+
+    def toExceptionWithCause(implicit ev: E <:< Throwable): Failure.Exception[W, E] = {
+      val exn = toException
+      errors.tail.foreach(exn.addSuppressed(_))
+      val _   = exn.initCause(errors.head)
+      exn
+    }
+  }
+  final object Failure {
+    final case class Exception[+W, +E](failure: Failure[W, E]) extends IllegalArgumentException(failure.message)
```
adamgfraser, line 288: Why not just wrap a `NonEmptyChunk[E]` here?

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -261,6 +271,21 @@ object ZValidation extends LowPriorityValidationImplicits {
...
     lazy val errorsUnordered: NonEmptyMultiSet[E] = NonEmptyMultiSet.fromIterable(errors.head, errors.tail)
+
+    lazy val message: String =
+      s"errors:\n  ${errors.map(_.toString).mkString("\n  ")}\nlog:\n  ${log.map(_.toString).mkString("\n  ")}"
+
+    def toException: Failure.Exception[W, E] = Failure.Exception(this)
+
+    def toExceptionWithCause(implicit ev: E <:< Throwable): Failure.Exception[W, E] = {
+      val exn = toException
+      errors.tail.foreach(exn.addSuppressed(_))
```
adamgfraser, line 282: I think if the exception already contains all the errors there is no need to add them as suppressed exceptions here.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
adamgfraser: Why `IllegalArgumentException`? Seems overly specific.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -218,6 +218,16 @@ sealed trait ZValidation[+W, +E, +A] { self =>
       ZIO.succeedNow
     )
 
+  final def unsafeGetOrThrow: A = this match {
```
adamgfraser, line 221: I don't think I would expose unsafe variants here. We could instead make this something like `toEitherThrowable` that would return an `Either[Throwable, A]` by just wrapping the failures in a new subtype of `Throwable`.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
adamgfraser: I would look at the implementation of `FiberFailure` in ZIO. We could have something similar like `ValidationFailure` if we wanted.

sideeffffect: > Why not just wrap a `NonEmptyChunk[E]` here?

I thought it would be useful to have also the `log` available

> Why `IllegalArgumentException`? Seems overly specific.

The most typical use case for Validation is _validate_ input arguments, so `IllegalArgumentException` seemed like a natural choice. What do you think would be a better choice?

> `FiberFailure`

OK, I'll check it out :+1:

adamgfraser: That's fine. It could just be:

```scala
final case class ValidationFailure(failure: Validation.Failure[Any]) extends Throwable(null, null, true, false)
```

core/shared/src/main/scala/zio/prelude/ZValidation.scala
sideeffffect: But "suppressed exceptions" are a standard mechanism on a Throwable that all the ecosystem and tools can with with, `errors` is just our custom field that nobody knows about. Same about "cause".

adamgfraser: We can just override `getMessage` to render all the failures. That way we can be polymorphic regarding the error type.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
sideeffffect: OK, I've dropped it and added `toEitherException`.
But I think this will be an common need for our users to assert that the Validation succeeded and throw in the exceptional case that it didn't.
Do you have any idea how to do it elegantly and without boilerplate on the side of our users?

core/shared/src/main/scala/zio/prelude/ZValidation.scala
sideeffffect: I have reworked it as `ZValidationFailureException`

core/shared/src/main/scala/zio/prelude/ZValidation.scala
sideeffffect: `ZValidationFailureException` is intended to as much regular JVM exception as possible. The goal is as good interoperability with the rest of the JVM ecosystem and tools as possible. This goes beyond just `message` but also includes `cause`, `suppressed`, etc.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
adamgfraser: I think the most typical thing for a user to do would be something like:

```scala
myValidation.fold(
  errors => throw new MyErrorType(s"validation of myDataType in myModule failed with errors: $errors.toString")
  value  => ???
}
```

We could potentially clean this up slightly with something like `getOrElse`:

```scala
trait ZValidation[+W, +E, +A] { self =>
  def getOrElse[A1 >: A](f: NonEmptyChunk[E] => A1): A1 =
    self.fold(f, identity)
}
```

Then you could do:

```scala
val a = myValidation.getOrElse { errors =>
 thrown new MyErrorType(s"validation of myDataType in myModule failed with errors: $errors.toString)
}
```

That is more general in that it allows the user to handle the error however they like and also potentially lets us avoid having to define this one particular subtype of `Throwable` that is not going to be what the user wants most of the time. `Either` for example doesn't define some built in way for users to convert arbitrary error types to `Throwable` but leaves that up to the user.

sideeffffect: I've added `getOrHandleException` to _handle_ this use case :smile_cat:

adamgfraser (commented):
Hmmm, I'm inclined to think we should just implement `getOrElse`.

There isn't a generic way to convert a collection of `E` values into a `Throwable`. We're seeing that just discussing the implementation of these methods. Should it be an `IllegalArgumentException` or not? Should it include the log? Should it use suppressed exceptions?

Since there isn't a generic way to do it we should just provide operators like `fold` or `getOrElse` that allow users to do it as easily as possible.

sideeffffect:
Sure there is :smiley: `ZValidationFailureException` is 100% the same thing as `ZValidation.Failure`, only a subtype of `Exception`, so it has `log` and whether it is `IllegalArgumentException` or not is immaterial. The whole point of it is to make it easy to `throw`. We can think of it as an interop with the wider JVM world.

I have extensive experience with Validated from cats, where I needed a functionality like this fairly often and had to do it myself, which was annoying, so I wanted to improve ZIO Prelude to make it more appealing compared to the competition. If we don't provide users with functions like `toException` or `getOrHandleException`, they will have to write more boilerplate code, I don't think I've been the only one with such use case.

The `toExceptionWithCause` variant is just a cherry on top. It builds on top of the basic variant and gives the user the benefit that the "JVM interop" is improved, but works only if `E`s are `Throwable`s. I think this extension of the basic variant is worthwhile and will be used in practice, not only because people make their domain errors subtypes of `Exception`.

Good point about the `getOrElse`, I've added it. (We already have `fold`.)

adamgfraser:
Hmmm. My experience has been that organizations that use `Validation` also create their own error types when they do need to throw exceptions, though I'm sure practices vary. This seems somewhat ad hoc to me and I don't see it as being so important that we need to do this as opposed to just letting people throw whatever errors they want to based on the information contained in `Validation`. So I am mildly 👎 on this.

sideeffffect:
Sometimes it's just easier to throw the exception you're given, without necessarily creating a custom exception yourself for such situation. Sometimes creating a custom exception class is just overkill. With this PR merged, ZIO Prelude would have a choice to do that. Typically, it would be useful in situations where an error is really unexpected and happens only in case of a severe failure.
Example where I would find this helpful is this: https://github.com/Idealiste-cz/ideal-voting-backend/blob/e27c04a/ideal-voting-server/src/main/scala/cz/idealiste/idealvoting/server/Db.scala#L79

But I've create a smaller PR, hopefully less controversial, with parts of functionality of this one:
https://github.com/zio/zio-prelude/pull/610

PR #610, opened 2021-04-26
Add ZValidation#getOrElse and ZValidation.Failure#message

sideeffffect:
Less controversial take on https://github.com/zio/zio-prelude/pull/603

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -105,6 +105,14 @@ sealed trait ZValidation[+W, +E, +A] { self =>
     }
 
   /**
+   * Returns the value, if successful, or the transformed (using `f`) failure.
+   */
+  final def getOrElse[A1 >: A](f: Failure[W, E] => A1): A1 = this match {
+    case Success(_, value)       => value
+    case failure @ Failure(_, _) => f(failure)
+  }
+
+  /**
    * Writes an entry to the log.
    */
   final def log[W1 >: W](w1: W1): ZValidation[W1, E, A] =
@@ -274,9 +282,15 @@ sealed trait ZValidation[+W, +E, +A] { self =>
 object ZValidation extends LowPriorityValidationImplicits {
 
   final case class Failure[+W, +E](log: Chunk[W], errors: NonEmptyChunk[E]) extends ZValidation[W, E, Nothing] {
+
+    /** The errors represented in a way which discards the order in which they occurred. */
     lazy val errorsUnordered: NonEmptyMultiSet[E] = NonEmptyMultiSet.fromIterable(errors.head, errors.tail)
+
+    /** The description of the failure. */
+    lazy val message: String =
+      s"errors:\n  ${errors.map(_.toString).mkString("\n  ")}\nlog:\n  ${log.map(_.toString).mkString("\n  ")}"
   }
-  final case class Success[+W, +A](log: Chunk[W], value: A)                 extends ZValidation[W, Nothing, A]
+  final case class Success[+W, +A](log: Chunk[W], value: A) extends ZValidation[W, Nothing, A]
 
   /**
    * The `Covariant` instance for `ZValidation`.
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -104,6 +104,14 @@ sealed trait ZValidation[+W, +E, +A] { self =>
       case Success(w, _) => w
     }
 
+  /**
+   * Returns the value, if successful, or the transformed (using `f`) failure.
+   */
+  final def getOrElse[A1 >: A](f: Failure[W, E] => A1): A1 = this match {
```
adamgfraser, line 110: Maybe one variant that just gives you the errors and another `getOrElseLog` variant that gives you a tuple of the log and the errors?

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -274,9 +282,15 @@ sealed trait ZValidation[+W, +E, +A] { self =>
 object ZValidation extends LowPriorityValidationImplicits {
 
   final case class Failure[+W, +E](log: Chunk[W], errors: NonEmptyChunk[E]) extends ZValidation[W, E, Nothing] {
+
+    /** The errors represented in a way which discards the order in which they occurred. */
     lazy val errorsUnordered: NonEmptyMultiSet[E] = NonEmptyMultiSet.fromIterable(errors.head, errors.tail)
+
+    /** The description of the failure. */
+    lazy val message: String =
```
adamgfraser, line 290: Isn't this covered by `Debug` instance?

adamgfraser (changes requested):
Looks good! Just a couple of minor comments.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
sideeffffect: I think the current `Debug` instance is all-right as it is, but `message` is formatted with newlines and so on to enhance readability, which I think is very useful for many purposes.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
sideeffffect: With `Failure`, you may get your hands on `message`, which is my point here :smile_cat: 

But I can add another method that will take just `f: NonEmptyChunk[E] => A1`. Do you think, it would be useful?

adamgfraser: It just seems a little irregular to expose one of the constructors in a fold variant as opposed to the underlying data values, at least where it isn't a fixed point type. Maybe a `prettyPrint` or `render` method on `ZValidation` could address this use case?

sideeffffect: What I'm trying to achieve here, is to make it super easy to do something like this:
```scala
.getOrElse(e => sys.error(e.message)) // this whole thing returns the successful value (or throws)
```
So I need to have the member `message` on `Failure`.

adamgfraser: Could we put it on `ZValidation` itself so it is like `validation.getOrElse(_ => sys.error(validation.render))`? Definitely want to make it easy to do something like that. Just having separate operators on the subtype of an algebraic data type and returning the subtype instead of its contents in a fold seems like a bit of a code smell.

sideeffffect: But then you have to use the variable (`validation` in your example) twice. In a situation, when you're juggling with multiple values of `Validation`, you have to make special care that you use the right one at the right place -- that's an unnecessary room for errors.

What I'm trying to replicate is an analogue of `valueOr` from cats:
https://typelevel.org/cats/api/cats/syntax/EitherOps.html#valueOr[BB%3E:B](f:A=%3EBB):BB

I've renamed it, so that it better reflects the meaning, `getOrElse` is already established for a slightly different concept.

PR #627, opened 2021-05-16
Make zio/zlayer IdentityBoth

regiskuckaertz:
I tried to use `ForEach` with ZIO but it failed b/c this instance is missing.


adamgfraser (approved):
Thank you!

PR #632, opened 2021-05-19
Return Same Collection Type In ZValidation.CollectAllPar

adamgfraser:
Honor the "returns the same collection type" guarantee in `ZValidation.collectAllPar` by using the `ForEach` abstraction.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -375,8 +375,8 @@ object ZValidation extends LowPriorityValidationImplicits {
    * that either returns the values of all of them, if they all succeed, or
    * else fails with all of their errors.
    */
-  def collectAllPar[W, E, A](validations: Iterable[ZValidation[W, E, A]]): ZValidation[W, E, List[A]] =
-    validations.foldRight[ZValidation[W, E, List[A]]](succeed(List.empty))(_.zipWithPar(_)(_ :: _))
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -375,8 +375,8 @@ object ZValidation extends LowPriorityValidationImplicits {
    * that either returns the values of all of them, if they all succeed, or
    * else fails with all of their errors.
    */
-  def collectAllPar[W, E, A](validations: Iterable[ZValidation[W, E, A]]): ZValidation[W, E, List[A]] =
-    validations.foldRight[ZValidation[W, E, List[A]]](succeed(List.empty))(_.zipWithPar(_)(_ :: _))
+  def collectAllPar[F[+_]: ForEach, W, E, A](validations: F[ZValidation[W, E, A]]): ZValidation[W, E, F[A]] =
```
phderome, line 378: You're doing this for better cohesion of the library in general (and if so are there other similar methods dependent on `Iterable` that have a `CoVariant` that could benefit from a similar change)? I brought up the fact in the past that some Prelude interfaces were extending `Iterable` and thus lacked some efficient implementations of methods implemented generically by Scala stdlib. Does `ForEach` definitely mitigate this issue?

phderome (approved):
Looks good.

core/shared/src/main/scala/zio/prelude/ZValidation.scala
adamgfraser: I checked the rest of the code base and didn't see this issue elsewhere so I think we are good here.

PR #636, opened 2021-05-23
Update to Scala 2.13.6

sideeffffect:
closes https://github.com/zio/zio-prelude/pull/630
closes https://github.com/zio/zio-prelude/pull/634

### DIFF
--- a/project/BuildHelper.scala
+++ b/project/BuildHelper.scala
@@ -294,7 +294,7 @@ object BuildHelper {
   )
 
   def nativeSettings = Seq(
-    Test / skip := true,
+    Test / test / skip := true,
     doc / skip := true,
     Compile / doc / sources := Seq.empty
   )
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
       fail-fast: false
       matrix:
         java: ['adopt@1.8', 'adopt@1.11']
-        scala: ['2.11.12', '2.12.13', '2.13.5', '3.0.0']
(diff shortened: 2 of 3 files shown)


adamgfraser (approved):

adamgfraser (approved):

project/BuildHelper.scala
```diff
@@ -294,7 +294,7 @@ object BuildHelper {
   )
 
   def nativeSettings = Seq(
-    Test / skip := true,
+    Test / test / skip := true,
```
bjaglin, line 297: Does this have any effect? I don't see where `skip` is looked up within `test` in sbt.

sideeffffect: Ha! Maybe it actually doesn't have any effect, but now at least `Test` isn't disabled entirely, so (sbt-)scalafix can work with it.
But how can we prevent `Test / test` from running the tests then? `Test / test := {}`?

bjaglin: > Ha! Maybe it actually doesn't have any effect, but now at least Test isn't disabled entirely, so (sbt-)scalafix can work with it.

:+1: it also makes sure that shared test sources build for all platforms, which wasn't the case despite the explicit reference in https://github.com/zio/zio-prelude/blob/b6ca2944eb5703561f27bef187b96cdfa9c8039c/build.sbt#L30-L33 

>  But how can we prevent `Test / test` from running the tests then? `Test / test := {}`?

I believe that would work, yes.

I assume CI skips it because of the usage of the command below, but a `test` at the root probably runs the native tests now (and wasn't before because the test framework wouldn't find anything compiled).

PR #648, opened 2021-06-07
[ZPure] Properly handle clearLogOnError when the whole computation fails

ghostdogpr:
The logic to clear the log was not applied when the whole computation failed (without any fold).
Fixed that and added tests for it.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
+++ b/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
@@ -743,7 +743,7 @@ sealed trait ZPure[+W, -S1, +S2, -R, +E, +A] { self =>
           zPure.flag match {
             case FlagType.ClearLogOnError =>
               clearLogOnError.push(zPure.value)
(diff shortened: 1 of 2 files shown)


sideeffffect (dismissed):

ghostdogpr:
Argh, found another bug, will fix and add another test

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -743,8 +743,12 @@ sealed trait ZPure[+W, -S1, +S2, -R, +E, +A] { self =>
           zPure.flag match {
             case FlagType.ClearLogOnError =>
               clearLogOnError.push(zPure.value)
-              curZPure = zPure.continue.fold(
-                e => { clearLogOnError.popOrElse(false); e },
+              curZPure = zPure.continue.bimap(
```
ghostdogpr, line 746: Using `fold` was wrong as it was recovering from the effect, used `bimap` instead.

sideeffffect (approved):

PR #653, opened 2021-06-19
rename groupBy to groupByNonEmpty on ForEach

felher:
Renames `groupBy` to `groupByNonEmpty` on `ForEach` as agreed upon on the discord server.

It names exactly the difference the method has to the std-lib method and `groupBy` is hard to reach because on most data types, it is already defined.

If you want to keep the old name around and just delegate from one to the other just let me know.


felher:
Oh, and FWIW, I renamed `groupByM` too for the sake of consistency, even though it does not have the collision problem.

adamgfraser (approved):

PR #655, opened 2021-06-25
ParSeq: Then distributes over Both even in the presence of Empty

sideeffffect:
Analogous to https://github.com/zio/zio/pull/5242 but for ParSeq.
But there's one thing: the property based testing of left/rightDistributivityLaw on ParSeq actually does _not_ uncover the bug.

For some reason the problematic case doesn't end up getting generated. I tried to have a look at `Gens.parSeq` play with it and tweak it, but to no success. But it looks like a carbon copy of `GenZIO.causes` which does generate the problematic case... So I'm baffled :smile: 

Would you @adamgfraser have any idea why is that the case?


adamgfraser (approved):
@sideeffffect Will take a look later!

PR #656, opened 2021-06-27
add `foldMapM` to `ForEach`

felher:
Today I found out that there is no `foldMapM` and since there is `foldLeft`, `foldLeftM` and `foldMap`, I think a `foldMapM` would be consistent and useful.

If there is anything I should change or you don't think `foldMapM` does warrant inclusion, just let me know.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ForEach.scala
+++ b/core/shared/src/main/scala/zio/prelude/ForEach.scala
@@ -107,6 +107,15 @@ trait ForEach[F[+_]] extends Covariant[F] { self =>
   }
 
   /**
+   * Effectfully maps each element of the collection to a type `B` for which an
+   * `Identity` is defined using the function `f`, then reduces those values to
+   * a single summary using the `combine` operation of `Identity`, or the
+   * `identity` element if the collection is empty.
+   */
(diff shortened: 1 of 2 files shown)


felher:
Huh. Has the timed out test something to do with the tests I added? They run very fast on my machine and it is not immediately clear to me how they could lead to a timeout.

sideeffffect:
Thanks for your contribution @felher ! I've restarted the build.

felher:
Thanks for restarting! Got the same result, though...
I looked at the log of the canceled tests the ones I've added seem to run fine and don't seem to take too much time:

tests

Maybe I've messed something else up?

adamgfraser (approved):

sideeffffect:
Stuck builds in GitHub Actions like this just happen. Or do we know the reason why, @adamgfraser ?

adamgfraser:
@sideeffffect Builds are generally much more stable in other projects. My current hypothesis is that these issues relate to our extensive use of lazy values leading to deadlocks during initialization.

sideeffffect:
Ha! I didn't think of that :thinking: But that one could be a nightmare to debug :crying_cat_face:

PR #664, opened 2021-07-16
Add Documentation

adamgfraser:


sideeffffect (dismissed):
Awesome! :heart:

docs/overview/index.md
```diff
@@ -3,19 +3,17 @@ id: overview_index
...
+- **New Types** - Zero overhead new types to allow you to increase the type safety of your code base with zero overhead and minimal boilerplate.
+- **ZPure** - A description of a computation that supports logging, context, state, and errors, providing all the functionality traditionally offered by monad transformers with dramatically better performance and ergonomics.
 
 ## Installation
 
-TODO: Installation details
+Include ZIO Prelude in your project by adding the following to your build.sbt file:
+
+```scala
+libraryDependencies += "dev.zio" %% "zio-prelude" % "1.0.0-RC5
```
phderome, line 18: is there a way to automate the version artifact so no one updates it when it becomes stale (1.0.0-RC5)?

docs/overview/index.md
```diff
@@ -3,19 +3,17 @@ id: overview_index
...
-- [SF Scala: Reimagining Functional Type Classes](https://youtu.be/OwmHgL9F_9Q) - a talk by John A. De Goes and Adam Fraser
-- [The Terror-Free Guide to Introducing Functional Scala at Work](https://www.slideshare.net/jv2301/the-terrorfree-guide-to-introducing-functional-scala-at-work) - slides by Jorge Vásquez,
-- [Exploring ZIO Prelude The game-changer for type classes in Scala](https://youtu.be/OzoMofqsPg8) - talk by Jorge Vásquez
-- [What's Ap with zio-prelude?](https://justinhj.github.io/2020/08/02/whats-ap-with-zio-prelude.html) - an article by justinhj
+- **Functional Abstractions** - Functional abstractions to describe different ways of combining data, making it easy for you to combine complex data types in a principled way.
+- **Functional Data Types** - Additional data types to supplement the ones in the Scala standard library such as `Validation` and `NonEmptyList` to enable more accurate domain modeling and handle common problems like data validation.
+- **New Types** - Zero overhead new types to allow you to increase the type safety of your code base with zero overhead and minimal boilerplate.
```
phderome, line 10: do you want to touch on `SmartType` at all here or not, using a word like `refined` possibly?

docs/overview/index.md
```diff
@@ -3,19 +3,17 @@ id: overview_index
 title: "Summary"
 ---
 
-ZIO Prelude is an alternative approach to functional abstractions in Scala, which throws out the classic functor hierarchy in favor of a modular algebraic approach that is smaller, easier to understand and teach, and more expressive.
+ZIO Prelude is a library focused on providing a core set of functional data types and abstractions that can help you solve a variety of day to day problems. The tools provided by ZIO Prelude fall into the following main categories:
```
phderome, line 6: this is very compelling: focus on what it is and nothing else, don't distract us about what it is not.

docs/functionaldatatypes/equivalence.md
```diff
@@ -0,0 +1,80 @@
...
+
+```scala mdoc
+def vectorListEquivalence[A]: Equivalence[Vector[A], List[A]] =
+  Equivalence(_.toList, _.toVector)
+```
+
+Given these two equivalence relationships we can then define an equivalence relationship between a `Vector[A]` and a `Chunk[A]`.
+
+```scala mdoc
+def vectorChunkEquivelence[A]: Equivalence[Vector[A], Chunk[A]] =
```
phderome, line 51: typo a/e: you wrote `vectorChunkEquivelence` should be `...valence`

docs/functionaldatatypes/equivalence.md
```diff
@@ -0,0 +1,80 @@
...
+      implicit val equivalence = listChunkEquivalence[Int]
+      val listGen = Gen.listOf(Gen.anyInt)
+      val chunkGen = Gen.chunkOf(Gen.anyInt)
+      checkAllLaws(Equivalence)(listGen, chunkGen)
+    }
+  }
+}
+```
+
+ZIO Test will generate a large number of `A` and `B` values and check that the identity law holds. This can be helpful to catch corner cases where we think two things are equivalent but they are really not.
```
phderome, line 78: +1 for integrating with ZIO Test in this section.

docs/functionaldatatypes/index.md
```diff
@@ -0,0 +1,13 @@
+---
+id: index
+title: "Summary"
+---
+
+ZIO Prelude includes several data types to help us model our domains more accurately and solve common problems.
+
+- **[Equivalence](equivalence.md)** - A description of an equivalence relationship between two data types.
+- **[NonEmptyList](nonemptylist.md)** - A list that is guaranteed to be non-empty to more accurately model situations where we know a collection has at least one element.
+- **[These](these.md)** - A data type that may either be a `Left` with an `A`, a `Right` with a `B`, or a `Both` with an `A` and a `B`, useful for modeling problems such as merging streams of data.
```
phderome, line 10: good statement for `These` with good example.

docs/functionaldatatypes/nonemptylist.md
```diff
@@ -0,0 +1,195 @@
...
+title: "NonEmptyList"
+---
+
+`NonEmptyList` represents a `List` that is guaranteed not to be empty.
+
+This is useful to allow more accurate domain modeling in situations where we know a collection is not empty.
+
+For example, we might be handling a collection of errors.
+
+If an error has occurred at all then we know the collection is not empty. However, the Scala compiler does not know that and so we are forced to handle the case where the collection is empty even though we know that cannot occur.
```
phderome, line 12: sugg: the Scala compiler does not know that in general for the standard library collections such as `Iterable`, `List`, `Vector`.
My point really is to be a bit more concrete about what Scala compiler does not know by clarifying what collections we have in mind. The actual wording I propose is not particularly important but illustrates the point.

phderome, line 12: pedantic: it's a little hand-wavy that Scala compiler does not know that without stating the standard library collections we have in mind (List, Vector, Iterable) though clearly just about everybody would think of those anyhow.

docs/functionaldatatypes/index.md
phderome: No `NonEmptySet` yet I guess.

docs/functionaldatatypes/nonemptylist.md
```diff
@@ -0,0 +1,195 @@
...
+```scala mdoc:nest
+object NonEmptyList {
+  def fromIterableOption[A](as: Iterable[A]): Option[NonEmptyList[A]] =
+    ???
+}
+```
+
+Notice that `fromIterableOption` returns an `Option[NonEmptyList[A]]` because the original collection might be empty and so we might not be able to construct a `NonEmptyList` from it.
+
+If we know the collection is not empty we can use the `fromIterable` operator and provide the first element of the collection to "prove" that it is not empty.
```
phderome, line 63: pedantic: I find double quotes around prove distracting, you may drop them. Alternatively, you could say "to satisfy the NonEmptyList type requirement".

docs/functionaldatatypes/nonemptylist.md
```diff
@@ -0,0 +1,195 @@
...
+We can also construct a `NonEmptyList` from the `::` case of a `List`.
+
+```scala mdoc:nest
+object NonEmptyList {
+  def fromCons[A](as: ::[A]): NonEmptyList[A] =
+    ???
+}
+```
+
+This is useful if we already have a `::` value, for example from a `ZIO` operator like `ZIO.validate` that returns `::[E]` if it fails.
```
phderome, line 81: +1: important context/justification.

docs/functionaldatatypes/nonemptylist.md
```diff
@@ -0,0 +1,195 @@
...
+For example, the `filter` operator is not implemented directly on `NonEmptyList`. But you can still filter a `NonEmptyList` because we can always view a `NonEmptyList` as a `List`.
+
+```scala mdoc
+val filtered: List[Int] =
+  nonEmptyList.filter(_ % 2 == 0) // okay
+```
+
+The return type of `List[Int]` is as specific as possible because the `filter` operator could filter out all the elements of the list and return a list that is empty.
+
+Thus, `NonEmptyList` lets us work in a way that is as close as possible to being as if `NonEmptyList` was a subtype of `List`, which conceptually it is, without us being able to actually rewrite the Scala standard library.
```
phderome, line 195: wording "without us being able to actually rewrite the Scala standard library." is unclear and seems irrelevant. No one expects us to rewrite the Scala standard library. This reads well by skipping that wording entirely.

phderome: The whole page is excellent and uses compelling context.

docs/functionaldatatypes/these.md
```diff
@@ -0,0 +1,57 @@
...
+  case class Right[+B](b: B) extends These[Nothing, B]
+  case class Both[+A, +B](a: A, b: B) extends These[A, B]
+}
+```
+
+You can think of `These` as like `Either` except that it has one additional case. Whereas an `Either` contains either an `A` or a `B` but never both, a `These` can contain both an `A` and a `B`.
+
+A simple example of a situation where `These` might arise is in merging two upstreams in a streaming application. We could be signaled when at least one upstream has data for us and could have a situation where either only the left upstream has data for us, only the right upstream has data for us, or both upstreams have data for us.
+
+Another example of using the `These` data type is in implementing the `zipAll` operator on collections.
```
phderome, line 20: you could clarify with wording use "hypothetical implementation" since ZIO `Chunk` actually uses Builder and not  `These`

docs/functionaldatatypes/validation.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+Thus, a `Validation` looks very much like an `Either` except that it can contain multiple errors instead of just one.
+
+The distinguishing feature of `Validation` is that it can accumulate multiple errors. This makes it extremely convenient for problems like data validation where we want to accumulate all validation errors and report them back instead of failing immediately on the first validation error.
+
+To see this, let's consider a simple example of validating a `Person` data type.
+
+```scala mdoc
+case class Person(name: String, age: Int)
```
phderome, line 28: wish list: reuse same example in `SmartType` section where types for `name` and `age` are `SmartType`. That would show combination of Prelude features together as you did in a Zymposium talk.

docs/functionaldatatypes/validation.md
```diff
@@ -0,0 +1,286 @@
...
+
+However, there is a problem with this. If we try to validate a `Person` with both an invalid `name` and an invalid `age` we will just get a single failure indicating that the name was empty.
+
+This is because `Either` embodies "fail fast" semantics where if one part of a computation fails the subsequent parts of the computation will never be performed.
+
+Normally this is what we want because there is no point in doing further work if a computation has already failed. This is similar to how in traditional imperative programs if an exception is thrown on one line then further lines will not be executed.
+
+However, it is not what we want here.
+
+We could imagine that the `name` and `age` were transmitted by an internal business customer. They would probably not be happy if we told them the name was invalid, they corrected that and sent it again, and then we failed again because the age was invalid.
```
phderome, line 59: pedantic: you may replace "internal customer" and "internal business customer" with "customer" and "business customer" as I don't see the benefit of classifying the customer as such.

docs/functionaldatatypes/validation.md
```diff
@@ -0,0 +1,286 @@
...
+Another useful constructor of `Validation` values is the `apply` method on `Validation`.
+
+```scala mdoc:nest
+object Validation {
+  def apply[A](a: => A): Validation[Throwable, A] =
+    ???
+}
+```
+
+The `apply` operator takes a by name argument and evaluates that argument, converting any non-fatal thrown exceptions into `Validation` failures. This is useful when working with code that might throw exceptions to safely import that code into a `Validation` value.
```
phderome, line 118: In practice, I've found this very useful recently.

docs/functionaldatatypes/validation.md
```diff
@@ -0,0 +1,286 @@
...
+    n <- validateInt(s)
+    _ <- validateAge(n)
+  } yield n
+```
+
+Here the second validation depended on the first one because if we could not parse the string into an integer then we would not even have an integer to validate. Just like with other data types we can express this with the `flatMap` operator and use a _for comprehension_ to chain multiple validations together.
+
+Note that when we chain validations like this we only do the second validation if the first one is successful so we will never see both errors here. If all we are doing is chaining then we don't actually need `Validation` and could just use `Either`.
+
+However, frequently we want to do some validations "in parallel" and other "sequentially" and having `flatMap` on `Validation` makes it very easy for us to do this. Note that when we say in parallel here no actual parallelism is going on here, we just mean that we are performing both validations even if one of them fails.
```
phderome, line 239: pedantic (API related, not doc related): that parallel wording feels unfortunate to me, but I hesitate to suggest it warrants a binary compatibility break in the future.

docs/functionaldatatypes/validation.md
```diff
@@ -0,0 +1,286 @@
...
+
+Now we are combining "parallel" and "sequential" validations.
+
+Validating the age will proceed by validating that the age can be parsed into an `Int` and then that the age is not negative. This will return either a validated age or a validation error if either of these steps failed.
+
+We are also validating that the name is not empty. The final result will be either a validated person or a validation error containing both any errors that occurred in validating the name and age.
+
+By combining operators like `validateWith` and `flatMap` you can build complex validation logic that handles any of your problems in this or related domains.
+
+## Extracting Validation Values
```
phderome, line 256: pedantic (API not doc): any thought of having a type check such as `isValid` predicate (and not doing something silly like `toOption.nonEmpty`)? It feels like it was left out intentionally somehow but `Either, Try` have them.

docs/functionaldatatypes/validation.md
```diff
@@ -0,0 +1,286 @@
...
+
+```scala mdoc:nest
+import zio.{IO, NonEmptyChunk}
+
+trait Validation[+E, +A] {
+  def toEither: Either[NonEmptyChunk[E], A]
+  def toEitherWith[E2](f: NonEmptyChunk[E] => E2): Either[E2, A]
+  def toZIO: IO[E, A]
+}
+```
```
phderome, line 280: fyi: you left out `toOption` and `toTry`

docs/functionaldatatypes/zset.md
```diff
@@ -0,0 +1,276 @@
...
+```scala mdoc:reset
+import zio.prelude.ZSet
+import zio.prelude.newtypes.Natural
+
+type MultiSet[+A] = ZSet[A, Natural]
+```
+
+This ends up being a quite useful data type and can represent anything where the number of times an element appears matters but the order doesn't matter, such as a shopping cart.
+
+By using other measures of how many times an item appears we can describe other variants of sets such as `Int` for sets where items can appear a negative number of times or `Double` for sets where the number of times an item appears can be fractional.
```
phderome, line 33: semi-pedantic: I believe you but this usage is rather novel to me, maybe an external reference to common usage of such idioms (or a quote)? I can imagine a debt of $5.03 being a negative double set of dollars but I am not sure I would want to model that as a "generalized" set.

docs/functionaldatatypes/zset.md
```diff
@@ -0,0 +1,276 @@
...
+```
+
+If we have an `Iterable` we can convert it into a `ZSet` using the `fromIterable` constructor. The measure of how many times each item appears will just be how many times it appears in the `Iterable`.
+
+```scala mdoc
+val multiSet: MultiSet[Int] =
+ ZSet.fromIterable(List(1, 1, 2))
+```
+
+Finally, we can just construct an empty `ZSet` using the `empty` operator and add elements to it later.
```
phderome, line 62: nitpick on "add elements to it later" as we're in a immutable data structure mindset, maybe some rewording could take place?
For example ... "and combine this empty `ZSet` with some non empty `ZSet` later"?

docs/functionaldatatypes/zset.md
```diff
@@ -0,0 +1,276 @@
...
+
+As a generalization of sets, `ZSet` supports many of the same operators that exist on sets.
+
+### Set Membership
+
+We can determine how many times an element appears in the set using the `apply` operator. This requires that the measure type has an `Identity` defined for it.
+
+If the element exists in the set the measure of how many times it appears will be returned. Otherwise the identity element will be returned.
+
+Thus, `shoppingList("apples)` will return `2` and `shoppingList("eggs")` will return `0` while `set(1)` will return `true` and `set(4)` will return `false`.
```
phderome, line 79: missing trailing double quote in `("apples)`

phderome:
```suggestion
Thus, `shoppingList("apples")` will return `2` and `shoppingList("eggs")` will return `0` while `set(1)` will return `true` and `set(4)` will return `false`.
```

docs/functionaldatatypes/zset.md
phderome:
```suggestion
Finally, we can just construct an empty `ZSet` using the `empty` operator and combine this empty ZSet with some non empty ZSet later.
```
(to avoid implying mutation)

docs/functionaldatatypes/zset.md
```diff
@@ -0,0 +1,276 @@
...
+  left.intersect(right)
+```
+
+The result will be `ZSet(2 -> 2, 3 -> 3)` because the only commonality of the two sets is that they contain these elements.
+
+## Set Difference
+
+We can also compute the difference of two sets.
+
+The difference of two sets is obtained by subtracting the measure of how many times each element appears in the second set from how many times it appears in the first set. This requires an `Inverse` defined on the `Sum` of the measure type to tell us what it means to "subtract" for this type.
```
phderome, line 123: I like this statement as it definitely encourages users to have "algebraic" reasoning/mindset when using Prelude abstractions.

docs/functionaldatatypes/zset.md
```diff
@@ -0,0 +1,276 @@
...
+  ZSet.fromIterable(List("apple", "banana"))
+
+val right: MultiSet[String] =
+  ZSet.fromIterable(List("banana", "banana", "banana"))
+
+val diff: MultiSet[String] =
+  left.diff(right)
+```
+
+Now the measure of how many times each item appears is a natural number so "banana" cannot appear a negative number of times.
```
phderome, line 153: I also like this.

docs/functionaldatatypes/zset.md
```diff
@@ -0,0 +1,276 @@
...
+```scala mdoc
+val sum: ZSet[Int, Double] =
+  die.zipWith(die)(_ + _)
+```
+
+The resulting set has eleven elements, corresponding to the possible sums from two dice rolls of two to twelve. The measure for each value corresponds to the probability of that sum being rolled.
+
+So for example the probability of seven is about 16.7% whereas the probability of rolling two or twelve is only about 2.8%.
+
+We can take this even further using the `flatMap`.
```
phderome, line 220:
```suggestion
We can take this even further using `flatMap`.
```

docs/functionaldatatypes/zset.md
```diff
@@ -0,0 +1,276 @@
...
+
+The `zipWith` operator lets us combine two sets when the sets are independent of each other. With the `flatMap` operator we can create a new set based on each element of the first set and then "flatten" the resulting set down.
+
+We could rewrite the example above using `flatMap` and a _for comprehension_ like this:
```
phderome, line 224: nitpick: I don't like the style of `for comprehension` here.

docs/functionaldatatypes/zset.md
```diff
@@ -0,0 +1,276 @@
...
+For example, we could add one to each value in a set like this:
+
+```scala mdoc
+val setPlusOne: ZSet[Int, Boolean] =
+  set.map(_ + 1)
+```
+
+This requires a `Commutative` instance defined for the `Sum` of the measure type so that we can combine the measures if we map two elements to the same value.
+
+## Transforming The Measure
```
phderome, line 249:
```suggestion
## Transforming the Measure
```

docs/functionaldatatypes/zset.md
```diff
@@ -0,0 +1,276 @@
...
+
+val fruitCount: Int =
+  fruits.foldLeft(0)((n, _) => n + 1)
+```
+
+This will return `5` because there are five elements in the set we are folding over.
+
+We can convert a `ZSet` back to a `Map` from elements to a measure of how many times they appear using the `toMap` operator.
+
+We can also always convert a `ZSet` back to a `Set` using the `toSet` operator. Any elements in the `ZSet` with a measure equal to zero based on an `Identity` defined on the `Sum` of the measure type will be discarded from the resulting set.
```
phderome, line 276: hopefully this works well for a Double measures and numbers almost equal to 0.0 but distinct from 0.0.

docs/functionaldatatypes/zvalidation.md
```diff
@@ -0,0 +1,94 @@
+---
+id: zvalidation
+title: "ZValidation"
+---
+
+Like many other data types in ZIO, the `Validation` data type we learned about before is actually a type alias for a more general data type called `ZValidation`.
```
phderome, line 6: all good.

docs/newtypes/index.md
```diff
@@ -0,0 +1,310 @@
...
+  val initial: SequenceNumber =
+    SequenceNumber(0)
+}
+type SequenceNumber = SequenceNumber.Type
+
+val initial: SequenceNumber =
+  SequenceNumber.initial
+```
+
+This let's us construct instances of the new type in a very ergonomic way, where constructors for the new type are on the new type companion object just like constructors for any other type would be.
```
phderome, line 114:
```suggestion
This lets us construct instances of the new type in a very ergonomic way, where constructors for the new type are on the new type companion object just like constructors for any other type would be.
```

docs/newtypes/index.md
```diff
@@ -0,0 +1,310 @@
...
+
+```scala mdoc
+import zio.Chunk
+
+val sequenceNumbers: Chunk[SequenceNumber] =
+  SequenceNumber.wrapAll(Chunk(1, 2, 3))
+
+val ints: Chunk[Int] =
+  SequenceNumber.unwrapAll(sequenceNumbers)
+```
```
phderome, line 137: you could reiterate the point of no runtime overhead for deconstructing as well though the reader will probably assume it anyhow.

docs/newtypes/index.md
```diff
@@ -0,0 +1,310 @@
...
+type SequenceNumber = SequenceNumber.Type
+
+val sequenceNumber: SequenceNumber =
+  SequenceNumber(1)
+
+val nextSequenceNumber: SequenceNumber =
+  sequenceNumber.next
+```
+
+Notice that the `next` operator was available on `SequenceNumber` without having to do any imports because we defined it as an extension method within the `SequenceNumber` object.
```
phderome, line 164: very nice example.

docs/newtypes/index.md
```diff
@@ -0,0 +1,310 @@
...
+object AccountNumber extends Newtype[Int] {
+  implicit val AccountNumberEqual: Equal[AccountNumber] =
+    Equal.default
+}
+type AccountNumber = AccountNumber.Type
+
+AccountNumber(1) === AccountNumber(1) // true
+AccountNumber(1) === AccountNumber(2) // false
+// AccountNumber(1) === 1             // does not compile
+```
```
phderome, line 209: not sure if I am correct: you could add a `AccountNumber(1) == 1 // true, compiles, but misleading` and discuss?

docs/newtypes/index.md
```diff
@@ -0,0 +1,310 @@
...
+When we use `Newtype` the new type we are creating is completely distinct from the underlying type. This can be useful when we want to "hide" the underlying representation of the new type except when we explicitly unwrap it.
+
+However, it can lead to boilerplate in some cases where we need to explicitly unwrap the new type or reimplement functionality that already exists on the underlying type.
+
+For example, say we want to test whether one `SequenceNumber` is after another. Right now this will not work:
+
+```scala mdoc
+// SequenceNumber(2) > SequenceNumber(1) // does not compile
+```
+
```
phderome, line 224: fix: this is not rendered somehow!

docs/newtypes/index.md
```diff
@@ -0,0 +1,310 @@
...
+object SequenceNumber extends Subtype[Int]
+type SequenceNumber = SequenceNumber.Type
+
+SequenceNumber(2) > SequenceNumber(1) // okay
+```
+
+Now `SequenceNumber` is a type that is different from `Int` but is still a subtype of `Int`.
+
+This means that we can use a `SequenceNumber` any time we need an `Int` and can use operators defined on `Int` on `SequenceNumber`. However, we still get the type safety os not being able to use an `Int` or `AccountNumber` when a `SequenceNumber` is expected.
```
phderome, line 240:
```suggestion
This means that we can use a `SequenceNumber` any time we need an `Int` and can use operators defined on `Int` on `SequenceNumber`. However, we still get the type safety of not being able to use an `Int` or `AccountNumber` when a `SequenceNumber` is expected.
```

docs/newtypes/index.md
```diff
@@ -0,0 +1,310 @@
...
+
+## Smart Types
+
+So far, all the new types we have created have been distinct from the underlying types but have not imposed any additional constraints on the values that the underlying type can take. For example, the `SequenceNumber` above could in principle be any `Int` value, whether that is `42` or `-1`.
+
+In many cases that makes sense. We often want to use a new type to denote that this particular `Int` value represents a sequence number and shouldn't be confused with another `Int` value that represents an account number without imposing other constraints.
+
+However, in other cases we may want to restrict the values that the underlying type can take. For instance, perhaps a `SequenceNumber` should never be negative.
+
+We can model this in ZIO Prelude by extending `NewtypeSmart` and `SubtypeSmart` when creating our new type and providing an `Assertion` that describes the constraints on the underlying value. We call these types smart types because they build in the logic for their own validation.
```
phderome, line 250: you could say 1-2 lines why you don't give examples for `NewtypeSmart` and/or why they're less interesting.

docs/newtypes/index.md
```diff
@@ -0,0 +1,310 @@
...
+val validatedSequenceNumber: Validation[String, SequenceNumber] =
+  SequenceNumber.make(1)
+
+val validateSequenceNumbers: Validation[String, Chunk[SequenceNumber]] =
+  SequenceNumber.makeAll(Chunk(1, 2, 3))
+```
+
+We can see that `make` and `makeAll` return a `Validation` that will either be a success with a valid instance of the smart type or a failure with a helpful error message indicating why a valid instance of the smart type could not be constructed.
+
+Because we have access to the `apply`, `wrap`, and `wrapAll` constructors within the smart type object we can implement operators that allow us to create instances of the smart type internally with zero overhead when we know that the underlying data is valid.
```
phderome, line 290: fyi: nice to know, I was not aware of this previously.

docs/newtypes/index.md
```diff
@@ -0,0 +1,310 @@
...
+
+For example, we might want to expose an `unsafeMake` operator for our users to allow them to construct instances of the smart type directly without going through `Validation`. As its name implies this places responsibility on the user to ensure that the underlying data is valid but could be necessary for performance or ergonomics in some situations.
```
phderome, line 292: I kind of like this as well. I was constructing ad hoc values in some tests and was annoyed to unwrap a `Validation` in the test with not much of a reason to do so.

phderome: You may give an example with custom assertions or bring up why they may be useful.

phderome (approved):
excellent work that will be much appreciated by users!

Have a look at some of my comments or nitpicks and see whether you need to make changes.

docs/zpure/index.md
```diff
@@ -0,0 +1,350 @@
...
+  parseInt(s).catchAll(_ => ZPure.succeed(0))
+```
+
+Now the error type is `Nothing`, indicating that we have handled all our errors.
+
+There are many other error handling operators defined on `ZPure`. The `catchSome` operator allows us to only recover from certain errors.
+
+The `fold` and `foldM` operators allow us to handle both the failure and success cases at the same time. And the `orElse` operator allows us to specify a fallback computation that will be run if the original computation fails.
+
+Since these operators are the same as the ones on `ZIO` we will not cover them all in detail here.
```
phderome, line 156: I'd be inclined to make this statement sooner by stating something along: we'll discuss some of the important operators of ZPure that have correspondence and good documentation in ZIO but will not give all details here such as method signatures as the reader can obtain them directly in Scala docs or assume familiarity with ZIO. The list of operators include: catchSome, mapError, fold, foldM, ...

My point is here that as I read along I am thinking why don't I see the operator type signatures just like I did elsewhere in docs for Prelude? And the explanation comes in later.

docs/zpure/index.md
```diff
@@ -0,0 +1,350 @@
...
+
+One other concept from `ZIO` that carries over to `ZPure` is the ability to accumulate multiple errors.
+
+Normally when we use operators like `zip` or `flatMap` if an error occurs we will not go on to evaluate further parts of the computation until we get to an error handler that can potentially recover from it. This is typically what we want because if we have already failed there is no point in doing more work.
+
+However, `ZPure` also allows us to accumulate errors when we use the `zipWithPar` operator . This operator does not do actual parallelism, but is "parallel" in the sense that it will run both computations even if the first one fails and return any failures that occurred.
+
+You can use this to obtain behavior similar to the `Validation` data type in ZIO Prelude. All of the errors will be captured in a `Cause` data structure similar to the one from `ZIO`.
```
phderome, line 164: I almost saw these comments coming! Nice!

docs/zpure/index.md
```diff
@@ -0,0 +1,350 @@
...
+  validatePerson(name, age).sandbox
+```
+
+We can now see all the failures that occurred and handle them using our normal error handling operators. If we want to submerge the full cause again and just see the error type we can undo this with the `unsandbox` operator.
+
+Once again, a variety of other operators for dealing with the full cause of failure are available on `ZPure` analogous to the ones on `ZIO` but we will not cover them all here. With `sandbox` and `unsandbox` you should be able to handle any problems involving working with the full cause of failure and you can always look up more specialized operators later.
+
+Of course, if all we want to do is validate data the `Validation` type is more specialized than this and is what we should use. But it is very nice to be able to accumulate errors when you need to when we are already working in the context of a `ZPure` computation.
+
+## Working With Context
```
phderome, line 198: 100% for that section. It's awesome.

docs/zpure/index.md
```diff
@@ -0,0 +1,350 @@
...
+  withdraw(10).provideState(AccountState(100, true))
+```
+
+We will use the `runEither` operator, which is a variant of `run` that allows the computation to fail and just converts the failure to the left side of an `Either`. The `run` and `runEither` operators just return the final value, so let's use the `get` operator to get the final state before we run our computation.
+
+```scala moc
+val updatedAccountState: Either[AccountError, AccountState] =
+  (withdrawalComputation *> ZPure.get).runEither
+```
+
```
phderome, line 319: you could explain in here that we're "cleaning up data" in that we're not interested at all in the `Unit` value of `withdrawalComputation` all we want is resulting state whereas for `interestComputation` we were interested in a final double value. Do you follow me?

docs/zpure/index.md
```diff
@@ -0,0 +1,350 @@
...
+## Working With Logging
+
+The final capability provided by `ZPure` is logging.
```
phderome, line 322: I guess if we feel that the `log` could consume too much memory than `ZPure` is not what we need and `ZIO` is more adequate?

docs/zpure/index.md
```diff
@@ -0,0 +1,350 @@
...
+One important thing to note here is that since `ZPure` computations are never concurrent we don't need to worry about conflicting modifications to the state.
+
+For example, in the context of `ZIO` code analogous to the above would not be safe because the `get` and `set` would not be performed atomically so we would need to do the entire transaction in a single `modify` operation, returning a value indicating whether there were sufficient funds. In `ZPure` we don't have to do any of that so our code can be simpler.
+
+There are also `update` and `modify` operators on `ZPure` so for example we could factor out the logic of updating the account balance like this:
+
+```scala mdoc
+def decrementBalance(amount: Int): ZPure[Nothing, AccountState, AccountState, Any, Nothing, Unit] =
+  ZPure.update(state => AccountState(state.balance - amount, state.open))
+```
```
tdrozdowski, line 304: I presume that you intend for `decrementBalance(amount: Int)` to replace the `else` clause in the `withdraw` example above?  If so, seems like it it would be nice to see that here prior to the run.

docs/functionalabstractions/index.md
```diff
@@ -3,4 +3,289 @@ id: index
...
 ---
 
-Coming soon.
+ZIO Prelude features a next generation approach to functional abstractions. This approach is based on the following ideas:
+
+1. **Algebraic** - Abstractions should describe fundamental algebraic properties.
+2. **Compositional** - These abstractions should describe properties that are orthogonal to each other, allowing definition of higher level abstractions as the composition of more basic ones.
+3. **Lawful** - Abstractions should be defined in terms of laws.
+
+The functional abstractions in ZIO Prelude can be broadly divided into two types.
```
phderome, line 12: "two kinds of types" might be better?

docs/functionalabstractions/index.md
```diff
@@ -3,4 +3,289 @@ id: index
...
+
+- **[Abstractions For Concrete Types](concretetypes/index.md)** - These abstractions define properties of concrete types, such as `Int` and `String`, as well as ways of combining those values.
+- **[Abstractions For Parameterized Types](parameterizedtypes/index.md)** - These abstractions define properties of parameterized types such as `List` and `ZIO` and ways of combining them.
+
+As we will see, there is a deep symmetry between the abstractions defined on concrete and parameterized types, such as concepts of associative operations, commutative operations, and identity . This reflects the fundamental nature of these algebraic properties and their ability to unify what were previously separate concepts.
```
phderome, line 17: remove space before dot after identity.

docs/functionalabstractions/index.md
```diff
@@ -3,4 +3,289 @@ id: index
...
+      map + (topic -> map.getOrElse(topic, Votes(0)).combine(votes))
+    })
+}
+```
+
+This isn't the worst but it isn't really the kind of code we want to be writing. We want to be thinking about the logic of our application rather than how to combine maps.
+
+This is where ZIO Prelude can help.
+
+The way we're combining these maps actually follows a pattern. If a value is in either map we include it in the combined map and if a value appears in both maps we combine the keys.
```
phderome, line 205: I understand it differently. See what you think of my wording, which I guess you'll reject?

If a key is in a single map we include it in the combined map with its associated value and if a key appears in both maps we combine the associated values of the maps for that key.

docs/functionalabstractions/concretetypes/index.md
```diff
@@ -3,4 +3,111 @@ id: index
 title: "Introduction"
 ---
 
-Coming soon.
+ZIO Prelude features a set of fundamental functional abstractions for working with concrete data types.
+
+These abstractions fall into two categories.
+
+## Properties Of Concrete Types
```
phderome, line 10: I prefer lower cases in titles for small words like of/the. I think it's conventional actually.

docs/functionalabstractions/concretetypes/index.md
```diff
@@ -3,4 +3,111 @@ id: index
...
+
+The `Hash` abstraction builds on the `Equal` abstraction in a different way, describing a type that can be hashed. Its fundamental representation is as a function `A => Int`.
+
+The `Hash` abstraction allows us to define our own way of hashing data types instead of being forced to use the implementation of `hashCode` for existing data types. It also ensures that the definition of the hash is consistent with the definition of equality, another common source of bugs.
+
+### Debug
+
+The `Debug` abstraction describes a type that can be rendered for debugging purposes. Its fundamental representation is as a function `A => Repr`, where `A` is a structured representation of the value that supports rendering in various formats.
+
+The `Debug` abstraction allows us to define how a data type should be rendered, for example rendering an `Array` to display its values rather than just its memory location. The more structured representation also allows rendering in different formats, for example including fully qualified names in the rendering of the data type so the rendering is itself valid Scala code.
```
phderome, line 42: +1

docs/functionalabstractions/concretetypes/index.md
```diff
@@ -3,4 +3,111 @@ id: index
...
+
+The most basic such abstraction is `Associative`, which describes a combining operation that is associative.
+
+```scala
+(a <> b) <> c === a <> (b <> c)
+```
+
+This essentially means that if we are combining three values then the order of operations doesn't matter. We can combine `a` and `b` and then combine the result with `c` or we can combine `b` and `c` and then combine `a` with the result.
+
+This is quite an important property for being able to build more complex structures from simpler ones. Otherwise the result of sequentially building more complex structures from simpler ones is not well defined because it depends on the order of operations and there are an exponentially increasing number of ways we could order these operations as we combine more and more values.
```
phderome, line 66: If you feel like it, you could mention zipping/tupling suffers from not being associative.

docs/functionalabstractions/concretetypes/index.md
```diff
@@ -3,4 +3,111 @@ id: index
...
+
+There are also an extremely large number of data types that can be combined using associative operations.
+
+For example integers can be combined associatively using addition and multiplication, among other operations. String concatenation is also associative, as is concatenation of lists of any concrete type.
+
+For these reasons the `Associative` abstraction forms the root of the set of abstractions for describing ways to combine concrete types in ZIO Prelude.
+
+### Commutative
```
phderome, line 74: I suppose that there's no interest/need of mentioning/warning that in classic math being commutative does not entail being associative as well?

docs/functionalabstractions/concretetypes/index.md
```diff
@@ -3,4 +3,111 @@ id: index
...
+
+This identity element is a "neutral value" that can be combined with any other value an unlimited number of times with the combining function and does not change the other value.
+
+For example `0` is an identity element with respect to addition, `1` is an identity element with respect to multiplication, the empty string is an identity element with respect to string concatenation, and the empty list is an identity element with respect to list concatenation.
+
+An identity element gives us a "starting point" for combining values that we know will never change the value. It also tells us when we can safely eliminate a value from the combining operation because we know the value will never change the result.
+
+### Inverse
+
+The `Inverse` abstraction further builds on the `Identity` abstraction with the concept of an inverse operation.
```
phderome, line 105: I think we need more explanation here along the lines of source code of the same that justifies the view of this being a binary operator rather than unary as the math analog is unary and so a first reading is confusing.

docs/functionalabstractions/concretetypes/index.md
```diff
@@ -3,4 +3,111 @@ id: index
 title: "Introduction"
 ---
 
-Coming soon.
+ZIO Prelude features a set of fundamental functional abstractions for working with concrete data types.
```
phderome, line 6: Looks good over all.

docs/functionalabstractions/concretetypes/equal.md
```diff
@@ -3,4 +3,135 @@ id: equal
...
+If we are using `==` we can't do anything about that because `==` is always based on how the data type implements `equals`. But if we are using `===` from ZIO Prelude we can define our own `Equal` instance that implements a more sensible notion of equality.
+
+## Defining Equal Instances
+
+Equal instances are automatically available for data types from the Scala library and ZIO with a meaningful notion of equality, as well as data types made up of those types.
+
+We can always check if an instance of a type class exists for a given data type by using the `apply` operator to tell the Scala compiler to find the appropriate instance for us. This will result in a compilation error if the instance cannot be found.
+
+```scala mdoc
+Equal[List[Either[String, Int]]] // okay
```
phderome, line 86: add a failing example for instance a case class wrapping a function? Since functions have no `Equal` a case class wrapping a function should not.

docs/functionalabstractions/concretetypes/equal.md
```diff
@@ -3,4 +3,135 @@ id: equal
...
+In the example above, we can convert any `Pair` into a tuple without losing any information. And ZIO Prelude already defines an `Equal` instance for any tuple of data types for which `Equal` instances are defined.
+
+So we can rewrite our implementation of `PairEqual` like this:
+
+```scala mdoc:nest
+object Pair {
+  implicit def PairEqual[A: Equal, B: Equal]: Equal[Pair[A, B]] =
+    Equal[(A, B)].contramap(pair => (pair.first, pair.second))
+}
+```
```
phderome, line 131: +1

docs/functionalabstractions/concretetypes/ord.md
```diff
@@ -3,4 +3,190 @@ id: ord
...
+object Ordering {
+  case object LessThan extends Ordering
+  case object GreaterThan extends Ordering
+  case object Equals extends Ordering
+}
+```
+
+That is, we can compare any two values of type `A` and either the `left` value is less than the `right`, the `left` value is greater than the `right`, or the two values are equal.
+
+Furthermore, the results of these comparisons must satisfy the properties that we expect of a total ordering. For example, if `a` is greater than `b` and `b` is greater than `c` then `a` must be greater than `c` and if `a` is greater than `b` then `b` must be less than `a`.
```
phderome, line 31: you are really speaking about **transitivity** of the relation here not total ordering. Although the ideas are related given the method `compare(left: A, right; A): Ordering` which is a total function rather than a partial function. But the way you have it written looks a little imprecise.

docs/functionalabstractions/concretetypes/ord.md
```diff
@@ -3,4 +3,190 @@ id: ord
...
+
+val compareConsumerAccounts: Boolean =
+  consumerAccount1 < consumerAccount2 // does not compile
+```
+
+This code doesn't compile! The Scala compiler is unable to compare the consumer accounts because of the lack of variance on `scala.math.Ordering`.
+
+We could widen the type of `ConsumerAccount` to `CustomerAccount` but that is unergonomic. And it wouldn't even work if, for example, we were trying to sort the list and needed to maintain the type information that these were consumer accounts rather than just customer accounts.
+
+The `Ord` abstraction helps us avoid problems like this and integrates seamlessly with the rest of ZIO Prelude.
```
phderome, line 148: great

docs/functionalabstractions/concretetypes/ord.md
```diff
@@ -3,4 +3,190 @@ id: ord
...
+
+## Defining Ord Instances
+
+`Ord` instances are already defined for all the types in `ZIO` and the Scala standard library that have a total ordering.
+
+ZIO Prelude is also very good at automatically deriving `Ord` instances for more complex data types like tuples. For example if we have a pair of `Double` values representing x and y coordinates an ordering is already automatically defined for it.
```
phderome, line 154: Is not it rather Scalac that is responsible for automatic derivation or there's Prelude doing something specific in terms of typeclass derivation here?

docs/functionalabstractions/concretetypes/ord.md
```diff
@@ -3,4 +3,190 @@ id: ord
...
+
+object Point {
+  implicit val PointOrd: Ord[Point] =
+    Ord.make { (left, right) =>
+      (left.y =?= right.y) <> (left.x =?= right.x)
+    }
+}
+```
+
+The `<>` operator combines two `Ordering` values, using the result of the first one if it is `GreaterThan` or `LessThan` and otherwise using the result of the second one. You can think of this like sorting by last name and then first name.
```
phderome, line 177: nice. It's rather boiler-plate-y not to have a mechanism like that.

docs/functionalabstractions/concretetypes/ord.md
phderome: General you may back-tick `x` and `y` throughout this file.

docs/functionalabstractions/concretetypes/hash.md
```diff
@@ -3,4 +3,118 @@ id: hash
 title: "Hash"
 ---
 
-Coming soon.
+`Hash[A]` describes the ability to hash a value of type `A`.
```
phderome, line 6: looks good!

docs/functionalabstractions/concretetypes/debug.md
```diff
@@ -3,4 +3,187 @@ id: debug
...
+
+The `Debug` abstraction is the functional equivalent of the `toString` operator but it has several advantages over just using `toString`.
+
+First, as with other functional abstractions we can define when it makes sense to render a data type at all.
+
+We can call `toString` on anything including classes whose string representation only points to their memory address. Fortunately this is more often the cause of annoyance than real bugs but it is certainly nice to know when there isn't much point in trying to render something.
+
+Related to this, we can define our own way of rendering for data types that are not under our control. For example, if we are debugging code involving arrays we can frequently be frustrated when the string representation of the array consists of its memory location rather than the elements actually in the array.
+
+With ZIO Prelude, we can easily define a `Debug` instance for `Array`. We can even handle nested arrays in a principled way.
```
phderome, line 38: I am not sure what "in a principled way" adds in terms of clarification. I would be inclined to skip it. To me the more important take-away is the sentence that follows (captures rendering information in a structured data format)

docs/functionalabstractions/concretetypes/debug.md
```diff
@@ -3,4 +3,187 @@ id: debug
...
+
+The final advantage that the `Debug` abstraction has over the Scala standard library and older functional programming libraries is that it captures rendering information in a structured data format instead of just a `String`.
```
phderome, line 40: +1

docs/functionalabstractions/concretetypes/debug.md
```diff
@@ -3,4 +3,187 @@ id: debug
...
+
+When we render a data type we may want to do it in various ways. We may want to do it using short names for a readable description in the context of a discussion like this one, rendering it as something like `Validation[String, Int]`.
+
+In another context like being able to copy and paste our rendering into a REPL, IDE, or worksheet and having it compile we may want to use fully qualified names, such as `zio.prelude.Validation[scala.String, scala.Int]`. Or we may decide that including the fully qualified name for standard library types is a little too much, and want to render this as `zio.prelude.Validation[String, Int]`.
+
+All of these are valid choices, but just rendering values as a `String` forces us to choose one. A `String` does not have enough structure to preserve all the information we need for rendering different ways so we are forced to choose one arbitrarily.
+
+This also allows us to define meaningful laws for the `Debug` abstraction. In other functional programming libraries interfaces that provide similar functionality are essentially lawless since there is nothing describing what this string representation should look like.
```
phderome, line 48: I'd not see anything particularly wrong in mentioning `Show` "typeclass" but leaving it out to be understood works as well.

docs/functionalabstractions/concretetypes/debug.md
```diff
@@ -3,4 +3,187 @@ id: debug
...
+
+In contrast, in ZIO Prelude `Debug` follows a well defined law that the Scala rendering of any data type should itself be valid Scala code.
+
+## Structured Rendering
+
```
phderome, line 53: you may back-tick `sealed trait`.

docs/functionalabstractions/concretetypes/debug.md
```diff
@@ -3,4 +3,187 @@ id: debug
...
+The `Repr` data type preserves all the information we need to render a data type to a `String` in various ways. There is nothing magic about this data type, it is just a sealed trait with a variety of cases representing the different possible pieces of information we could need for rendering.
+
+If you just want to render data types from ZIO or the Scala standard library that have meaningful string representations then you can just use the `debug` operator but if you want to define `Debug` instances for your own data type it is helpful to understand something about `Repr`.
```
phderome, line 56: proposal: it is helpful to understand **usage of** `Repr`.

docs/functionalabstractions/concretetypes/debug.md
```diff
@@ -3,4 +3,187 @@ id: debug
...
+
+```scala mdoc:nest
+object Repr {
+  case class Object(namespace: List[java.lang.String], name: java.lang.String) extends Repr
+}
+```
+
+As its name implies, this lets us define representations for data types like objects that aren't existing primitive types but also do not depend on other data types for their rendering. The information needed to render an object is just the name of the object and its namespace.
+
+For example, the `Repr` of `None` would be `Repr.Object(List("scala"), "None")`. Note how the namespace gives us the information we need to render the `Repr` as valid Scala code and gives us the ability to decide at the time of rendering whether we want to include some or all of this or just the name itself.
```
phderome, line 90: +1

docs/functionalabstractions/concretetypes/debug.md
```diff
@@ -3,4 +3,187 @@ id: debug
...
+
+  implicit def ValidationDebug[E: Debug, A: Debug]: Debug[Validation[E, A]] =
+    Debug.make {
+      case Success(a)  => Debug.Repr.VConstructor(List("zio", "prelude"), "Validation.Success", List(a.debug))
+      case Failure(es) => Debug.Repr.VConstructor(List("zio", "prelude"), "Validation.Failure", List(es.debug))
+    }
+}
+```
+
+Here we used context bounds of `Debug` for the `E` and `A` type parameters to require that `Debug` instances for them exist. This makes sense because there is no way we can create a meaningful string representation of a value if we can't create meaningful string representations of its components.
```
phderome, line 177: +1

docs/functionalabstractions/concretetypes/debug.md
```diff
@@ -3,4 +3,187 @@ id: debug
 title: "Debug"
 ---
 
-Coming soon.
+`Debug[A]` describes the ability to render a value of type `A` to a human readable format for debugging purposes.
```
phderome, line 6: This is good, it contains everything one needs to get started, and it even has some nice examples. Having worked with it, readers might miss out some of the nice capabilities by reading this doc too quick or not thinking hard enough. More substantial and meaningful examples from the community (or a talk) could be quite helpful to make this point across.

docs/functionalabstractions/concretetypes/associative.md
```diff
@@ -3,4 +3,257 @@ id: associative
...
+However, we will typically not need to do that for new types like `Sum` and `Prod` because they are subtypes of `Int`.
+
+```scala mdoc:nest
+val int: Int =
+  prodInt
+```
+
+Let's use these types to solve our problem from above.
+
+```scala mdoc
```
phderome, line 157: The fact that the examples that follow here compile without ambiguity makes your point all right. It might be even more obvious if you'd provide the unwrapped resulting `Int` value in the examples.

docs/functionalabstractions/concretetypes/associative.md
```diff
@@ -3,4 +3,257 @@ id: associative
...
+This maps each element in a collection to a new data type for which an `Associative` instance is defined and then reduces all of those values to a single summary value with the `combine` operator.
+
+We could instead count the number of occurrences of each word like this:
+
+```scala mdoc:nest
+def wordCount(lines: NonEmptyChunk[String]): Map[String, Int] =
+  lines.reduceMap { line =>
+    Sum.wrapAll(line.split(" ").groupBy(identity).view.mapValues(_.length).toMap)
+  }
+```
```
phderome, line 255: nice example.

phderome (dismissed):

docs/overview/index.md
adamgfraser: Done.

docs/overview/index.md
adamgfraser: Done.

docs/functionaldatatypes/equivalence.md
adamgfraser: Fixed.

docs/functionaldatatypes/index.md
adamgfraser: There actually is but it is a little more of a wrapper for a normal set than a full blown data type.

docs/functionaldatatypes/nonemptylist.md
adamgfraser: Changed to "to establish that it is not empty".

docs/functionaldatatypes/nonemptylist.md
adamgfraser: Changed to incorporate specific collections.

docs/functionaldatatypes/nonemptylist.md
adamgfraser: Good suggestion.

docs/functionaldatatypes/these.md
adamgfraser: Clarified to "implement a variant of".

docs/functionaldatatypes/validation.md
adamgfraser: Done.

docs/functionaldatatypes/validation.md
adamgfraser: Good feedback.

docs/functionaldatatypes/validation.md
adamgfraser: Changed to accumulating errors and chaining validations.

docs/functionaldatatypes/zset.md
adamgfraser: Thanks!

docs/functionaldatatypes/zset.md
adamgfraser: Thanks!

docs/functionaldatatypes/zset.md
adamgfraser: Changed.

docs/functionaldatatypes/zset.md
adamgfraser: Changed.

docs/functionaldatatypes/zset.md
adamgfraser: Yeah maybe we should require that the measure type be `Boolean` to convert it back so the user has to provide a function. I do think if the user is using a data type that is subject to floating point precision like that they are probably aware of the issues with it.

docs/newtypes/index.md
adamgfraser: Fixed.

docs/newtypes/index.md
adamgfraser: Good idea! Done now.

docs/newtypes/index.md
adamgfraser: I think I discuss that right below?

docs/newtypes/index.md
adamgfraser: Fixed.

docs/newtypes/index.md
adamgfraser: Fixed.

docs/newtypes/index.md
adamgfraser: Added.

docs/functionaldatatypes/validation.md
adamgfraser: I think it would be a good addition. Will add in a separate PR.

docs/newtypes/index.md
adamgfraser: Good idea. Added a couple sentences about that.

docs/zpure/index.md
adamgfraser: Added something about this.

docs/zpure/index.md
adamgfraser: Done!

docs/functionalabstractions/index.md
adamgfraser: Went with two categories.

docs/zpure/index.md
adamgfraser: Yeah it is an in memory log so if you think you are going to log more stuff than can fit in memory then you need to use `ZIO` to manage sending it somewhere.

docs/functionalabstractions/index.md
adamgfraser: Fixed.

docs/functionalabstractions/concretetypes/index.md
adamgfraser: Adjusted titles.

docs/functionalabstractions/concretetypes/index.md
adamgfraser: It is associative subject to an equivalence relationship and that will be done automatically in ZIO 2.0.

docs/functionalabstractions/index.md
adamgfraser: No I think that is good. Changed along those lines.

docs/functionalabstractions/concretetypes/index.md
adamgfraser: I think maybe it is more appropriate for the detailed section than this summary.

docs/functionalabstractions/concretetypes/equal.md
adamgfraser: Added.

docs/functionalabstractions/concretetypes/ord.md
adamgfraser: One of the properties a total ordering has to have is it is a transitive relationship. You're right I'm trying to talk about the properties a total ordering has to have without actually listing out all of them.

docs/functionalabstractions/concretetypes/ord.md
adamgfraser: The Scala compiler is doing the derivation but it is based on the "axioms" we give it in terms of implicit instances for primitive types and implicit definitions to derive instances for more complex types.

docs/functionalabstractions/concretetypes/debug.md
adamgfraser: Changed.

docs/functionalabstractions/concretetypes/debug.md
adamgfraser: Changed.

docs/functionalabstractions/concretetypes/associative.md
adamgfraser: Done.

adamgfraser:
This is ready for review.

docs/zpure/index.md
```diff
@@ -0,0 +1,348 @@
...
+
+There are many other error handling operators defined on `ZPure`. The `catchSome` operator allows us to only recover from certain errors.
+
+The `fold` and `foldM` operators allow us to handle both the failure and success cases at the same time. And the `orElse` operator allows us to specify a fallback computation that will be run if the original computation fails.
+
+One other concept from `ZIO` that carries over to `ZPure` is the ability to accumulate multiple errors.
+
+Normally when we use operators like `zip` or `flatMap` if an error occurs we will not go on to evaluate further parts of the computation until we get to an error handler that can potentially recover from it. This is typically what we want because if we have already failed there is no point in doing more work.
+
+However, `ZPure` also allows us to accumulate errors when we use the `zipWithPar` operator . This operator does not do actual parallelism, but is "parallel" in the sense that it will run both computations even if the first one fails and return any failures that occurred.
```
ithinkicancode, line 160: Just a comment because it might be too late for any code change and I'm sure this question must have come up before.
If we want to avoid the possible confusion for having "Par" in `zipWithPar`, would `map2` be a better choice, which would also be familiar to people who've already used similar methods in other libraries?

docs/zpure/index.md
```diff
@@ -0,0 +1,348 @@
...
+  } yield ()
+```
+
+This example combines a couple of the features we have talked about so far.
+
+First, we are using a for comprehension to chain together multiple computations
+
+Second, we are using both the error type and the state type here. If the customer has insufficient funds we fail immediately, otherwise we update the state.
+
+One important thing to note here is that since `ZPure` computations are never concurrent we don't need to worry about conflicting modifications to the state.
```
ithinkicancode, line 293: Interesting. So a ZPure computation doesn't allow/expect any concurrent operations? I gotta remember this.

docs/zpure/index.md
adamgfraser: `map2` is the equivalent of `zipWith` which runs the left computation, then runs the right computation if that completes successfully, and combines the results. `zipWithPar` runs both computations regardless of whether one or both of them fails and then combines the results.

docs/zpure/index.md
adamgfraser: Correct. A simple way to think about this is you generally shouldn't be doing concurrent operations on your own, that should be handled by a data type for you. `ZPure` doesn't have any operators that do actual concurrency, that's what you need `ZIO` for.

docs/functionaldatatypes/these.md
```diff
@@ -0,0 +1,63 @@
...
+}
+```
+
+You can think of `These` as like `Either` except that it has one additional case. Whereas an `Either` contains either an `A` or a `B` but never both, a `These` can contain both an `A` and a `B`.
+
+A simple example of a situation where `These` might arise is in merging two upstreams in a streaming application. We could be signaled when at least one upstream has data for us and could have a situation where either only the left upstream has data for us, only the right upstream has data for us, or both upstreams have data for us.
+
+Another example of using the `These` data type is in implementing a variant of the `zipAll` operator on collections.
+
+When we zip two collections, if the two collections have different sizes the default `zipWith` operator will just drop the "extra" elements from the ledger collection.
```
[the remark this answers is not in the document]
adamgfraser: @ithinkicancode Yes!

docs/functionaldatatypes/these.md
```diff
@@ -0,0 +1,63 @@
...
+
+```scala mdoc
+def zipAllWith[A, B, C](
+  as: Chunk[A],
+  bs: Chunk[B]
+)(left: A => C, right: B => C, both: (A, B) => C): Chunk[C] =
+  ???
+```
+
+Now the `both` function will be called as long as there are elements of both collections to zip together. If the left collection is longer than the `left` function will be called for the extra elements of the left collection, and if the right collection is longer than the `right` operator will be called for the extra elements of the right collection.
```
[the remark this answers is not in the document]
adamgfraser: Fixed now!

PR #666, opened 2021-07-18
Miscellaneous Cleanups

adamgfraser:
Couple of things I noticed as I was working on documentation:

- Deleted extraneous type parameters on `Validation` operators.
- Implemented `as` and `asError` on `Validation`. `asError` seems particularly useful if you have a `Validation` that doesn't have meaningful error information like one from `Validation.fromEither` and want to add some.
- Rename `Validation.collectAllPar` to `Validation.validateAll`. This is a breaking change but seems like the direction we're going with `validateWith` and the reaction to that has been pretty good.
- Removes some unnecessary constraints on `ZSet` operators.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -42,6 +42,20 @@ sealed trait ZValidation[+W, +E, +A] { self =>
     log(w1)
 
   /**
+   * Maps the successful value of this `ZValidation` to the specified constant
+   * value.
+   */
+  final def as[B](b: B): ZValidation[W, E, B] =
+    map(_ => b)
+
+  /**
+   * Maps the error value of this `ZValidation` to the specified constant
+   * value.
+   */
+  final def asError[E2](e: E2): ZValidation[W, E2, A] =
+    mapError(_ => e)
+
+  /**
    * Returns whether this `ZValidation` and the specified `ZValidation` are
    * equal to each other.
    */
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -41,6 +41,20 @@ sealed trait ZValidation[+W, +E, +A] { self =>
...
+   * value.
+   */
+  final def as[B](b: B): ZValidation[W, E, B] =
+    map(_ => b)
+
+  /**
+   * Maps the error value of this `ZValidation` to the specified constant
+   * value.
+   */
+  final def asError[E2](e: E2): ZValidation[W, E2, A] =
```
phderome, line 55: +1

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -41,6 +41,20 @@ sealed trait ZValidation[+W, +E, +A] { self =>
   final def ??[W1 >: W](w1: W1): ZValidation[W1, E, A] =
     log(w1)
 
+  /**
+   * Maps the successful value of this `ZValidation` to the specified constant
+   * value.
+   */
+  final def as[B](b: B): ZValidation[W, E, B] =
```
phderome, line 48: +1

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -375,7 +389,7 @@ object ZValidation extends LowPriorityValidationImplicits {
    * that either returns the values of all of them, if they all succeed, or
    * else fails with all of their errors.
    */
-  def collectAllPar[F[+_]: ForEach, W, E, A](validations: F[ZValidation[W, E, A]]): ZValidation[W, E, F[A]] =
+  def validateAll[F[+_]: ForEach, W, E, A](validations: F[ZValidation[W, E, A]]): ZValidation[W, E, F[A]] =
```
phderome, line 392: way better name indeed.

phderome (approved):
I like this. Good thing that some spurious inaccurate implicit requirements were identified and rectified.

I also like the new aliases for union and intersect as well as operators as and asError.

sideeffffect (approved):
Nice :+1:

PR #667, opened 2021-07-21
Implement CommutativeIdentity

adamgfraser:
Allows requiring that both a way of commutative `combine` operator and an `identity` value exist without creating ambiguities regarding the `Associative` instance.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/coherent/coherent.scala
+++ b/core/shared/src/main/scala/zio/prelude/coherent/coherent.scala
@@ -118,6 +118,16 @@ object CommutativeEqual {
     }
 }
 
+trait CommutativeIdentity[A] extends Commutative[A] with Identity[A]
+
+object CommutativeIdentity {
+  implicit def derive[A](implicit commutative0: Commutative[A], identity0: Identity[A]): CommutativeIdentity[A] =
+    new CommutativeIdentity[A] {
+      def combine(l: => A, r: => A): A = commutative0.combine(l, r)
+      def identity: A                  = identity0.identity
+    }
+}
+
 trait CovariantDeriveEqual[F[+_]] extends Covariant[F] with DeriveEqual[F]
 
 object CovariantDeriveEqual {


sideeffffect (approved):
:+1: 
I thought that requiring `Commutative[A] with Identity[A]` is enough. In what context did you need this? :thinking:

adamgfraser:
@sideeffffect Yes you're right that works too.

```scala
  def mergeAllParCommutative[R, E, A, B](
    in: Iterable[ZIO[R, E, A]]
  )(f: A => B)(implicit ev: Commutative[B] with Identity[B]): ZIO[R, E, B] =
    ZIO.mergeAllPar(in)(Identity[B].identity)((b, a) => b <> f(a)) // okay
```

It comes up when you want to have it in a type bound.

```scala
  def mergeAllParCommutative[R, E, A, B: Commutative: Identity](
    in: Iterable[ZIO[R, E, A]]
  )(f: A => B): ZIO[R, E, B] =
    ZIO.mergeAllPar(in)(Identity[B].identity)((b, a) => b <> f(a)) // ambiguous implicit values
```

adamgfraser:
Also being able to use `Commutative[A] with Identity[A]` relies on the instance for both abstractions being defined in the same value which is best practice if possible but might not be the case if both of them are being inductively derived.

```scala
  def mergeAllParCommutative[R, E, A, B](
    in: Iterable[ZIO[R, E, A]]
  )(f: A => B)(implicit ev: Commutative[B] with Identity[B]): ZIO[R, E, B] =
    ZIO.mergeAllPar(in)(Identity[B].identity)((b, a) => b <> f(a))

  case class Foo(n: Int)

  implicit val FooCommutative: Commutative[Foo] = ???
  implicit val FooIdentity: Identity[Foo] = ???

  mergeAllParCommutative(List(ZIO.succeed(1)))(n => Foo(n))  // does not compile
```

The `CommutativeIdentity` instance provides evidence to the Scala compiler that the `Associative` implementation is coherent even if the `Commutative` and `Identity` instances are separate values.

sideeffffect:
All tests are green, do you want to merge?

PR #669, opened 2021-07-25
Add AssociativeEither Instances

adamgfraser:


sideeffffect (dismissed):
Oh, I haven't even realized that these are missing :smile: 
Nice addition! :+1:

core/shared/src/test/scala/zio/prelude/AssociativeEitherSpec.scala
```diff
@@ -8,7 +8,8 @@ object AssociativeEitherSpec extends DefaultRunnableSpec {
   def spec: ZSpec[Environment, Failure] =
     suite("AssociativeEitherSpec")(
       suite("laws")(
-        testM("either")(checkAllLaws(AssociativeEither)(GenF.either(Gen.anyInt), Gen.anyInt))
+        testM("either")(checkAllLaws(AssociativeEither)(GenF.either(Gen.anyInt), Gen.anyInt)),
+        testM("associativeEither")(checkAllLaws(AssociativeEither)(GenFs.nonEmptyList, Gen.anyInt))
```
sideeffffect, line 12: Shouldn't this be `"nonEmptyList"` (instead of `"associativeEither"`)?

adamgfraser: Yes!

sideeffffect:
Scalafmt isn't happy.
But also, Scala 2.11 and 2.12 crashes. Actually, I've noticed that Scala 2.12 has been crashing like this since since 2.12.13 :disappointed: I hoped that Scala authors would notice this and fix it in 2.12.14, but apparently not. @adamgfraser are you aware of any open tickets against Scala compiler related to crashes like the one we see here?

sideeffffect:
Oh, so now everything's green :tada:

sideeffffect (dismissed):
Nice work! Thanks Adam :+1: 

One things that comes to mind: Would it make sense to define `X =>> IdentityEither[Map[_, X]]`?

adamgfraser:
I wondered about that as well.

The interesting thing there is I think there are actually two instances you could define.

The first would be partially applying the value type and then you would have a very similar version to set where it is basically take all the keys from the left map and map them to a `Left`, take all the keys from the right map and map them to a `Right`, and then throw the whole thing together.

The second would be partially applying the key type and then you would have an instance that was more like `orElse` that returned the bindings from both maps but in the case of conflicts just kept the left value, wrapping up the values in `Left` or `Right` to tell where they originally came from.

One didn't seem clearly more logical than the other so punted on it for now but was definitely thinking about the same thing.

sideeffffect (approved):

sideeffffect:
All green. Wanna merge? :)

adamgfraser:
Let's do it!

PR #670, opened 2021-07-26
Delete HashIdentityBoth and HashIdentityEither

adamgfraser:
These instances are not lawful. To be lawful the `IdentityEither` instance would have to just hash the value without including the fact that it is a `Left` or a `Right` in the hash code. The `IdentityBoth` instance would have to hash the values together using an operator that was itself associative and potentially commutative such as `^`. However, this would lead to the result that the resulting `Hash` would hash `Left` and `Right` values to the same hash code if the underlying values were the same, which seems undesirable. It would also violate the Scala hash code consistency property which seems like it has been a valuable one for us. I think it is better to just delete these instances.


sideeffffect (approved):
Interesting :thinking: OK, let's remove them.

PR #671, opened 2021-07-26
Implement ArrayDebug

adamgfraser:
Implements a `Debug` instance for `Array` given a `Debug` instance for its element type. This is a little unusual since `Array` is a mutable data structure but it is frequently used internally in functional code and the default rendering is typically not very helpful. Since all we are doing is describing a way to render an `Array` to a human readable format for debugging purposes I think this makes sense but open to other opinions.


sideeffffect (approved):
Agreed, it's good to have `Debug[Array[A]]`.
Would it make sense to have `Debug[IArray[A]]` for Scala 3?

adamgfraser:
Yes. Will add.

sideeffffect:
Are you already working on `Debug[IArray[A]]`? I can add it myself if you haven't yet started :wink:

adamgfraser:
Just pushed!

PR #682, opened 2021-08-05
Implement Max and Min Extension Methods

adamgfraser:
Available on any data type for which an `Ord` is defined.


sideeffffect (approved):
Nice addition :+1: 
Do you Adam think it would make sense to have `min`/`max` also on `Ord` proper (i.e. no only as "extension methods")? Then we could re-use the implementation in the extension methods...

sideeffffect:
All green :green_circle:

adamgfraser:
@sideeffffect Yes I think that is a good idea. Will add.

sideeffffect (approved):

sideeffffect:
All green :green_circle:

PR #683, opened 2021-08-06
Implement Bigint and BigDecimal Instances

adamgfraser:
Implements instances for `BigInt` and `BigDecimal`.


sideeffffect (approved):
Oh, it seems like we forgot about Debug and Hash/Ord when we added BigDecimal :smile: 
:+1:

sideeffffect:
All green :green_circle:

PR #684, opened 2021-08-06
Remove Identity Instances For Max And Min

adamgfraser:
While we can technically treat values like `Int.MinValue` and `Int.MaxValue` as identities for max and min I think this relies on an implementation detail to some extent. This is somewhat subjective but where it comes home for me is in something like this:

```scala
import zio.prelude._
import zio.prelude.newtypes._

val as: List[Int] =
  List.empty

as.foldMap(a => Max(a)) // -2147483648 ???
```

My expectation would be that we would instead have to do this:

```scala
as.foldMap(a => Option(Max(a))) // None
```


sideeffffect (approved):
Sounds reasonable

sideeffffect:
All green :green_circle:

PR #687, opened 2021-08-08
Add Duration (ZIO and Scala) instances to the Associative, Equal and Debug hierarchy

sideeffffect:


adamgfraser (dismissed):

adamgfraser:
@sideeffffect Good to go when tests are passing!

core/shared/src/test/scala/zio/prelude/Common.scala
```diff
@@ -6,5 +6,5 @@ import zio.test.Gen
 import scala.concurrent.duration.{Duration => ScalaDuration}
 
 object Common {
-  def anyFiniteDurationScala: Gen[Random, ScalaDuration] = Gen.long(0L, Long.MaxValue).map(ScalaDuration.fromNanos)
+  def anyFiniteDurationScala: Gen[Random, ScalaDuration] = Gen.long(0L, Long.MaxValue / 10).map(ScalaDuration.fromNanos)
```
adamgfraser, line 9: Why the change here?

sideeffffect: Scala's Duration detects _integer overflow_. Adding 2 values whose sum is > `Long.MaxValue` will throw.
https://github.com/scala/scala/blob/2.13.x/src/library/scala/concurrent/duration/Duration.scala#L616

adamgfraser: Hmmm, I think that means we can't define an `Associative` instance. What you're saying is that combining duration values doesn't even satisfy the property of closure which is a basic assumption of the different functional abstractions for describing combining operations.

sideeffffect: Ah, you're saying that we can afford to have Associative for `Long`, because even though it can return a "wrong" result (e.g. when overflowing), it still at least returns a `Long`, so it satisfies the property of closure.
But because Scala `Duration` throws in some cases, it does not satisfy the property of closure.

But in some ways, Scala `Duration` is similar to `Long` (for which we _do_ define Associative). It fails when overflowing. The only difference is that while the `Long` instance "silently" fails and overflows to negative, the instance for Scala `Duration` fails "loudly" by throwing an exception.

So, both instances are actually flawed, but throwing an exception is probably worse than overflowing...
I'll remove the Scala instance then, we can add it later if we decide to do so.

adamgfraser: Yeah. I'm not sure it is even a "wrong" result. It isn't the result of doing addition on natural numbers but it is the spec and if you understand the spec many algorithms can be correct in the presence of that. From our perspective combining two longs always gives a long and combining three longs always gives the same long, "right" or "wrong" regardless of the order of parentheses. Whereas by throwing we're saying combining two durations doesn't even necessarily produce a duration and we're throwing exceptions in a part of code that is expected to be pure and so likely isn't prepared to manage those exceptions.

sideeffffect:
All green :green_circle:

adamgfraser (approved):

PR #700, opened 2021-08-23
run `doc` on all Scala versions in CI

sideeffffect:


sideeffffect:
@mijicd you might probably want to see this too

mijicd (approved):

PR #702, opened 2021-08-24
Rename Validation#failChunk to Validation#failNonEmptyChunk

adamgfraser:
It takes a `NonEmptyChunk` rather than a `Chunk` so I think this is more appropriate.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -399,9 +399,10 @@ object ZValidation extends LowPriorityValidationImplicits {
     Failure(Chunk.empty, NonEmptyChunk(error))
 
   /**
-   * Constructs a `ZValidation` that fails with the specified chunk of errors.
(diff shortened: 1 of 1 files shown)


sideeffffect (approved):
If you're sure this is a good idea, then OK.
But IMHO I wouldn't stress the difference between `Chunk` and `NonEmptyChunk` in method names -- it's not _that_ important and makes the names longer/harder to read.
Similarly for (NonEmpty)Set/Map/whatever.

(Unless there are distinct methods -- one for NonEmpty and the other for regular.)

adamgfraser:
The current name is definitely shorter.

There are a couple of reasons that make me think this is the right thing to do:
1. In ZIO 2.0 we have really been striving for clarity and consistency even at some cost to conciseness because of the ability to use autocomplete.
2. Here I think this is more clear. In particular, if you try to call `failChunk` with a `Chunk` your code won't even compile because a `Chunk` isn't necessary a `NonEmptyChunk` so I think the current name is somewhat confusing.
3. I think it is more consistent with other names (e.g. `NonEmptySet.fromNonEmptyChunk`).

PR #708, opened 2021-08-31
Add show method to Debug

sideeffffect:
/cc @guizmaii


guizmaii:
It'll help people migrating from Cats, I think.

Thanks @sideeffffect 🙂

sideeffffect:
Btw, I think that `render` is better, because this name is consistent with other parts of the Debug machinery

adamgfraser (approved):

PR #710, opened 2021-08-31
Add Scaladoc For Debug

adamgfraser:


core/shared/src/main/scala/zio/prelude/Debug.scala
```diff
@@ -67,26 +122,27 @@ object Debug extends DebugVersionSpecific {
...
+     * Render the `Repr` to a human readable format using the specified
+     * `Renderer`.
+     */
+    def render(renderer: Renderer): String =
+      renderer(self)
 
-  sealed trait Repr { self =>
-    def render(renderer: Renderer): String = renderer(self)
-    def render: String                     = render(Renderer.Simple)
-    override def toString: String          = render // to show a nice view in IDEs, REPL, etc
```
adamgfraser, line 89: I thought this was a little confusing because it means you don't see the structured representation you just see the rendering when you are working with `Repr` which makes it hard to tell what the `Repr` is. I think with `render` already existing on `Repr` and on `Debug` we have enough tools where this shouldn't be necessary.

sideeffffect: I thought it's better this way, but I don't have basically any practical experience with `Debug`...

sideeffffect (approved):
Thank you! Really nice to see `Debug` getting some love :+1:

core/shared/src/main/scala/zio/prelude/Debug.scala
adamgfraser: Yeah I think `Repr` doesn't get used a lot other than just going to a specific rendering. The main issue here is if you are trying to debug a `Repr` instance, like because it isn't given you the rendering you want. So you want to to see what the `Repr` is but if `toString` just gives you a specific rendering then it is pretty difficult to tell what the actual structure of the `Repr` is.

PR #713, opened 2021-09-01
Implement ForEach#partitionMap

### LINKED ISSUE #709: Implement partition operations using `Alternative` and a form of `Foldable`
Given a abstraction like `Foldable`, and `Bifoldable` (or through some other mechanisms given its absence in the library for now), we would need operations such as follows:

1. `partitionEither`

Given `F[A]`, being able to decide on each value of `A` to get either a `LeftB]` or `Right[C]`, and accumulate results on both ends using `Foldable[F]` and `Alternative[F]` to produce `(F[B], F[C])`. In the absence of `Foldable` due to lack of laws, we will need to do some concrete implementations, probably using`Traverse`.

2. Extend it to `partitionEitherM` to produce `M[(F[B], F[C])]` given a Monad[M], such that this can be returned in a `Task` if needed.

3. `partitionFoldBoth`. Could we have unlawful package and feed it in  with `Foldable` and `Bifoldable` ? Example: It can easily allow us to have `partitionFoldBoth `?

```sala

  // The pair of values (`Both` part) can happen inside an `H` (Ex: `Const`, `Either`) which ultimately, which gets pealed off in the out. Example: implement `partitionEither` in terms of `partitionFoldBoth`
  def partitionFoldBoth[F[_, _], B, C](fa: F[A])( f: A => H[B, C])(implicit A: Alternative[F], F: Foldable[F], H: Bifoldable[H]): (F[B], F[C]) = ???
```

4. Extend 3 to support `Monad`ic result returning `G[(F[B], F[C])]` allowing us to return everything in `ZIO` for example.

These operations can be super handy, and this will turn out to be easy to deal with for users, given some good examples.

Please excuse if these operations are handled in a different way already by the library.

adamgfraser:
Resolves #709.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ForEach.scala
+++ b/core/shared/src/main/scala/zio/prelude/ForEach.scala
@@ -225,6 +225,40 @@ trait ForEach[F[+_]] extends Covariant[F] { self =>
     !isEmpty(fa)
 
   /**
+   * Partitions the collection based on the specified function.
+   */
+  def partitionMap[A, B, C](
+    fa: F[A]
+  )(f: A => Either[B, C])(implicit both: IdentityBoth[F], either: IdentityEither[F]): (F[B], F[C]) = {
+    implicit val covariant: Covariant[F]       = self
+    implicit val leftIdentity: Identity[F[B]]  = Identity.fromIdentityEitherCovariant
+    implicit val rightIdentity: Identity[F[C]] = Identity.fromIdentityEitherCovariant
+    foldMap(fa) { a =>
+      f(a) match {
+        case Left(b)  => (b.succeed, either.none)
+        case Right(c) => (either.none, c.succeed)
+      }
+    }
+  }
+
+  /**
+   * Partitions the collection based on the specified effectual function.
+   */
+  def partitionMapM[G[+_]: IdentityFlatten: Covariant, A, B, C](
+    fa: F[A]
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/prelude/ForEach.scala
```diff
@@ -224,6 +224,40 @@ trait ForEach[F[+_]] extends Covariant[F] { self =>
...
+      }
+    }
+  }
+
+  /**
+   * Partitions the collection based on the specified effectual function.
+   */
+  def partitionMapM[G[+_]: IdentityFlatten: Covariant, A, B, C](
+    fa: F[A]
+  )(f: A => G[Either[B, C]])(implicit both: IdentityBoth[F], either: IdentityEither[F]): G[(F[B], F[C])] = {
```
afsalthaj, line 249: Could we generalise on Either to be just another M[_] that has a bifoldable. ?

sideeffffect (approved):

core/shared/src/main/scala/zio/prelude/ForEach.scala
adamgfraser: Our equivalent would be `BiForEach` which we don't have right now but potentially could. I'm not sure it is useful enough. What data types do you typically use this with other than `Either`?

PR #716, opened 2021-09-02
Attempt To Resolve CI Flakiness

adamgfraser:


sideeffffect:
Wow. I would guess that it would be `val`s, not `lazy val`s that are causing issues. So I would be trying adding more `lazy`, but I don't really know what I'm doing :smile: 

But apparently we still get 2 stalled builds :disappointed:

sideeffffect (approved):
Feel free to go ahead with this experiment :crossed_fingers:

adamgfraser:
Well these failures indicate that my hypothesis was clearly wrong. I'm not sure what the cause is but I can tell you that I have not seen this behavior in any other ZIO ecosystem project so I think it is something we are doing and not ZIO or Github Actions.

sideeffffect:
Or maybe we're pushing _something_ to the point it sometimes break. _something_ could be Scala's `lazy val`s, ZIO, ... :shrug: I really have no idea :smile_cat:

PR #717, opened 2021-09-02
Implement zipAllWith

### LINKED ISSUE #714: Have ZipThese that can emit a disjunction especially to deal with collections
```scala
  trait ZipThese[F[_]] {
     def zipThese[A, B](fa: F[A], fb: F[B]): F[These[A, B]]  
  }

```

Given the primitive `zipThese`, we are able to do a variety of derived combinators such as `zipAll` ?

Example:

```scala
  def zipAll[A, B](fa: F[A], fb: F[B], a: A, b: B): F[(A, B)] = ???
```

```scala
import zipThese._

 val m1 = Map("AU" -> 1.0)
 val m2 = Map("AU" -> 2.0,   "US" -> 3.0)
 
ZipThese[MapF].zipAll(m1, m2, 0.0, 0.0)

//res: Map[String, (Double, Double)] = Map("AU" -> (1.0, 2.0), "US" -> (0.0, "2.0"))
```

`ZipThese` can be `Align` in classic package if needed.

adamgfraser:
Resolves #714.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ForEach.scala
+++ b/core/shared/src/main/scala/zio/prelude/ForEach.scala
@@ -305,7 +305,10 @@ trait ForEach[F[+_]] extends Covariant[F] { self =>
    * Zips the left collection and right collection together, using `None` to
    * handle the case where one collection is larger than the other.
    */
-  def zipAll[A, B, C](fa: F[A], fb: F[B])(implicit both: IdentityBoth[F], either: IdentityEither[F]): F[(Option[A], Option[B])] =
+  def zipAll[A, B, C](fa: F[A], fb: F[B])(implicit
+    both: IdentityBoth[F],
+    either: IdentityEither[F]
+  ): F[(Option[A], Option[B])] =
     zipAllWith(fa, fb) {
       case These.Both(a, b) => ((Some(a), Some(b)))
       case These.Left(a)    => ((Some(a), None))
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/zio/prelude/ForEach.scala
```diff
@@ -301,6 +301,35 @@ trait ForEach[F[+_]] extends Covariant[F] { self =>
...
     foldLeft(fa)(List.empty[A])((as, a) => a :: as).reverse
 
+  /**
+   * Zips the left collection and right collection together, using `None` to
+   * handle the case where one collection is larger than the other.
+   */
+  def zipAll[A, B, C](fa: F[A], fb: F[B])(implicit
+    both: IdentityBoth[F],
+    either: IdentityEither[F]
+  ): F[(Option[A], Option[B])] =
```
sideeffffect, line 311: Why not just `F[These[A, B]]` (instead of `F[(Option[A], Option[B])]`)?

sideeffffect (dismissed):
Looks good, just one suggestion

core/shared/src/main/scala/zio/prelude/ForEach.scala
adamgfraser: Good idea. I originally did it this way to mirror the signature of `zipAll` on `Chunk` but you're right that is a more precise signature.

sideeffffect (approved):
Nice :+1:

PR #721, opened 2021-09-06
use the correct map to fold over in the example

olambo:
the example is not currently looking at the correct map to determine existing Votes


adamgfraser (approved):
Thank you! 🙏

PR #727, opened 2021-09-10
Enable fatal-warnings

sideeffffect:
Scalafix, when run on its own, suppresses `fatal-warnings`, so that it can operate smoothly (and not get shot down because there is a warning in the codebase).

But in our case, we want to verify both that Scalafix is happy _and_ that the codebase doesn't contain any warnings.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -17,7 +17,7 @@ inThisBuild(
 )
 
 addCommandAlias("fix", "; all compile:scalafix test:scalafix; all scalafmtSbt scalafmtAll")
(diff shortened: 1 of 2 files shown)


adamgfraser (dismissed):
We might want to think about whether there is another example we can use here. I think the purpose of this is to show that validation works with values that are only available at runtime in addition to those that are available at compile time but it goes a little against the lessons we want to teach about suspending effects.

Copying @kitlangton.

adamgfraser (approved):

PR #742, opened 2021-09-18
Implement Newtype.derive

adamgfraser:
Provides a convenience method for constructing type class instances for new types given type class instances for their underlying types.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/NewtypeModule.scala
+++ b/core/shared/src/main/scala/zio/prelude/NewtypeModule.scala
@@ -177,6 +177,14 @@ private[prelude] sealed trait NewtypeModule {
     type Type
 
     /**
+     * Derives an instance of a type class for the new type given an instance
+     * of the type class for the underlying type. The caller is responsible for
+     * the type class being a valid instance for the new type.
+     */
+    protected def derive[TypeClass[_]](ev: TypeClass[A]): TypeClass[Type] =
+      ev.asInstanceOf[TypeClass[Type]]
+
+    /**
      * Allows pattern matching on newtype instances to convert them back to
      * instances of the underlying type.
      */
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/prelude/NewtypeModule.scala
```diff
@@ -176,6 +176,14 @@ private[prelude] sealed trait NewtypeModule {
   sealed trait Newtype[A] extends NewtypeVersionSpecific[A] {
     type Type
 
+    /**
+     * Derives an instance of a type class for the new type given an instance
+     * of the type class for the underlying type. The caller is responsible for
+     * the type class being a valid instance for the new type.
+     */
+    protected def derive[TypeClass[_]](ev: TypeClass[A]): TypeClass[Type] =
```
hmemcpy, line 184: Shouldn't this be `implicit ev: TypeClass[A]` to leverage the compiler for summoning the implicit instance?

adamgfraser: Yes! Was playing with the signature and accidentally deleted the most important part!

hmemcpy:
This would definitely make it more usable for defining a "base" Newtype with some predefined typeclass instances. Ideally this would work similarly to the semi-auto approach of the @newtype macro, by calling e.g.:

```scala
object Widget extends Newtype[Long] {
  implicit val codec: Codec[Widget] = deriving
```

This works (except minor type mismatch warning issues that IntelliJ shows), and would allow to define a bunch of instances with minimal boilerplate...

Perhaps for Scala 3 this could be leveraged, somehow, to use the `derives` keyword?

```scala
object Widget extends Newtype[Long] derives Codec
```

kitlangton (approved):
Awesome!

sideeffffect (approved):

PR #790, opened 2021-11-04
Better error message when forgetting "inline" in Scala 3 Newtype

kitlangton:
CleanShot 2021-11-04 at 10 31 36@2x


adamgfraser (approved):

adamgfraser:
Looks like your warnings were too successful and caused CI to fail but otherwise looks great! 😃

adamgfraser (approved):

PR #793, opened 2021-11-07
Allow custom assertion in scala 2 and 3

mschuwalow:


sideeffffect:
Scala 3 doesn't seem to be happy. Do you think you could have a look at it?

mschuwalow:
> Scala 3 doesn't seem to be happy. Do you think you could have a look at it?

Was a bit busy during the week :(
Updated now. It would be good to make this a bit more ergonomic, but I couldn't come up with anything

mijicd:
Any updates on this one? :eyes:

mschuwalow:
I'll rebase this on top of latest master.

@kitlangton @adamgfraser @sideeffffect maybe we can find some time during the hackathon to pair on this? We are still very interested in having this functionality

sideeffffect:
I'll be happy to help! Let's see how things will be after the rebase :wink:

core/shared/src/main/scala-3/zio/prelude/Validator.scala
```diff
@@ -0,0 +1,26 @@
...
+    assertion(value).as(value)
+
+  final def validateInlineImpl(expr: Expr[A])(using Quotes): Expr[A] =
+    expr.value match {
+      case Some(a) =>
+        assertion(a) match {
+          case Right(()) => expr
+          case Left(err) =>
+            quotes.reflect.report.error(err.render(a.toString))
+            ???
```
sideeffffect, line 19: I assume these `???` are intentional here.

sideeffffect (dismissed):
Looks good to me, thanks Max! @adamgfraser do you also want to have a look at this?

adamgfraser:
@kitlangton Do you want to take a look at this? Maybe we can leverage your recent work on unifying the classic and smart assertions in ZIO Test?

core/shared/src/main/scala-3/zio/prelude/Newtype.scala
```diff
@@ -5,14 +5,38 @@ import zio.NonEmptyChunk
 import scala.quoted.*
 import zio.prelude.ConsoleUtils.*
 
-abstract class Newtype[A] {
+abstract class NewtypeCustom[A] {
```
sideeffffect, line 8: Do we want to make it private? Because we don't have an `NewtypeCustom` for Scala 2 flavor, right? By making this public, we're introducing an asymmetry between Scala 2 and Scala 3 flavors of ZIO Prelude.

sideeffffect (commented):
Besides the `NewtypeCustom` comment, I think this looks good from what I can tell (which is not much, since I'm no macro expert :smile_cat: )
Any comments, @kitlangton ?

core/shared/src/main/scala-3/zio/prelude/Newtype.scala
adamgfraser: I think the goal of this PR is to expose this functionality for end users to define their own custom compile time assertions (without forcing the additional complexity of doing so on normal users) so I think this needs to be public.

My question would be whether we can expose similar functionality in Scala 2? If not I don't think it is the end of the world if this is a Scala 3 only feature if we don't think there is a feasible way to support an equivalent on Scala 2.

mschuwalow: For scala 2 no extra ceremony should be necessary https://github.com/zio/zio-prelude/pull/793/files#diff-cb2c8d5aab7fb0cedcd2c4d8dbfad7777e0a9761ea3f7a5f09ad205695825880R205-R210

https://github.com/zio/zio-prelude/pull/793/files#diff-109e476632c38a367dd0aa6740e67ad6801296fbed9592f638360a41d2334922R67-R73

core/shared/src/main/scala-3/zio/prelude/Newtype.scala
```diff
@@ -72,98 +91,13 @@ abstract class Newtype[A] {
...
 
-          case _ =>
-              report.errorAndAbort(s"NO VARARGS!?")
-        }
-      
-      case None =>
-        report.errorAndAbort(s"NO REFINEMENT!?")
-    }
-  }
+  inline def validateInline(inline value: A): A =
```
sideeffffect, line 100: Btw, can this implementation be "upstreamed" from `Newtype` to `NewtypeCustom`? Are `inline` members overridable?

mschuwalow: No, sadly they are effectively final

sideeffffect (dismissed):

mschuwalow:
@kitlangton can this be merged?

macros/shared/src/main/scala/zio/prelude/Assertion.scala
```diff
@@ -76,6 +76,10 @@ object Assertion {
 
   def matches(regexString: String): Assertion[String] = Matches(regexString)
 
+  def predicate[A](f: A => Boolean): Assertion[A] = Predicate(f, "custom function")
+
+  def predicate[A](f: A => Boolean, name: String): Assertion[A] = Predicate(f, name)
+
```
kitlangton, line 82: The one issue I foresee (And, so so so so sorry for taking so damn long to respond 😅) is that the `Assertion` DSL was intentionally hamstrung in order to be completely "Unliftable" (able to be parsed and run at compile-time). I think we discussed once before that while this predicate type _can_ be read at compile-time with some trickery, that is only valid if the predicate is defined in a separate compilation cycle.

I previously added compile-errors to guide the user through this delicate process (in this file core/shared/src/main/scala-3/zio/prelude/Macros.scala), but if we add this `predicate` method, I think we'd at least need to further document that potential edge case. That's one option. Perhaps easier to implement, just some guidance in the case the user doesn't define their assertion in a separate compilation unit (which I fear might be the common case).

The other option would be to completely separate these two modes. Have your super-powered, super-flexible assertion _not_ use this DSL. And your validation type could accept a predicate directly. I wouldn't even bother with having an assertion DSL. In fact, I think that might be the better experience. The assertion DSL, while a step up from the type-level DSL (IMHO), is still not as nice as just writing normal scala functions—which your implementation allows for 😄 

I somewhat prefer the second approach. I think it's a bit cleaner, so when you're using one implementation, it isn't made slightly trickier or more awkward by having to deal with the concerns of the other.

I hope that all makes sense! And once again, apologies for the huge delay. Also, I might be not fully understanding the code, and completely open to discussion. Happy to do any pairing, or hop on a zoom call some time (maybe with our buddy @adamgfraser as well 👍 ) and helping to push this over the line. Thanks again for all your efforts,  Maxim. <3

core/shared/src/main/scala-3/zio/prelude/Validator.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import scala.quoted._
+
+abstract class Validator[A](
+  val assertion: Assertion[A]
+) extends Liftables {
+
+  final def validateInlineImpl(expr: Expr[A])(using Quotes, Type[A]): Expr[A] = {
+    expr match {
+      case LiteralUnlift(a) =>
```
mschuwalow, line 11: IMO it would be nicer to use FromExpr here, but for now I stayed consistent with the existing macro

macros/shared/src/main/scala/zio/prelude/Assertion.scala
mschuwalow: > The one issue I foresee (And, so so so so sorry for taking so damn long to respond 😅) is that the Assertion DSL was intentionally hamstrung in order to be completely "Unliftable" (able to be parsed and run at compile-time). I think we discussed once before that while this predicate type can be read at compile-time with some trickery, that is only valid if the predicate is defined in a separate compilation cycle.

> I previously added compile-errors to guide the user through this delicate process (in this file core/shared/src/main/scala-3/zio/prelude/Macros.scala), but if we add this predicate method, I think we'd at least need to further document that potential edge case. That's one option. Perhaps easier to implement, just some guidance in the case the user doesn't define their assertion in a separate compilation unit (which I fear might be the common case).

The way it works in this pr is that in dotty we would fail compilation, pointing out to the user that they need to switch to NewtypeCustom https://github.com/zio/zio-prelude/pull/793/files#diff-923a5a88adf9d9ce695dd7622dea3b6bb66a51ced1e3a88f939a7129087dea52R88. Potentially it makes sense to have more guidance for the user here, though I have no good ideas.

For scala 2 we can extract the function using eval, which seems to work even for functions defined in the same compilation unit. 
https://github.com/zio/zio-prelude/pull/793/files#diff-cb2c8d5aab7fb0cedcd2c4d8dbfad7777e0a9761ea3f7a5f09ad205695825880R206
Usage here:
https://github.com/zio/zio-prelude/pull/793/files#diff-109e476632c38a367dd0aa6740e67ad6801296fbed9592f638360a41d2334922R69
I'm not sure about the limitations of eval / I never managed to get it to blow up so far. Do we need additional safeguards here?

> The other option would be to completely separate these two modes. Have your super-powered, super-flexible assertion not use this DSL. And your validation type could accept a predicate directly. I wouldn't even bother with having an assertion DSL. In fact, I think that might be the better experience. The assertion DSL, while a step up from the type-level DSL (IMHO), is still not as nice as just writing normal scala functions—which your implementation allows for 😄

> I somewhat prefer the second approach. I think it's a bit cleaner, so when you're using one implementation, it isn't made slightly trickier or more awkward by having to deal with the concerns of the other.

An argument for keeping the assertion dsl is that the error messages in case of failed validation are much nicer this way.
It also allows users to combine the two, i.e. use normal assertions to catch 90% of invalid input with great error messages and use a lifted library function for the remaining ones. I could see something like this being useful for phone numbers or similar, though this is a bit of a contrived example.

Potentially it makes sense to rename `predicate` to `custom` to make it clearer that this is really an escape hatch. I'll do that and add a short docstring to it explaining the caveats.

> I hope that all makes sense! And once again, apologies for the huge delay. Also, I might be not fully understanding the code, and completely open to discussion. Happy to do any pairing, or hop on a zoom call some time (maybe with our buddy @adamgfraser as well 👍 ) and helping to push this over the line. Thanks again for all your efforts, Maxim. <3

Sounds great to me. Will ping you on discord 👍

mschuwalow: I played a bit around with it a bit more and I think you are right, having them be completely separate is the better way to go.
Updated the pr, please take another look

core/shared/src/main/scala-2/zio/prelude/Newtype.scala
```diff
@@ -225,7 +225,7 @@ abstract class Newtype[A] extends NewtypeVersionSpecific {
 
   /**
    * This method is used to generate Newtype that can be validated at
-   * compile-time. This must wrap a [[Assertion]] and be assigned to
```
mschuwalow, line 228: I'm not sure what broke this link in the scaladoc.
After this pr it started failing with `Could not find any member to link for "Assertion".`  for 2.11. All other scala versions are fine.

kitlangton: Scala's weird sometimes :D

kitlangton (approved):
Wee hoo!

adamgfraser:
@mschuwalow Awesome work on this!

core/shared/src/main/scala-3/zio/prelude/Newtype.scala
guizmaii: @mschuwalow Interestingly, I think I've discovered an issue with `SubtypeCuston`. It doesn't seem to behave like a `Subtype` but like a `Newtype`

See:
- https://github.com/zio/zio-prelude/pull/1327
- https://github.com/zio/zio-prelude/actions/runs/9316877282/job/25645999588?pr=1327

Edit:
Fixed here: https://github.com/zio/zio-prelude/pull/1328

PR #796, opened 2021-11-09
Add Assertion Variants

adamgfraser:


sideeffffect:
Scala 3 doesn't like it :disappointed: ... Maybe a workaround, like renaming to `isSuccessVAnything` is needed? :shrug:

adamgfraser:
Yes. Let's see if there is anything we can do with the smart types first since similar variants work in ZIO Test without the smart types. But if not we can always fall back on renaming.

adamgfraser:
@kitlangton Can you take a look at this when you get the chance?

sideeffffect:
This may be worth resurrecting :thinking:  /cc @jdegoes

PR #798, opened 2021-11-11
Adds Associative.intercalate and ForEach.intercalate

SimY4:
Adding some useful helper methods to ForEach and Associative type classes

### DIFF
--- a/core-tests/shared/src/test/scala/zio/prelude/ForEachSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/prelude/ForEachSpec.scala
@@ -153,6 +153,13 @@ object ForEachSpec extends DefaultRunnableSpec {
             assert(actual)(equalTo(expected))
           }
         },
+        testM("intersperse") {
+          check(genList, genInt) { (as, s) =>
+            val actual   = ForEach[List].intersperse(as.map(Sum.apply), Sum(s))
+            val expected = Sum(as.sum + math.max(0, as.size - 1) * s)
+            assert(actual)(equalTo(expected))
+          }
+        },
         testM("map") {
           check(genList, genIntFunction) { (as, f) =>
             val actual   = ForEach[List].map(f)(as)
--- a/core/shared/src/main/scala/zio/prelude/Associative.scala
+++ b/core/shared/src/main/scala/zio/prelude/Associative.scala
@@ -45,8 +45,14 @@ import scala.annotation.tailrec
  * can be combined in ways that are associative, commutative, and have an
  * identity element, supporting much more interesting modes of composition.
  */
-trait Associative[A] {
+trait Associative[A] { self =>
   def combine(l: => A, r: => A): A
+
+  final def intersperse(middle: A): Associative[A] =
+    new Associative[A] {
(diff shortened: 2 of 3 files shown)


core/shared/src/main/scala/zio/prelude/ForEach.scala
```diff
@@ -167,6 +167,16 @@ trait ForEach[F[+_]] extends Covariant[F] { self =>
   def isEmpty[A](fa: F[A]): Boolean =
     foldMap(fa)(_ => And.create(false))
 
+  /**
+   * Folds over the elements of this collection using an associative operation
+   * with the middle element intercalated between every element.
+   */
+  def intercalate[A](fa: F[A], middle: A)(implicit I: Identity[A]): A =
```
adamgfraser, line 174: `intersperse`?

core/shared/src/main/scala/zio/prelude/ForEach.scala
```diff
@@ -416,6 +426,8 @@ trait ForEachSyntax {
       F.forEach_(self)(f)
     def isEmpty(implicit F: ForEach[F]): Boolean                                                                =
       F.isEmpty(self)
+    def intercalate[A](fa: F[A], middle: A)(implicit I: Identity[A]): A                                         =
```
adamgfraser, line 429: We don't want to take an `fa` here but instead should call this on `self` from above.

core-tests/shared/src/test/scala/zio/prelude/ForEachSpec.scala
```diff
@@ -153,6 +159,13 @@ object ForEachSpec extends DefaultRunnableSpec {
             assert(actual)(equalTo(expected))
           }
         },
+        testM("intercalate") {
```
adamgfraser, line 162: This test will be significantly faster using integers instead of strings.

core/shared/src/main/scala/zio/prelude/Associative.scala
```diff
@@ -45,9 +45,15 @@ import scala.annotation.tailrec
  * can be combined in ways that are associative, commutative, and have an
  * identity element, supporting much more interesting modes of composition.
  */
-trait Associative[A] {
+trait Associative[A] { self =>
   def combine(l: => A, r: => A): A
 
+  final def intercalate(middle: A): A =
```
adamgfraser, line 51: Shouldn't the return type here be `Associative[A]`?

adamgfraser (changes requested):
Thanks for adding! See comments below.

core/shared/src/main/scala/zio/prelude/ForEach.scala
SimY4: Thank you for reviewing this PR so quickly. And I'm sorry I rushed it too fast. I should've caught it myself. I'll fix it up.

sideeffffect:
Needs small changes to pass CI @SimY4 :wink:

adamgfraser (approved):
Congratulations on your first contribution to ZIO Prelude!

PR #799, opened 2021-11-12
Implement ZValidation.partition

adamgfraser:


sideeffffect (approved):
Staying in ZValidation to keep the logs, that's smart! :+1: 
I have opened a PR that adds a method `.value` that can pull out the successful value if `E <: Nothing` that should complement this nicely.
https://github.com/zio/zio-prelude/pull/802

sideeffffect (commented):
Perhaps we could also add it as extension method on `ForEach`? :thinking: 
What do you think, Adam?

adamgfraser:
Yes, I think we could have a `partitionMapV` variant to go along with the existing `partitionMap`.

sideeffffect (approved):

PR #803, opened 2021-11-13
ZValidation: add toEitherException, toFuture and make toTry not discard errors

sideeffffect:
Related issue https://github.com/zio/zio-prelude/issues/791

Lighter version of these PRs:
closes https://github.com/zio/zio-prelude/pull/610
closes https://github.com/zio/zio-prelude/pull/603

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -208,6 +208,14 @@ sealed trait ZValidation[+W, +E, +A] { self =>
     fold(Left(_), Right(_))
 
   /**
+   * Transforms `ZValidation` to an `Either`, where the failure case will be represented as an `Exception` with the first error as `cause`.
+   */
+  final def toEitherException(implicit ev: E <:< Throwable): Either[Failure.Exception[W, E], A] = self match {
+    case Success(_, value)       => Right(value)
+    case failure @ Failure(_, _) => Left(failure.toException)
+  }
+
+  /**
    * Transforms this `ZValidation` to an `Either`, discarding the order in which the errors occurred and discarding the log.
    */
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -287,9 +304,27 @@ sealed trait ZValidation[+W, +E, +A] { self =>
...
+    lazy val message: String =
+      s"errors:\n  ${errors.map(_.toString).mkString("\n  ")}\nlog:\n  ${log.map(_.toString).mkString("\n  ")}"
+
+    /** The same failure, but represented as an `Exception` which also has the `cause` set to the first error that occurred */
+    def toException(implicit ev: E <:< Throwable): Failure.Exception[W, E] = Failure.Exception(this)
   }
-  final case class Success[+W, +A](log: Chunk[W], value: A)                 extends ZValidation[W, Nothing, A]
+
+  object Failure {
+    final case class Exception[+W, +E](failure: Failure[W, E])(implicit ev: E <:< Throwable)
```
adamgfraser, line 320: We're exposing a lot of implementations details here and we are also potentially creating some inaccurate information. The returned `Throwable` is a subtype of `RuntimeException` even though it is possible that none of the original throwable are subtypes of `RuntimeException`. I would suggest we just take the first error and then add the others as suppressed exceptions to that, discarding the log. There are a variety of other operators that users can use to get the log if they want.

sideeffffect: You mean to mutate an existing Exception by adding others as suppressed? Wouldn't that be a bit misleading? How could users recognize the original suppressed exceptions in that particular Exception from those that we would add?

Or do you think it would be better to use a mere `Exception` instead of `RuntimeException`.

I don't have a strong opinion on this, so let me know, if you still think adding rest as suppressed to the first one is the best way to go.

PR #811, opened 2021-11-24
add fromOptions constructor for the These data type

yisraelU:
I came across this use case , where I had 2 options and found it easier to work with an option of These .
Does this make sense ?


sideeffffect (changes requested):
What a nice contribution! :heart: 
Please run `sbt fix` to format the codebase so that the build is green -- it's not formatted properly
https://github.com/zio/zio-prelude/runs/4328333237?check_suite_focus=true
After that, we can go ahead and merge it, right @adamgfraser ?

sideeffffect (approved):

sideeffffect:
Thanks again Yisrael :+1:

yisraelU:
Thank you, hope to contribute some more.

sideeffffect:
Looking forward to it! :muscle:

PR #813, opened 2021-11-26
Fold method 

yisraelU:
Added a fold method to the These trait to allow reduction given an Associative on the target  type


core/shared/src/main/scala/zio/prelude/These.scala
```diff
@@ -31,6 +31,14 @@ sealed trait These[+A, +B] { self =>
       case (Both(l, r1), Right(r2))     => Both(l, B1.combine(r1, r2))
       case (Both(l1, r1), Both(l2, r2)) => Both(A1.combine(l1, l2), B1.combine(r1, r2))
     }
+
+  def fold[A1 >: A](f: A => A1, g: B => A1)(implicit A1: Associative[A1]): A1 =
```
adamgfraser, line 35: I think the most idiomatic name for this operator is `reduceMap`. This would also let use `fold` for the traditional operator that let's the user specify functions to handle the `left`, `right`, and `both` cases.

regiskuckaertz, line 35:
```suggestion
  def fold[A1](f: A => A1, g: B => A1)(implicit A1: Associative[A1]): A1 =
```
Unless I'm missing something, the result type can be anything.

adamgfraser: That is 💯  right. There is another variant `foldAssociative` that would not take any user provided function and require `A1 >: A` but for this version I would probably call it `C` for clarity and it can be completely unconstrained.

yisraelU: @regiskuckaertz thanks for catching that.
Not sure what I was thinking 🤔

yisraelU: updated

adamgfraser (approved):
Thank you! 🙏

PR #815, opened 2021-11-30
Update to ZIO 2.0.0-M6-2

sideeffffect:
This is still WIP, because I'm struggling a bit with this update.
Something, can be easily fixed on the ZIO side, like https://github.com/zio/zio/pull/6112
But some things are more serious, like the new encoding of environment (`ZEnvironment`). It makes it (almost, hopefully) impossible to define many instances, because it often requires a `Tag` or `IsNotIntersection` which are not available (AFAIK).
If you had some time @adamgfraser @kitlangton , I'd greatly appreciate if you could give this a whirl too, since I suppose you're more experienced with this new `ZEnvironment` business. :pray:

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/AssociativeCompose.scala
+++ b/core/shared/src/main/scala/zio/prelude/AssociativeCompose.scala
@@ -16,6 +16,7 @@
 
 package zio.prelude
 
+import zio.ZLayer.ZLayerProvideSomeOps
 import zio._
 
 trait AssociativeCompose[=>:[-_, +_]] {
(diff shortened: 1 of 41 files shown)


core/shared/src/main/scala/zio/prelude/AssociativeCompose.scala
```diff
@@ -45,21 +46,24 @@ object AssociativeCompose {
   }
 
   implicit val URIOIdentityCompose: IdentityCompose[URIO] = new IdentityCompose[URIO] {
-    def identity[A]: URIO[A, A] = URIO.environment
+    def identity[A]: URIO[A, A] = URIO.service[A]
```
sideeffffect, line 49: For example, this gives
```
zio-prelude/core/shared/src/main/scala/zio/prelude/AssociativeCompose.scala:49:47: could not find implicit value for izumi.reflect.Tag[A]. Did you forget to put on a Tag, TagK or TagKK context bound on one of the parameters in A? e.g. def x[T: Tag, F[_]: TagK] = ...
[error] 
[error] 
[error] <trace>: 
[error]   deriving Tag for A, dealiased: A:
[error]   could not find implicit value for Tag[A]: A is a type parameter without an implicit Tag!
[error]     def identity[A]: URIO[A, A] = URIO.service[A]
[error]                                               ^
```

core/shared/src/main/scala/zio/prelude/AssociativeEither.scala
```diff
@@ -218,7 +218,7 @@ object AssociativeEither {
   implicit def ZLayerAssociativeEither[R, E]: AssociativeEither[({ type lambda[+a] = ZLayer[R, E, a] })#lambda] =
     new AssociativeEither[({ type lambda[+a] = ZLayer[R, E, a] })#lambda] {
       def either[A, B](fa: => ZLayer[R, E, A], fb: => ZLayer[R, E, B]): ZLayer[R, E, Either[A, B]] =
-        fa.map(Left(_)) orElse fb.map(Right(_))
+        fa.map(e => ZEnvironment(Left(e.get[A]))) orElse fb.map(e => ZEnvironment(Right(e.get[B])))
```
sideeffffect, line 221: This gives
```
zio-prelude/core/shared/src/main/scala/zio/prelude/AssociativeEither.scala:221:44: Cannot find implicit for IsNotIntersection[A]
[error]         fa.map(e => ZEnvironment(Left(e.get[A]))) orElse fb.map(e => ZEnvironment(Right(e.get[B])))
[error]                                            ^
```

core/shared/src/main/scala/zio/prelude/AssociativeCompose.scala
adamgfraser: Yeah `URIO[A, A]` is not an identity here since it accesses a specific service from the environment. It is basically `ZEnvironment[A] => A` at this point and is only defined if there is a tag for `A` and `A` is a single service. `URIO[A, ZEnvironment[A]]` would be closer to an identity since that would be conceptually `ZEnvironment[A] => ZEnvironment[A]`.

core/shared/src/main/scala/zio/prelude/AssociativeEither.scala
adamgfraser: This is not going to be well defined anymore because conceptually you could say that give a layer that produces `ZEnvironment[A]` and a layer that produces `ZEnvironment[B]` we can either produce `ZEnvironment[A]` or `ZEnvironment[B]`. But the actual signature says we can produce a `ZEnvironment[Either[A, B]]` which indicates an environment containing a single either `A` or `B` rather than an environment containing potentially a whole set of services `A` or an environment containing a whole different set of services `B`.

adamgfraser:
With the deletion of `Has` we are intentionally moving away from the environment just being a reader effect so would expect that some of these operators would no longer be well defined and in fact their corresponding versions on ZIO itself were previously deprecated.

sideeffffect:
Casualities are

### ZLayer

 * ZLayerIdentityBoth
 * ZLayerCommutativeBoth
 * ZLayerCovariant
 * ZLayerFailureCovariant
 * ZLayerContravariant
 * ZLayerZivariant
 * URLayerApplicationCompose
 * URLayerEitherCompose

### ZIO
 * ZIOContravariant
 * URIOApplicationCompose
 * URIOEitherCompose

### ZManaged
 * ZManagedContravariant
 * URManagedApplicationCompose
 * URManagedEitherCompose

### ZStream
 * ZStreamContravariant

### ZSTM
 * ZSTMZivariantContravariant

sideeffffect:
Please have another look :pray:

adamgfraser:
Love the last commit message!

PR #819, opened 2021-12-02
constructor from either and fold method

yisraelU:
Added a constructor to lift either into a These
Added a fold method that takes 3 user defined functions
A few questions
should the fold method be curried?
Also what about adding map,flatmap, and friends so These can be used in a for comprehension? 
Does this make sense  in terms of the goal of These


core/shared/src/main/scala/zio/prelude/These.scala
```diff
@@ -32,6 +32,13 @@ sealed trait These[+A, +B] { self =>
       case (Both(l1, r1), Both(l2, r2)) => Both(A1.combine(l1, l2), B1.combine(r1, r2))
     }
 
+  def fold[C](left: A => C, right: B => C, both: (A, B) => C): C =
```
adamgfraser, line 35: I think I would curry the `both` parameter since that is a function from two arguments two one while the others are functions from one argument to one though that is slightly subjective. That is also consistent with methods like the signature of operators like `ZStream.zipAllWith`.

core/shared/src/main/scala/zio/prelude/These.scala
```diff
@@ -108,4 +115,10 @@ object These {
       case (_, Some(b))       => Some(Right(b))
       case _                  => None
     }
+
+  /**
+   * constructor for lifting Either into a These
```
adamgfraser, line 120: Very minor but capitalization and punctuation on documentation here.

adamgfraser:
#229 adds those operators. I will update it so we can review it and get it merged.

core/shared/src/main/scala/zio/prelude/These.scala
yisraelU: Thanks for the review.
Will do

adamgfraser (approved):

PR #823, opened 2021-12-04
Build Out Recursion Schemes

adamgfraser:


afsalthaj:
Thanks Adam for this.

sideeffffect (approved):

core/shared/src/main/scala/zio/prelude/recursive/Annotated.scala
```diff
@@ -0,0 +1,7 @@
+package zio.prelude.recursive
+
+/**
+ * An `Annotated[Case, A]` is a recursive data structure where each case `Case`
+ * has been annotated with some additional information `A`.
+ */
+final case class Annotated[Case[+_], A](caseValue: Case[Annotated[Case, A]], annotations: A)
```
jdegoes, line 7: Consider also: ZEnvironment.

adamgfraser: Ah, make it `annotations: ZEnvironment[A]` so you can "look up" different annotations? I like that.

PR #851, opened 2022-01-16
Add Annotations To Recursive

### LINKED ISSUE #850: Add annotations to Recursive via ZEnvironment
`Recursive` can have a `annotations` field which is a `ZEnvironment`, initially empty. This will facilitate adding arbitrary type-directed annotations to nodes of a recursive structure.

/cc @DamianReeves

adamgfraser:
Resolves #850.


sideeffffect (approved):
Broken formatting, but nothing what a little `fix` won't handle :wink: 
Otherwise good from what I can tell. But maybe you would like John to have a look at it as well.

DamianReeves:
This looks great and good to go

PR #854, opened 2022-01-22
Create Specialized Variants For Effectual Folds Over Recursive Data Structures

adamgfraser:
We can make these operators more accessible by creating specialized methods for various ZIO data types in addition to the more general `foldM`.


sideeffffect (commented):
Looks good. But why don't we target `master` instead of `zio-2`? :thinking:

adamgfraser:
@sideeffffect I think we're just starting to diverge between the two branches with `ZEnvironment` in ZIO 2.0 and the way `Annotated` is encoded with it so thought it made sense to do new development on this branch though this particular change would be fine on both versions.

sideeffffect:
In that case, I think it would be better to target `master` and continue merging `master` to `zio-2` as we've been doing so far. What do you think?

adamgfraser:
Sure.

PR #859, opened 2022-01-26
fix Assertion.matches when using |

strokyl:
Before my fix the following test was failing:
```
    test("matches must fail when a regexp using | only match a part of the string") {
      assert((Assertion.matches("swimming|biking").apply("toto like biking")))(isLeft(anything))
    },
```
This was du to:
```
s"^$regexString$$".r
```
That would produce the regexp: `^swimming|biking$` which is equivalent to: `(^swimming)|(biking$)`.
Another solution could have been to use:
```
s"^$(?:regexString)$$".r
```

### DIFF
--- a/core-tests/shared/src/test/scala/zio/prelude/AssertionSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/prelude/AssertionSpec.scala
@@ -1,0 +1,19 @@
+package zio.prelude
+
+import zio.test.Assertion._
+import zio.test.{DefaultRunnableSpec, ZSpec, assert}
+
+object AssertionSpec extends DefaultRunnableSpec {
+
+  def spec: ZSpec[Environment, Failure] = suite("Assertion")(
+    test("matches must fail when the regex only match a part of the string") {
+      assert((Assertion.matches("biking").apply("toto like biking")))(isLeft(anything))
+    },
+    test("matches must fail when a regexp using | only match a part of the string") {
+      assert((Assertion.matches("swimming|biking").apply("toto like biking")))(isLeft(anything))
+    },
+    test("matches must work when the regex only match the full string") {
+      assert((Assertion.matches("t.*g").apply("toto like biking")))(isRight(isUnit))
+    }
+  )
+}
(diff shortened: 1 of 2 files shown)


adamgfraser (dismissed):
Good to merge when CI is passing!

core-tests/shared/src/test/scala/zio/prelude/AssertionSpec.scala
```diff
@@ -0,0 +1,20 @@
+package zio.prelude
+
+import zio.prelude.Assertion
+import zio.test.Assertion._
+import zio.test.{Assertion => _, _}
```
sideeffffect, line 5: I think you need to figure out a different way to do this, because Scala 2.11 doesn't like this :(
Perhaps importing things in `zio.test` explicitly?

sideeffffect (changes requested):
Great catch! :rocket: 
You have to make changes to make it compile on Scala 2.11.
Please let us know if you get stuck :wink:

strokyl:
@adamgfraser
While making it work for scala 2.11, I spot issues with `Regepxt.Anything` and `Regexp.Anychar`, so I added a second commit for them.
Maybe with `Anything`, I am wrong it should just have been named `Nothing` or maybe we want anything to not match string with a newline.
Also, I don't have time yet to look in details but I am pretty sure that `CharacterSet` will have an issue with some specific chars at least with `]`.
In general, from previous experiences, I know doing this kind of manual regexp construction is tricky, and some good testing is necessary to make sure the implementation is correct. If you think it's valuable, I can contribute more tests for that part when I found the time. But maybe, we should discuss that in a separate `Issues`.

sideeffffect:
Maybe `Regex.quote` would help you here?

strokyl:
Ok because of examples:
```
  object Email extends Subtype[String] {
    override def assertion = assert {
      matches {
        start ~
          anyRegexOf(alphanumeric, literal("-"), literal(".")).+ ~
          literal("@") ~
          anyRegexOf(alphanumeric, literal("-")).+ ~
          literal(".").+ ~
          anyRegexOf(alphanumeric, literal("-")).between(2, 4) ~
          end
      }
    }
  }
```
If I start fixing  Regexp.Anything and Regexp.Anychar, I will have to fix `CharactereSet`.
Going to just fix the initial spotted issue. ANd I will fix the rest in a separate PR.

adamgfraser (dismissed):
Thanks for working on this! 🙏

sideeffffect (approved):

strokyl:
@adamgfraser Look like I do not have the right to merge it myself

PR #887, opened 2022-03-10
Update Documentation Regarding ForEach

### LINKED ISSUE #884: ForEach.reduceIdentity / sum is somewhat confusing
the whole use of ForEach is non-obvious (a documentation issue) and when trying to use the ForEach sum function there's a name collision with the scala standard collections sum - resulting (in my combination of imports at least) in the wrong sum being used. 
Thus the .sum is more or less pointless. And it replicates the function of reduceIdentity

But reduceIdentity is absolutely not an obvious name. For a start, the programmer in question has to realise that there is an identity in implicit scope and then that it wants to be reduced - "reduce" probably being in the category of names considered "unergonomic" perhaps?

I'm not sure what the solution here is. And I appreciate that the Identity in question could be multiplication.

Maybe combineAll - like in cats would be easiest. Certainly it would save having to change code if converting.

adamgfraser:
Resolves #884.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ForEach.scala
+++ b/core/shared/src/main/scala/zio/prelude/ForEach.scala
@@ -408,7 +408,7 @@ trait ForEachSyntax {
    * Provides infix syntax for traversing collections.
    */
   implicit class ForEachOps[F[+_], A](private val self: F[A]) {
-    def concatenate(implicit F: ForEach[F], A: Identity[A]): A                                               =
+    def concatenate(implicit F: ForEach[F], A: Identity[A]): A                                                  =
       F.concatenate(self)
     def forEach[G[+_]: IdentityBoth: Covariant, B](f: A => G[B])(implicit F: ForEach[F]): G[F[B]]               =
       F.forEach(self)(f)
--- a/core/shared/src/main/scala/zio/prelude/ForEach.scala
+++ b/core/shared/src/main/scala/zio/prelude/ForEach.scala
@@ -304,12 +304,11 @@ trait ForEach[F[+_]] extends Covariant[F] { self =>
     reduceAssociative(fa)(Idempotent[A].idempotent)
 
   /**
-    * Reduces the collection to a summary value using the associative operation.
-    * Alias for `fold`.
-    */
-   def reduceIdentity[A: Identity](fa: F[A]): A =
-     foldMap(fa)(identity[A])
-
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/zio/prelude/ForEach.scala
```diff
@@ -38,6 +38,13 @@ trait ForEach[F[+_]] extends Covariant[F] { self =>
    */
   def forEach[G[+_]: IdentityBoth: Covariant, A, B](fa: F[A])(f: A => G[B]): G[F[B]]
 
+  /**
+   * Reduces the collection to a summary value using the associative operation.
+   * Alias for `fold`.
+   */
+  def concatenate[A: Identity](fa: F[A]): A =
```
sideeffffect, line 45: `concatenate` doesn't fit in with any of the other methods: `reduceAssociative`, `reduceMapOption`, ...
Shouldn't we just rename "reduce*" to "combineAll*"?

adamgfraser: I think `reduce` is generally a good name and is quite consistent with the Scala standard library which defines `fold` for a way of combining that takes a zero element and `reduce` for a way of combining that doesn't. This is also consistent with ZIO where we have operators like `reduceAll`, `reduceAllPar`, and `reduceAllParN` as well as a variety of fold variants.

I think the problem is more that this particular way of combining is so common, and actually does take a zero value, that we want a more straightforward name for it. Also `reduceIdentity` was a bit of a workaround since we originally called it `fold` versus `foldMap` since it doesn't require any mapping but that conflicts with `fold` defined on standard collections.

Maybe we could call it `reduceAll` or `combineAll` but seems a bit close to but also inconsistent with the names of all the other combining operators.

sideeffffect: It's a pickle :)
I like the `reduceIdentity` name. I also like `reduceAll`. If we kept any of these two and users would still have issues finding the method, why don't we introduce an alias? I think we could name the alias `concatenate`, but we would still keep the original method.
What do you think?

adamgfraser: I'm good with that. Updated now.

sideeffffect (approved):
:heart: 
(Sorry for late reply, I've been incredibly busy lately)

adamgfraser:
No worries! I think we all have!

PR #888, opened 2022-03-10
Update Documentation Regarding Newtype Constructors

### LINKED ISSUE #885: Newtype - minor change to docs
The one thing the docs don't actually say (though the author probably thinks it's implied) 
"Smart Newtypes can only be validated at compile-time when called with literals, such as 9000 or "Fancy Pants". When wrapping variables or run-time values, you can use the make or makeAll."

is that if there is an assertion in the type e.g.
```
  object MustHaveX extends Newtype[String] {
    override def assertion = assert {
      contains("X")
    }
  }
```
then an attempt to call the standard apply method will fail with a compiler message as below:
```
 Newtype Assertion Failed 
Could not validate Assertion at compile-time.
Either use a literal or call "NewTypeExperiments.this.MustHaveX.unsafeWrap(Expr[Nothing](NewTypeExperiments.this.sneaksNoX))"

  val unchecked = MustHaveX(sneaksNoX)
```

adamgfraser:
Resolves #885.

### DIFF
--- a/docs/newtypes/index.md
+++ b/docs/newtypes/index.md
@@ -289,7 +289,9 @@ val oops = SequenceNumber(-10)
 // • -10 did not satisfy greaterThanOrEqualTo(0)
 ```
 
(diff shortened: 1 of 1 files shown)


sideeffffect (approved):
Nice :+1:

PR #889, opened 2022-03-15
Adds the Optional type

vigoo:
This is the `Optional` type from `zio-k8s`. It is same as the standard `Option` type except that it has implicit conversion from the wrapped type to `Option.Present`. It is intended to be used in situations when an external data model full of optional fields is wrapped and wrapping everything explicitly in `Some` is too much boilerplate for the user.

The reason for moving here from `zio-k8s` is so it can be used in other similar projects of the ZIO ecosystem such as `zio-aws`.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/data/Optional.scala
+++ b/core/shared/src/main/scala/zio/prelude/data/Optional.scala
@@ -4,22 +4,23 @@ import zio.Chunk
 
 import scala.language.implicitConversions
 
-/** Custom option type to be used for wrapping external data models where most of the fields are
-  * defined to be optional.
-  *
-  * Instances of Optional are either [[Optional.Present]] or [[Optional.Absent]].
-  *
-  * The only difference between this type and [[scala.Option]] is that there is an implicit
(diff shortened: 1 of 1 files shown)


jdegoes (approved):

jdegoes:
Needs instances of all relevant type classes

adamgfraser (approved):

core/shared/src/main/scala/zio/prelude/data/Optional.scala
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * Custom option type to be used for wrapping external data models where most of the fields are
+ * defined to be optional.
+ *
+ * Instances of Optional are either [[Optional.Present]] or [[Optional.Absent]].
+ *
+ * The only difference between this type and [[scala.Option]] is that there is an implicit
+ * conversion defined from `A`` to `Optional[A]`, and from `Option[A]`` to `Optional[A]`.
+ */
+sealed trait Optional[+A] { self =>
```
sideeffffect, line 16: Shouldn't this perhaps extend `Iterable`?

Issue #919, opened 2022-05-24
Add Validation#toEitherAssociative and Validation#toZIOAssociative methods


SimY4:
There's a way to convert Validation into Either or ZIO by placing accumulated errors into NonEmptyChunk or Cause hierarchy. 

I propose adding an additional way of conversion through provided Associative type class that will provide shared custom rules for errors aggregation.

```
trait Validation[W, E, A] {
  def toEitherAssociative(implicit A: Associative[E]): Either[E, A] = ???
  def toZIOAssociative(implicit A: Associative[E]): IO[E, A] = ???
}
```

adamgfraser:
Sounds good to me.

PR #920, opened 2022-05-27
#919: ZValidation associative error aggregation.

SimY4:


core/shared/src/main/scala/zio/prelude/Associative.scala
```diff
@@ -311,6 +311,15 @@ object Associative extends AssociativeLowPriority {
       def inverse(l: => Sum[Char], r: => Sum[Char]): Sum[Char] = Sum((l - r).toChar)
     }
 
+  /**
+   * The `Identity` instance for `Cause`.
+   */
+  implicit def CauseCommutativeIdentity[A]: Commutative[Cause[A]] with Identity[Cause[A]] =
```
adamgfraser, line 317: This is somewhat arbitrary as `Cause` can be combined using both `both` and `then`. If we provide these instances we should probably do so with a new type like we do for integer addition and multiplication.

SimY4: @adamgfraser should it be Associative for `Cause[A]` but commutative for `AndF[Cause[A]]`?

sideeffffect: I believe it should be
```scala
Identity[Prod[zio.Cause[A]]]`
```
```scala
Commutative[Sum[zio.Cause[A]]] with Identity[Sum[zio.Cause[A]]]
```

https://github.com/zio/zio-prelude/pull/351/files#diff-f4fe370b6a2a72b31cf0b4c9b9f8294856f13514c7752f578d9dd0d7a3f1f407R1394-R1407

SimY4: @sideeffffect isn't `Sum` and `Prod` both commutative? Could get confusing really fast.

adamgfraser: I think we could use some specialized new types for this.

sideeffffect: @SimY4 It's been a long time since I played with `Identity`, `Commutative` and `Sum` and `Prod`. But generally speaking, not all "prod" operations are commutative, famous example is matrix multiplication.
And form my experiments from the PR linked above, I think that `zio.Cause` is another case.

@adamgfraser did you mean something like
```scala
  case class CauseSum[A](underlying: Sum[zio.Cause[A]]) {
    ...
  }
```
?

adamgfraser: No. Something more like:

```scala
object AndThen extends SubtypeF
type AndThen[A] = AndThen.Type[A]

object Both extends SubtypeF
type Both[A] = Both.Type[A]
```

Basically I think you are right about `Sum` and `Prod` potentially being used here and that could be useful in some very abstract situations but I think in most situations you want more domain specific names for these.

sideeffffect: Ah, now I see what you meant :+1:

core/shared/src/main/scala/zio/prelude/Associative.scala
```diff
@@ -311,6 +311,32 @@ object Associative extends AssociativeLowPriority {
...
+        val rUnwrapped: Cause[A] = r
+        Both(lUnwrapped && rUnwrapped)
+      }
+      val identity: Both[Cause[A]]                                            = Both(Cause.empty)
+    }
+
+  /**
+   * The `Associative` and `Identity` instance for `Cause`.
+   */
+  implicit def CauseAssociativeIdentity[A]: Associative[AndThen[Cause[A]]] with Identity[AndThen[Cause[A]]] =
```
sideeffffect, line 330: The `Associative[AndThen[Cause[A]]]` is superfluous, this can be just `Identity[AndThen[Cause[A]]]`.

adamgfraser: Fixed in #923.

PR #921, opened 2022-05-27
modified my non-empty contributions to remove (hopefully) scala 2.13 …

TimPigden:
…dependency #836

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/NonEmptyMap.scala
+++ b/core/shared/src/main/scala/zio/prelude/NonEmptyMap.scala
@@ -24,7 +24,7 @@ import scala.language.implicitConversions
  * A non-empty wrapper for the scala immutable map. Note - this does not attempt to implement all features of
  * map but what the author considers to be the "normal ones".
  */
-case class NonEmptyMap[K, V] private (private val map: Map[K, V]) { self =>
+final class NonEmptyMap[K, V] private (private val map: Map[K, V]) { self =>
 
   private def newMap[V2](map: Map[K, V2]): NonEmptyMap[K, V2] = new NonEmptyMap(map)
 
@@ -175,7 +175,7 @@ object NonEmptyMap {
    * the elements will be non-empty - returns None if from is
    */
   def groupByOption[A, K](from: Iterable[A])(f: A => K): Option[NonEmptyMap[K, Iterable[A]]] =
-    from.headOption.map(_ => NonEmptyMap(from.groupBy(f)))
+    from.headOption.map(_ => new NonEmptyMap(from.groupBy(f)))
 
   /**
    * from a non-empty chunk we can create a non-empty map of non-empty chunks
@@ -183,7 +183,7 @@ object NonEmptyMap {
   def groupByFromNonEmptyChunk[A, K](from: NonEmptyChunk[A])(f: A => K): NonEmptyMap[K, NonEmptyChunk[A]] = {
     val gb       = from.groupBy(f)
     val asChunks = gb.map(p => (p._1 -> NonEmptyChunk.fromIterableOption(p._2).get)) // safe!
-    NonEmptyMap(asChunks)
+    new NonEmptyMap(asChunks)
   }
 
   /**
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/zio/prelude/NonEmptySortedSet.scala
```diff
@@ -0,0 +1,264 @@
...
+    NonEmptySortedSet.fromSet(head, tail.union(r))
+  }
+
+  /** Creates a `NonEmptySortedSet` containing elements from `l` and `r` */
+  def union[A](l: SortedSet[A], r: NonEmptySortedSet[A])(implicit ordering: SOrdering[A]): NonEmptySortedSet[A] =
+    union(r, l)
+
+  /**
+   * The `CommutativeEither` instance for `NonEmptySortedSet`.
+   * todo
```
sideeffffect, line 187: commented code

core/shared/src/main/scala/zio/prelude/NonEmptySortedSet.scala
```diff
@@ -0,0 +1,264 @@
...
+
+  /**
+   * Derives a `Hash[NonEmptySortedSet[A]]` and `PartialOrd[NonEmptySortedSet[A]]` (and thus `Equal[NonEmptyList[A]]`) instance.
+   */
+  implicit def NonEmptySortedSetHashPartialOrd[A]: Hash[NonEmptySortedSet[A]] with PartialOrd[NonEmptySortedSet[A]] =
+    HashPartialOrd.derive[Set[A]].contramap(_.toSet)
+
+  /**
+   * The `Invariant` instance for `NonEmptySortedSet`.
+   * todo - don't understand
```
sideeffffect, line 230: commented code

sideeffffect (commented):
A few places with commented code, otherwise looks good to me :+1:

core/shared/src/main/scala/zio/prelude/NonEmptyMap.scala
```diff
@@ -0,0 +1,224 @@
...
+
+import zio.NonEmptyChunk
+
+import scala.language.implicitConversions
+
+/**
+ * A non-empty wrapper for the scala immutable map. Note - this does not attempt to implement all features of
+ * map but what the author considers to be the "normal ones".
+ */
+case class NonEmptyMap[K, V] private (private val map: Map[K, V]) { self =>
```
regiskuckaertz, line 27:
```suggestion
case class NonEmptyMap[K, V] private (private val map: Map[K, V]) extends AnyVal { self =>
```
any reason not to?

TimPigden: sorry I don't understand what you want there or why

sideeffffect: @regiskuckaertz is suggesting a change to make this into a value class via `extends AnyVal`
image

sideeffffect: That being said, I don't think we want to do that. For example because of `equals`, we want that to be true only when both sides are `NonEmptyMap`

TimPigden: but why?

sideeffffect, line 27: I think this shouldn't be a `case class`, just a mere `class`, as are the other `NonEmpty*` classes

TimPigden: ok, these are subtle differences the impact of which I don't really undersand. But yes, conformity to the pattern is a definite plus

TimPigden:
Help - can someone tell me what the problem is now? It's talking about "expected fix". But I've no idea what's expected.

sideeffffect:
Run `sbt fix` and everything will be alright :wink:

sideeffffect (commented):
Looks good now :+1:

adamgfraser (approved):

PR #946, opened 2022-07-18
Implement Validation.validateAll For Sets

### LINKED ISSUE #945: Validation.validateAll() for Sets?
I don't know if that's easily possible, but I'd find it very handy to be able to use `Validation.validateAll()` on Sets, not just on Lists.

I find myself often having a `Set[Validation[E, A]]` which I want to turn into `Validation[E, Set[A]]`. Currently I have to do `Validation.validateAll(set.toList).map(_.toSet)` which works, but it feels unnecessary cumbersome.

adamgfraser:
Resolves #945.


sideeffffect (approved):
Using validation on Sets could be tricky, but I can see how it could be useful from time to time...

PR #949, opened 2022-07-27
I heard you like newtypes (fixes cyclic alias error)

### LINKED ISSUE #817: Subtype of Newtype cause cyclic subtyping error
Hi,

Until RC8 we could do subtypes of newtypes 
```scala 
 object GenericItemId extends Newtype[UUID]
 type GenericItemId = GenericItemId.Type

 object SpecificItemId extends Subtype[GenericItemId]
 type SpecificItemId = SpecificItemId.Type
```

This is a pattern we use a lot, sometimes with multiple nesting levels.

Since RC8, this fails with 

```
cyclic aliasing or subtyping involving type Type
type GenericItemId = GenericItemId.Type
```

kitlangton:
Fixes https://github.com/zio/zio-prelude/issues/817 (once CI passes! 🤞 ).

By they by, this is only fixed in 2.12 and up. We can't fix it in 2.11 without breaking the existing ability to have implicit resolution work for subtypes.


examples/shared/src/main/scala-2/examples/SmartTypes.scala
```diff
@@ -64,5 +64,6 @@ object SmartTypes extends App {
   }
   type Email = Email.Type
   val email: Email = Email("test@test.com")
+
```
sideeffffect, line 67: I don't mind adding this new line, but I just want to point it out :wink:

sideeffffect (changes requested):
There are dots in the names of files (instead of nesting directories): `zio.prelude` instead of `zio/prelude`.
Otherwise looks fine to me, thanks a lot! :pray:

examples/shared/src/main/scala-2/examples/SmartTypes.scala
kitlangton: lol. I didn't even notice. That's my OCD 😛, I find myself adding spacing before after the start/end braces.

kitlangton: Weird about the dots. I blame IntelliJ! Will fix ;)

kitlangton: CleanShot 2022-07-27 at 21 18 24@2x
CleanShot 2022-07-27 at 21 18 30@2x

ha, yeah. they look the same because of how IntelliJ renders this (though the correct one is yellow).

adamgfraser (approved):

mbaechler (approved):
It looks good, thanks for this!

sideeffffect (approved):
So shall we merge? :smile_cat:

neko-kai:
@kitlangton 
NOTE: This only works on 2.12 if `-Xsource:2.13`/`-Xsource:3` is used (which is used in this repo). This concerns client code as well, so for all 2.12 users without `-Xsource:2.13` implicit resolution won't work.

alphaho:
Hi @kitlangton, @adamgfraser , can we make a release with this change? I really want to use this feature in the work project.

adamgfraser:
@alphaho I am somewhat concerned about the comment from @neko-kai above. We also need to merge these changes into the ZIO 2 branch.

guizmaii:
Doesn't seem to work for Scala 2.13 https://github.com/zio/zio-prelude/issues/1183

:/

ghost:
> Doesn't seem to work for Scala 2.13 #1183

@guizmaii , your issue #1183 is about `Subtype of Subtype`. This does not work, even with 2.12.* (this PR fixes `Subtype of Newtype`)

You need to correct your issue, see https://github.com/zio/zio-prelude/issues/1183#issuecomment-1825819731

Issue #963, opened 2022-10-04
Flatmap `>>=` operator


carlos-verdes:
I could not find support for ` >>=` , it would be great for composition.
Something like:
```scala

val transformation: B => F[B] = ???
monad1.flatmap(transformation)

monad1 >>= transformation

sideeffffect:
Let me link to comment by @adamgfraser 
https://github.com/zio/zio-prelude/pull/965#issuecomment-1267020980

Long story short, symbolic methods have a higher cognitive cost that probably isn't worth it here.

I will close the ticket now. As always, we may reconsider adding this in the future.
But thanks for your interest in ZIO Prelude, we hope that you will enjoy it even without `>>=`.

carlos-verdes:
I started long time back with Scalaz, then found Cats, created my first framework using Free Monads... understood it was quite complex for new people to adopt and understand... then I moved to tagless final and found couple of videos on ZIO.

Right now I'm moving into full ZIO and obviously I missed all the cats functional type classes etc. so I'm giving an opportunity to zio-prelude (for me it has sense to use full set of libraries from the same community).
Honestly speaking what is important is the principles applied on the code not the small details like this (I can always use `.flatMap` or define myself my own extension method for `>>=`.

In any case thanks for the effort put creating this library and the prompt response (one of my dreams thought would be to have all Scala functional communities united, it will reduce a lot of friction for new people to adopt it).

PR #965, opened 2022-10-04
Symbolic alias for flatMap

sideeffffect:
closes https://github.com/zio/zio-prelude/issues/963
/cc @carlos-verdes

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
+++ b/core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
@@ -249,5 +249,11 @@ trait AssociativeFlattenSyntax {
      */
     def flatMap[B](f: A => F[B])(implicit flatten: AssociativeFlatten[F], covariant: Covariant[F]): F[B] =
       flatten.flatten(covariant.map(f)(fa))
+
+    /**
+     * A symbolic alias for `flatMap`.
+     */
+    def >==[B](f: A => F[B])(implicit flatten: AssociativeFlatten[F], covariant: Covariant[F]): F[B] =
+      flatMap(f)
   }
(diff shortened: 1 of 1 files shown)


adamgfraser:
Previous we deleted `>>=` from ZIO. Symbolic operators create issues for discovery and accessibility, particularly when they are not extremely common and when they are not binary operators that take the form `(A, A) => A`, both of which are false here. The savings in characters is also extremely small, `model.flatMap(transformation)` versus `model >>= transformation`.

There is some argument that since ZIO Prelude is the more of our take on functional abstractions perhaps we should be more open to this here than in ZIO itself but I think there is significant value in consistency and not in introducing additional symbolic operators so I think I am thumbs down on this.

carlos-verdes:
Thanks for your answer, at least I know I was right when I didn't found it :)

In my humble opinion it's a really common operator, I find it in many articles talking about Monads, Functors etc (lot of them have examples fro Haskel where this is a defined operator)and code looks quite clean.

There is an article I like to share to "new functional code joiners" where you can see the graphical representation of `>>=` as plunger (I found it even funny), I understand (and respect) this is your choice but I still don't understand what is the harm on having it.

The article just in case you are curios is this one:
https://adit.io/posts/2013-04-17-functors,_applicatives,_and_monads_in_pictures.html

image

sideeffffect:
I could live with it. So if you convince @adamgfraser , I'm good with it too :smile_cat:

carlos-verdes:
No need, I can live without it also :)
It's just it took sometime to realize it was not supported, but it's normal when you change between libraries to have this kind of adaptation period (similar happen with Functor and Covariant, in this case though you added an alias)

PR #971, opened 2022-10-12
Specify validateAll instead of collectAllPar

einnocent:
The method name was changed in https://github.com/zio/zio-prelude/pull/666


sideeffffect (approved):
Thank you very much :+1:

PR #972, opened 2022-10-14
Fix error messages for Assertion.endsWith

MichalPawlicki:


sideeffffect (approved):
Nice catch! Thank you :+1:

PR #990, opened 2022-10-28
Fix equalTo macro for dotty

### LINKED ISSUE #922: equalTo not working with the compile-time smart types
> Scala 3.1.2
> ZIO 2.0.0-RC6
```scala
object Five extends Newtype[Int]:
  override inline def assertion = equalTo(5)
Five(5)
```
results in
```
-- Error: ----------------------------------------------------------------------
1 |Five(5)
  |^^^^^^^
  | Assertion Failed 
  |
  |We were unable to read your assertion at compile-time.
  |
  |You must either define your assertion directly or refer to other inline definitions:
  |
  |    override inline def assertion = greaterThan(10) && lessThan(100).
  |
  |    or
  |
  |    inline def extracted = greaterThan(10) && lessThan(100).
  |    override inline def assertion = extracted.
  |            
  | This location contains code that was inlined from rs$line$20:1
1 error found
```
while
```scala
object Five extends Newtype[Int]:
  override inline def assertion = between(5,5)
Five(5)
```
compiles correctly.

mschuwalow:
resolves #922

The example it the issues seems to already work correctly with an explicit type annotation:
(My guess it has something to do with scala using the literal type of 5 for the expressions -> our macro match fails, but I know too little about this to be sure)
```
object Five extends Newtype[Int]:
  override inline def assertion = equalTo[Int](5)
Five(5)

```

Looking at the macro docs https://docs.scala-lang.org/scala3/guides/macros/quotes.html they seem to only recommend explicitly specifying the type for methods and infix operations.
If we drop them from the other expressions the example from the issue starts to work and it doesn't seem like we are introducing regressions


kitlangton (approved):
Woo! LGTM

PR #1001, opened 2022-11-19
Publish Docs to The NPM Registry

khajavi:


khajavi:
This is ready for review.

build.sbt
```diff
@@ -3,7 +3,7 @@ import BuildHelper._
 inThisBuild(
   List(
     organization := "dev.zio",
-    homepage     := Some(url("https://zio.github.io/zio-prelude/")),
+    homepage     := Some(url("https://zio.dev/zio-prelude/")),
```
sideeffffect, line 6: The page doesn't exist :(
Or will it appear once we merge this PR?

build.sbt
```diff
@@ -233,26 +232,15 @@ lazy val benchmarks = project
...
-    ScalaUnidoc / unidoc / target              := (LocalRootProject / baseDirectory).value / "website" / "static" / "api",
-    cleanFiles += (ScalaUnidoc / unidoc / target).value,
-    docusaurusCreateSite                       := docusaurusCreateSite.dependsOn(Compile / unidoc).value,
-    docusaurusPublishGhpages                   := docusaurusPublishGhpages.dependsOn(Compile / unidoc).value
+    scalacOptions -= "-Xfatal-warnings"
   )
   .settings(macroDefinitionSettings)
   .dependsOn(core.jvm, experimental.jvm, experimentalLaws.jvm, laws.jvm, scalaParallelCollections)
-  .enablePlugins(MdocPlugin, DocusaurusPlugin, ScalaUnidocPlugin)
+  .enablePlugins(WebsitePlugin)
```
sideeffffect, line 243: Does WebsitePlugin do the unidoc stuff?

project/plugins.sbt
```diff
@@ -15,3 +15,6 @@ addSbtPlugin("org.scalameta"                     % "sbt-mdoc"
 addSbtPlugin("org.scalameta"                     % "sbt-scalafmt"                  % "2.5.0")
 addSbtPlugin("pl.project13.scala"                % "sbt-jcstress"                  % "0.2.0")
 addSbtPlugin("pl.project13.scala"                % "sbt-jmh"                       % "0.4.3")
+addSbtPlugin("dev.zio"                           % "zio-sbt-website"               % "0.0.0+84-6fd7d64e-SNAPSHOT")
+
+resolvers += Resolver.sonatypeRepo("public")
```
sideeffffect, line 20: Isn't this redundant? I thought that it's already present by default.

.github/workflows/site.yml
```diff
@@ -1,35 +1,28 @@
...
-          distribution: temurin
-          java-version: 8 # publish for 8
-          check-latest: true
-      - name: Install libuv
-        run: sudo apt-get update && sudo apt-get install -y libuv1-dev
-      - name: Set Swap Space
-        uses: pierotofy/set-swap-space@master
+          fetch-depth: 0
+      - name: Setup Scala and Java
+        uses: olafurpg/setup-scala@v13
```
sideeffffect, line 20: I know that this is generated code and it's not a big deal anyway (yet!).

But olafurpg/setup-scala is essentially dead now.
https://github.com/olafurpg/setup-scala/issues/49
I think we should just use `actions/setup-java`. Maybe we should change the code in the generating plugin.

sideeffffect (commented):
Huge thanks for spending your time on this. :bow: 
May I just ask you for a clarification on a few things?

build.sbt
khajavi: There is another PR that will generate the homepage on zio.dev site: https://github.com/zio/zio/pull/7545

build.sbt
khajavi: Not right now, but I will add that to the subsequent iterations. Currently, the goal is to integrate all docs into one place. Then I will focus on AP‌I docs.

project/plugins.sbt
khajavi: I remember, it cause not found issue when downloading snapshot artifacts. I can check it again.

.github/workflows/site.yml
khajavi: Thank you for the feedback, I will change that on the zio-sbt plugin.

project/plugins.sbt
sideeffffect: For snapshots you need
```scala
resolvers ++= Resolver.sonatypeOssRepos("snapshots")
```
(don't use `sonatypeRepo` anymore, always just `sonatypeOssRepos`)

build.sbt
sideeffffect: Fair enouch

build.sbt
sideeffffect: Good

project/plugins.sbt
khajavi: Thank you! I removed that line and it works properly for th `zio-prelude` project. I need to investigate more to find out why I need to explicitly set the resolver for snapshot versions in some repos.

sideeffffect: You need that :point_up: because you're using snapshot of `zio-sbt-website`

khajavi: In my local machine that works properly, maybe because of cached artifacts. Anyway, I added that line.

sideeffffect (approved):

.github/workflows/ci.yml
```diff
@@ -110,7 +110,7 @@ jobs:
         with:
           swap-size-gb: 7
       - name: Check Website Generation
-        run: ./sbt docs/docusaurusCreateSite
+        run: ./sbt docs/compileDocs
```
sideeffffect, line 113: btw does this also test _building_ the actual website?

khajavi: No, not the whole! Right now it only checks the compilation of documents using mdoc. The next level is to check the validation of sidebar.js that is not included in this version of zio-sbt-website plugin.

PR #1009, opened 2022-12-03
Clean Up Docs

khajavi:
- Use the categorized sidebar
- Use SEO-friendly URLs
- Update the home page


adamgfraser (approved):

adamgfraser (approved):

sideeffffect:
Oh, so now we have the documentation of all the ZIO libs on one site. Nice! :+1: 

But I've noticed that the API reference link, even on ZIO Prelude, points to ZIO itself on javadoc.io, instead of to

https://zio.github.io/zio-prelude/api/zio/prelude/

image

Is there a plan to remedy that?

adamgfraser:
I think I would expect the bottom navigation bar to go to ZIO itself since all the content there is "global" to ZIO (the stars for the ZIO project, etc...) but it would be good to link to the Scaladoc for the individual projects somewhere.

Copying @khajavi.

sideeffffect:
The best thing would be to have just one single API reference for all official ZIO libraries (including ZIO proper, ZIO Prelude, and all the others). Of course, that would have to be hosted by us, javadoc.io won't help us here.

khajavi:
@sideeffffect Yes, I agree that having a single API reference for all official libraries is necessary. However, I have not yet started on that. If you have any suggestions, please share them with me.

PR #1012, opened 2022-12-07
Generate README.md File Automatically

khajavi:


README.md
```diff
@@ -1,48 +0,0 @@
-# ZIO Prelude
```
sideeffffect, line 1: README.md should still be part of the repository, right?

sideeffffect: Oh, so GitHub Action will change code in `master`, correct?

khajavi: Yes

adamgfraser (approved):

sideeffffect:
This is awesome work you're doing @khajavi :superhero: 

It seems like we're not out of the woods yet, though.
https://github.com/zio/zio-prelude/actions/runs/3650675180/jobs/6166952422

Perhaps the protected branch settings need to be tweaked? But those play an important role, right? So I hope there is another way.
Thanks! :pray:

PR #1048, opened 2023-01-10
Update Scala & ZIO versions 

guizmaii:


adamgfraser (dismissed):

.github/workflows/ci.yml
```diff
@@ -7,7 +7,7 @@ env:
 on:
   pull_request:
   push:
-    branches: ['master', 'zio-2']
+    branches: ['master', 'series/2.x']
```
adamgfraser: I think can just do `series/2x` at this point.

sideeffffect: Theoretically we could want to publish something in the `series/1.x` branch, right?

sideeffffect:
```suggestion
    branches: ['series/1.x', 'series/2.x']
```

adamgfraser: I don't think we should be publishing snapshots for it.

sideeffffect: I don't mind if we don't do that :shrug: :+1:

core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
```diff
@@ -72,17 +72,17 @@ object AssociativeFlatten {
   /**
    * The `AssociativeFlatten` instance for `Const`.
    */
-  implicit def ConstAssociativeFlatten[A]: AssociativeFlatten[({ type ConstA[+B] = Const[A, B] })#ConstA] =
-    new AssociativeFlatten[({ type ConstA[+B] = Const[A, B] })#ConstA] {
+  implicit def ConstAssociativeFlatten[A]: AssociativeFlatten[Const[A, +*]] =
+    new AssociativeFlatten[Const[A, +*]] {
```
sideeffffect: We may want to choose to switch to kind-projector syntax, but I'd prefer we do it in a dedicated PR (the manual syntax is used _pervasively_).
There also may be some issues related to this new syntax -- there were reasons to not use it in the beginning after all...

adamgfraser:
@guizmaii Can we just keep this PR focused on updating dependencies?

guizmaii:
> Can we just keep this PR focused on updating dependencies?

Yup, sorry about that. Rollbacked

adamgfraser (approved):

.github/workflows/ci.yml
```diff
@@ -7,7 +7,7 @@ env:
 on:
   pull_request:
   push:
-    branches: ['master', 'zio-2']
+    branches: ['series/1.x', 'series/2.x']
```
adamgfraser: Weren't we going to change this to just `series/2.x`?

adamgfraser (changes requested):
Just one change on the snapshot publishing.

.github/workflows/ci.yml
guizmaii: @adamgfraser I just applied this @sideeffffect proposal: https://github.com/zio/zio-prelude/pull/1048#pullrequestreview-1242270136

adamgfraser: @guizmaii I think based on that conversation we said we weren't going to include `series/1.x`.

guizmaii: @adamgfraser Done :)

sideeffffect (approved):
Thanks for all the love :heart:

adamgfraser (approved):

PR #1054, opened 2023-01-14
Provide classtags for newtypes

### LINKED ISSUE #1003: `ClassTag` for `NewType` stopped working in RC16
Not sure if this is a regression or I'm missing something:

```
$ cat newtype.sc 

import zio.prelude._

object Foo extends Newtype[String]
type Foo = Foo.Type

implicitly[scala.reflect.ClassTag[Foo]]

$ scala-cli compile -S 2.13.10 --dep dev.zio::zio-prelude:1.0.0-RC15 newtype.sc

$ scala-cli compile -S 2.13.10 --dep dev.zio::zio-prelude:1.0.0-RC16 newtype.sc
Compiling project (Scala 2.13.10, JVM)
[error] ./newtype.sc:7:1: No ClassTag available for newtype.Foo
[error] implicitly[scala.reflect.ClassTag[Foo]]
[error] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error compiling project (Scala 2.13.10, JVM)
Compilation failed
```

mschuwalow:
resolves #1003

### DIFF
--- a/core/shared/src/main/scala-2/zio/prelude/Newtype.scala
+++ b/core/shared/src/main/scala-2/zio/prelude/Newtype.scala
@@ -1,6 +1,7 @@
 package zio.prelude
 
 import zio.NonEmptyChunk
+
 import scala.reflect.ClassTag
 
 /**
@@ -322,9 +323,8 @@ object Newtype {
       }
   }
 
-  implicit def classTag[N <: Newtype[_]: ClassTagWrapper]: ClassTag[N#Type] = {
+  implicit def classTag[N <: Newtype[_]: ClassTagWrapper]: ClassTag[N#Type] =
     new ClassTag[N#Type] {
       def runtimeClass: Class[_] = ClassTagWrapper[N].classTag.runtimeClass
     }
-  }
 }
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala-2/zio/prelude/Newtype.scala
```diff
@@ -306,4 +308,23 @@ object Newtype {
...
+  private object ClassTagWrapper {
+    def apply[A](implicit ev: ClassTagWrapper[A]): ClassTagWrapper[A] = ev
+
+    implicit def classTagWrapperForNewtype[A](implicit underlying: ClassTag[A]): ClassTagWrapper[Newtype[A]] =
+      new ClassTagWrapper[Newtype[A]] {
+        def classTag: ClassTag[_] = underlying
+      }
+  }
+
+  implicit def classTag[N <: Newtype[_]: ClassTagWrapper]: ClassTag[N#Type] = {
```
mschuwalow: I'm not quite sure why we need the implicit def both here and in the trait, but 2.12 and 2.13 fail unless both are present.
In practice they will only evaluate this one, but it seems like the other one influences implicit search in some way.

adamgfraser (approved):

PR #1055, opened 2023-01-14
[scala3] Fix equalTo and notEqualTo Assertions

### LINKED ISSUE #1051: Scala3: The newtypes assertion doesn't seem to work
With a very simple type like:

```scala
  object NonEmptyString extends Subtype[String] {
    override inline def assertion = notEqualTo("")
  }
  type NonEmptyString = NonEmptyString.Type
```

I have the following error:
```scala
[info] compiling 1 Scala source to /Users/jules/workspace/scala3-zio-prelude-errors-reproducer/modules/program/target/scala-3.2.1/classes ...
[error] -- Error: /Users/jules/workspace/scala3-zio-prelude-errors-reproducer/modules/program/src/main/scala/com/guizmaii/scala/three/zio/prelude/errors/reproducer/app/Main.scala:7:35 
[error]  7 |  private val toto = NonEmptyString("toto")
[error]    |                     ^^^^^^^^^^^^^^^^^^^^^^
[error]    | Assertion Failed 
[error]    |
[error]    |We were unable to read your assertion at compile-time.
[error]    |
[error]    |You must either define your assertion directly or refer to other inline definitions:
[error]    |
[error]    |    override inline def assertion = greaterThan(10) && lessThan(100).
[error]    |
[error]    |    or
[error]    |
[error]    |    inline def extracted = greaterThan(10) && lessThan(100).
[error]    |    override inline def assertion = extracted.
[error]    |            
[error]    |----------------------------------------------------------------------------
[error]    |Inline stack trace
[error]    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]    |This location contains code that was inlined from Main.scala:7
[error]     ----------------------------------------------------------------------------
[error] one error found
```

FYI, I made a minimal reproducer in this repo: https://github.com/guizmaii/scala3-zio-prelude-errors-reproducer/pull/1

As you can see in the CI: https://github.com/guizmaii/scala3-zio-prelude-errors-reproducer/pull/1/checks, the test is passing for Scala 2.13 but is failing for 3.2

mschuwalow:
resolves #1051 

What seems to be happening is that during macro evaluation time the overall Assertion is typed as `Assertion[Any]` if no explicit type parameter is present, so our previous pattern fails.

This seems to only happen if the function has no typeclass constraint that constrains the type (equalTo, notEqualTo). Functions with typeclass constraints (between, greaterThan, etc.) work fine with both explicit type parameter and without.


adamgfraser (approved):

guizmaii:
Thank you so much @mschuwalow ❤️

PR #1066, opened 2023-01-31
Ability to define custom `Assertion`s via test-functions

### LINKED ISSUE #1065: Ability to define custom `Assertion`s
Currently the list of assertions is quite limited and since `Assertion` trait is sealed users cant define new assertion. For example, if user wants to define assertion to test that a string is blank (does not contain only space-chars), the only way to do that is to use regular expressions. Although there is a simpler way: just reuse `String.isBlank` function to do that, something like:

```scala
val isBlank: Assertion[String] =
  Assertion.fromFunction("isBlank")(_.isBlank)
```

If we want to ensure that a list contains some element, we could write something like:

```scala
def containsElem[A](elem: A): Assertion[List[A]] =
  Assertion.fromFunction(s"containsElem($elem)")(_.contains(elem))
```

For the record, `Assertion` from `zio-test` has similar ability to define custom assertions from user-supplied functions.

nartamonov:
Fixes #1065

### DIFF
--- a/core-tests/shared/src/test/scala/zio/prelude/AssertionSpec.scala
+++ b/core-tests/shared/src/test/scala/zio/prelude/AssertionSpec.scala
@@ -15,12 +15,10 @@ object AssertionSpec extends ZIOSpecDefault {
     test("matches must work when the regex only match the full string") {
       assert((Assertion.matches("t.*g").apply("toto like biking")))(isRight(isUnit))
     },
-
     suite("fromFunction")(
       test("must succeed if function returns true") {
         assert(isBlank.apply(""))(isRight(isUnit))
       },
-
       test("must fail if function returns false") {
         assert(isBlank.apply("non-blank"))(isLeft(anything))
       }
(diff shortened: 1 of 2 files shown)


sideeffffect (changes requested):
Thank you, looks really nice. But the code doesn't pass our build.
Run `sbt fix` to format your code and Scaladoc has issues with your new doc comment:
```
[error] /home/runner/work/zio-prelude/zio-prelude/macros/shared/src/main/scala/zio/prelude/Assertion.scala:57:49: Variable elem undefined in comment for method fromFunction in object Assertion
[error]    *     Assertion.fromFunction(s"containsElem($elem)")(_.contains(elem))
```

nartamonov:
@sideeffffect Thank you, fixed!

nartamonov:
Oopps. It seems it's not so easy as I thought because of macro limitations:

```
[error]  Newtype Assertion Failed
[error] You have defined your Assertion in a way that cannot be read at compile-time.
[error] Due to the limitations of macros, assertions cannot be abstracted into other definitions.
[error]
[error] Make certain your definition looks something like this:
[error]
[error]       import zio.prelude.Assertion._
[error]       override def assertion = assert(greaterThan(40) && lessThan(80))
[error]
[error]     NonBlankString("abc")
```

I'll try to find another way...

adamgfraser:
Isn't this already done in #793?

nartamonov:
@adamgfraser Oh, thank you, `assertCustom` (as implemented in #793) is exactly what I need, didn't discover it in codebase somehow :-) Happy to close this PR.

PR #1075, opened 2023-02-10
syntax improve

rituraj2342:


scala-parallel-collections/src/main/scala/zio/prelude/scalaparallelcollections/Syntax.scala
```diff
@@ -13,7 +13,7 @@ trait Syntax {
     )(r: ParMap[K, V]): PartialOrdering = {
       def commonValues(lesserMap: ParMap[K, V]): ParIterable[(V, V)] =
         // `toIterable` so that we don't incidentally create a map (and thus drop duplicate would-be keys)
-        lesserMap.toIterable.map { case (k, _) => (l(k), r(k)) }
```
adamgfraser: You just deleted this without addressing the comment above explaining why it was added.

scala-parallel-collections/src/main/scala/zio/prelude/scalaparallelcollections/AssociativeFlattenInstances.scala
```diff
@@ -1,15 +1,16 @@
...
 
+import scala.collection.parallel.immutable.ParMap
 import scala.collection.parallel.{immutable => par}
 
 trait AssociativeFlattenInstances {
 
   /**
    * The `AssociativeFlatten` instance for `ParMap`
    */
-  implicit def ParMapAssociativeFlatten[K]: AssociativeFlatten[({ type lambda[+v] = par.ParMap[K, v] })#lambda] =
```
adamgfraser, line 11: This is just renaming a variable.

adamgfraser (changes requested):
Thanks for your PR. However, this appears to just be renaming a variable and removing an operator without addressing the comment immediately above explaining why the operator was even added or even deleting the comment. We need to have more of a justification than this to make a change in the code.

scala-parallel-collections/src/main/scala/zio/prelude/scalaparallelcollections/Syntax.scala
rituraj2342: Oh yes..My bad. Reverting this change. Thanks

sideeffffect:
Dotty needs `-Ykind-projector`. But according to documentation, the `*` syntax is removed in 3.3.x and above.
https://docs.scala-lang.org/scala3/reference/changed-features/wildcards.html
So it seems to me that if we want to keep cross compiling on 2 and >= 3.3, we are stuck with the manual type lambda encoding :crying_cat_face:

adamgfraser:
Going to close this but thanks for your work!

PR #1079, opened 2023-02-14
Explore Derivation Of Functional Abstractions

adamgfraser:


core-tests/shared/src/test/scala-3/zio/prelude/DerivationSpec.scala
```diff
@@ -0,0 +1,31 @@
+package zio.prelude
+
+import zio._
+import zio.test._
+
+object DerivationSpec extends ZIOSpecDefault {
+
+  final case class Person(name: String, age: Int)
+
+  sealed trait Color
```
sideeffffect, line 10: How come you don't need the `derives Equal` clauses? I though that you need to use them :thinking: 

https://docs.scala-lang.org/scala3/reference/contextual/derivation.html

sideeffffect (approved):
Good initiative :+1:

core-tests/shared/src/test/scala-3/zio/prelude/DerivationSpec.scala
adamgfraser: Apparently not.

core/shared/src/main/scala-3/zio/prelude/EqualVersionSpecific.scala
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package zio.prelude
+
+import scala.deriving.*
+import scala.compiletime.{erasedValue, summonInline}
+
+trait EqualVersionSpecific {
+
+  inline given derived[A](using mirror: Mirror.Of[A]): Equal[A] =
```
regiskuckaertz, line 24:
```suggestion
  inline def derived[A](using mirror: Mirror.Of[A]): Equal[A] =
```
Otherwise it will derive for all types with no control right?

adamgfraser: `given` is in the documentation of this feature and using `def` creates an infinite loop in the derivation.

jdegoes (approved):

core/shared/src/main/scala-3/zio/prelude/EqualVersionSpecific.scala
sideeffffect: That's what I've been alluding to in my comment above: https://github.com/zio/zio-prelude/pull/1079#discussion_r1106119518

I also think that people would like to control it with `derives Equal` clauses.

adamgfraser: I opened an issue with the Scala 3 team regarding this.

regiskuckaertz: > `def` creates an infinite loop in the derivation.

Erf that is ... weird. This code is equivalent to fully automatic derivation. With a `def` users  would need to explicitly add the `derives` clause. I had many complaints with scanamo when instances were created implicitly.

 
I read the docs, the "recursively auto-derive" is an anti pattern imo, oh well I guess time will tell.

sideeffffect: So also according to @bishabosha we shouldn't put the `given` on the derived method to let the derivations be driven by `derives` clauses.

https://github.com/lampepfl/dotty/issues/16916

> using `def` creates an infinite loop in the derivation

Are you able replicate it? Maybe we should report that.

I'm trying my lack with changing the official documentation https://github.com/lampepfl/dotty/pull/16919 Let's see where this leads :smile:

bishabosha: > `given` is in the documentation of this feature and using `def` creates an infinite loop in the derivation.

is the infinite loop with any type, or a specific example?

adamgfraser: The derivation for `Color`.

PR #1080, opened 2023-02-14
Update Derivation Implementation

adamgfraser:
Apparently the documentation is not accurate. 😃

### DIFF
--- a/core/shared/src/main/scala-3/zio/prelude/EqualVersionSpecific.scala
+++ b/core/shared/src/main/scala-3/zio/prelude/EqualVersionSpecific.scala
@@ -17,20 +17,27 @@
 package zio.prelude
 
 import scala.deriving.*
-import scala.compiletime.{erasedValue, summonInline}
+import scala.compiletime.{erasedValue, summonFrom, summonInline}
 
 trait EqualVersionSpecific {
 
-  inline given derived[A](using mirror: Mirror.Of[A]): Equal[A] =
-    val instances = summonAll[mirror.MirroredElemTypes]
+  inline def derived[A](using mirror: Mirror.Of[A]): Equal[A] =
+    val instances = summonAll[A, mirror.MirroredElemTypes]
     inline mirror match
       case sum: Mirror.SumOf[A] => equalSum(sum, instances)
       case product: Mirror.ProductOf[A] => equalProduct(product, instances)
 
-  private inline def summonAll[A <: Tuple]: List[Equal[_]] =
+  private inline def summonAll[A0, A <: Tuple]: List[Equal[_]] =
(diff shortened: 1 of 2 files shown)


regiskuckaertz (approved):
😆

sideeffffect (approved):
Beautiful :heart:

core/shared/src/main/scala-3/zio/prelude/EqualVersionSpecific.scala
```diff
@@ -17,20 +17,27 @@
...
       case _: EmptyTuple => Nil
-      case _: (t *: ts)  => summonInline[Equal[t]] :: summonAll[ts]
+      case _: (t *: ts)  => summon[A0, t] :: summonAll[A0, ts]
+    }
+
+  private inline def summon[A0, A]: Equal[A] =
+    compiletime.summonFrom {
+      case _: (A =:= A0) => compiletime.error("Cannot derive Equal.")
+      case _: (A <:< A0) => Equal.derived(using summonInline[Mirror.Of[A]])
+      case equal: Equal[A] => equal
```
bishabosha, line 40: We should definitely add this pattern to the Scala 3 documentation, was there a specific example this helped with?

adamgfraser: The derivation for `Color`. Would be great if we could push more of this complexity into the derivation functionality in Scala 3 so users did not have to deal with it.

sideeffffect: Unfortunately it doesn't seem to work for recursive types like `List`.

https://scastie.scala-lang.org/y8VjOMW2TtuLsu2Jd9fIPA

On the other hand, @bishabosha 's solution from https://github.com/lampepfl/dotty/pull/16919#issuecomment-1431257814 works for `List`-like types too.

https://scastie.scala-lang.org/rq13aZFWQkq4xZA1sd79MA

adamgfraser: Will take a look.

bishabosha: > The derivation for `Color`. Would be great if we could push more of this complexity into the derivation functionality in Scala 3 so users did not have to deal with it.

Maybe, the idea was that Mirror + derived is intentionally the minimal possible implementation, and things like magnolia, shapeless, etc made a higher level api on top of that

bishabosha: but we can look at defining more helpers for sure

PR #1092, opened 2023-03-02
Fix #1091

### LINKED ISSUE #1091: Difference of Newtype behaviour in pattern matching between RC16 and RC17
Hi,

I tried to update zio-prelude in our project, but some tests fail because the behaviour of newtypes in pattern matching has changed between RC16 and RC17.

This prints `"toto"` with RC16 but prints `"tata"` with RC17:
```scala
import zio.prelude._
import zio.prelude.Assertion._

object Id extends Subtype[Int] {
  override def assertion = assert(greaterThanOrEqualTo(0))
}
type Id = Id.Type

Id(1) match {
  case Id(1) => println("toto")
  case _     => println("tata")
}
```

I'm not sure which behaviour is correct. I'd say that the one from RC16 is the correct one

RC16: https://scastie.scala-lang.org/cvjxJq3JQ6Czy5Vg0jqE5g
RC17: https://scastie.scala-lang.org/OrlZvhDrSDmfLqzgmLgCcQ

guizmaii:
Fix #1091

Related PR:
- https://github.com/zio/zio-prelude/pull/1054

### DIFF
--- a/core/shared/src/main/scala-2/zio/prelude/Newtype.scala
+++ b/core/shared/src/main/scala-2/zio/prelude/Newtype.scala
@@ -284,7 +284,7 @@ abstract class Newtype[A] extends NewtypeVersionSpecific {
    */
   def wrapAll[F[_]](value: F[A]): F[Type] = macro zio.prelude.Macros.wrapAll_impl[F, A, Type]
 
-  implicit def classTag(implicit underlying: ClassTag[A]): ClassTag[Type] = Newtype.classTag
+  // implicit def classTag(implicit underlying: ClassTag[A]): ClassTag[Type] = Newtype.classTag
 }
 
 object Newtype {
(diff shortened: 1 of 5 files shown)


adamgfraser:
@mschuwalow Can you take a look?

core/shared/src/main/scala-2/zio/prelude/Newtype.scala
```diff
@@ -284,7 +284,13 @@ abstract class Newtype[A] extends NewtypeVersionSpecific {
    */
   def wrapAll[F[_]](value: F[A]): F[Type] = macro zio.prelude.Macros.wrapAll_impl[F, A, Type]
 
-  implicit def classTag(implicit underlying: ClassTag[A]): ClassTag[Type] = Newtype.classTag
+  /**
```
sideeffffect: Shouldn't this rather be a regular comment with `/*` instead of `/**`?

guizmaii: Fixed :)

guizmaii:
Closed in favour of https://github.com/zio/zio-prelude/pull/1094

PR #1096, opened 2023-03-03
Apply ZIO SBT Ecosystem and CI Plugins

khajavi:


sideeffffect:
I'm all for improving our CI/sbt story (which I know currently is less than ideal)... But wow, the number of GitHub Actions jobs is _huge_! I'm worried that it wouldn't be reasonable to have such high number, each PR would be _very_ expensive and take a long time to complete.
Perhaps there could be a way to reduce the number. Maybe this can serve as some sort of inspiration https://en.wikipedia.org/wiki/Latin_hypercube_sampling

Or keep the original idea where JVM 17, Scala 2.13 and JVM platform are the default and we diverge from it only in one axis.

khajavi:
@sideeffffect 
With zio-sbt, we can change the granularity of tests. Currently, I need more granularity to find errors. Then I will make tests sequential.

adamgfraser (dismissed):

sideeffffect (commented):
We still want to keep 2.11.

khajavi:
> We still want to keep 2.11.

Ok, I need to override the `sbt-scalajs` plugin. Did you find out how to do that?

adamgfraser:
Can we not show the memory? I think it is extremely disruptive in tests and a symptom of other problems that need to be handled separately.

khajavi:
> Can we not show the memory? I think it is extremely disruptive in tests and a symptom of other problems that need to be handled separately.

Yes, I added it for debugging purposes to find out the proper size for heap and swap size. Now that the tests are green, I can remove them.

adamgfraser:
Okay. The other thing we should be doing here is making sure we are running each test file and each project sequentially. We already run tests in parallel so there is no point in also running different test files or projects in parallel and it can create additional memory pressure that leads to test failures or decreased test performance. In this project I added a couple of command aliases to do that.

khajavi:
> Okay. The other thing we should be doing here is making sure we are running each test file and each project sequentially. We already run tests in parallel so there is no point in also running different test files or projects in parallel and it can create additional memory pressure that leads to test failures or decreased test performance. In this project I added a couple of command aliases to do that.

I added an option to limit the parallel task on the matrix strategy. Also, grouped tests together based on scala and java versions, to make them more sequential.

adamgfraser:
@khajavi So it seems like this has resulted in a dramatic increase in CI times. It is taking around 20 minutes for these tests now versus it previously took around 4 minutes after we merged my PR upgrading the ZIO version. Can we please maintain the current execution semantics?

khajavi:
@adamgfraser Okay, I will try to reduce the duration of the CI process. If this does not lead to any improvement, we need to consider discontinuing automatic test generation.

khajavi:
I'll update this after each try:
- After setting to only test on Java 17=> decreased from 1h 28m to 1h 1m
- After setting max metaspace size to 4G => decreased to 1h and 4m

khajavi:
@adamgfraser To accomplish this task, I utilized the existing semantic, as you said, and attempted to migrate it to the zio-sbt plugin. This involved applying mostly the same workflows used in the original code to the version.

By doing so, I expect to have the same level of quality as the current workflow execution.

guizmaii:
@khajavi, Could you update this PR, maybe, please? I'd be interested in merging these changes 🙂

khajavi:
@guizmaii 
I'm currently engaged in some other things, but I might revisit this later.
Also, I think this PR is open to any collaboration.

PR #1102, opened 2023-03-10
Fix the auxiliary build roots (root211, root212, ...), so that they target JVM only

sideeffffect:


adamgfraser:
Why are we making further changes here? These tests are currently taking five to ten minutes.

sideeffffect:
These auxiliary build roots (root211, root213, ...) were never supposed to build non-JVM targets to begin with. The fact that they do, was an oversight on my part, a bug. The PR also corrects a few other minor CI/build issues, like the JVM versions to run the tests on, a upgrades some dependencies.

adamgfraser (approved):

Issue #380, opened 2020-11-17
Improve documentation in the Invariant hierarchy 


sideeffffect:
_Difficulty: Easy, but with high added value!_

The extension methods in the Invariant hierarchy are missing documentation. Given that most users will interact with it precisely via these extension methods, it would be very unfortunate not to have a documentation on them.

Please add documentation comments to all extension methods in the Invariant hierarchy. Copying the documentation from the method on the Type class itself is OK :wink:

PR #1117, opened 2023-04-23
#380 - documentation comments for some Invariant methods

RayanRal:
Issue #380 - added comment to couple of methods that were missing it. 
The issue is quite old, and most of the class is covered, so feel free to close if that's not needed


sideeffffect (approved):
Nice, thank you.
Would you like to write comments also for the extension methods in the Invariant hierarchy?
These are the methods in the classes `NonEmptyForEachOps`, `ForEachOps`, etc.

adamgfraser (approved):
Thank you! 🙏

PR #1127, opened 2023-05-25
Add Product and Serializable interfaces to These

erdeszt:
It helps with type inference and with avoiding issues where unrelated types are pushed into the same collection(see the `Product` and `Serializable` warts here: https://www.wartremover.org/doc/warts.html)


adamgfraser (approved):

sideeffffect (approved):
Totally, thanks!

PR #1128, opened 2023-05-25
Optimize `forEach_`

0pg:
I ran ForEachBenchmark for a single time. 
Their are significant changes in result for `zioForEach_Option` case using `ListForEach`
(because `ZIO` and `ZPure` both have their own `forEachDiscard` implementation)

before
```
[info] ForEachBenchmarks.zioForEach_Option    100000  thrpt    5   870.152 ±  11.041  ops/s
```

after
```
[info] ForEachBenchmarks.zioForEach_Option    100000  thrpt    5  4893.541 ± 282.137  ops/s
```


adamgfraser:
Due to the confusing implementation of the `tap` operator these implementations won't actually evaluate the effect but just discard it, which is why they are so fast in benchmarks. When you use the correct implementation the difference in performance is dramatically smaller, about 10%. See #1129.

0pg:
Oh I see it and also checked https://github.com/zio/zio-prelude/pull/1129 includes this changes.
I'll close this PR

PR #1129, opened 2023-05-25
Delete Confusing Extension Methods

adamgfraser:
The extension methods defined in the ZIO Prelude package object that are not associated with a functional abstraction are extremely confusing. For example the `tap` operator is available on any data and calls a function, immediately discarding the result. Someone who is working with an effect type that doesn't implement `tap` might call this method, thinking that it is going to describe running their effect and then performing some other effect on the side, discarding its result. But it won't actually evaluate the other effect at all, just construct it and discard it.


sideeffffect:
If it's the name `tap` that's confusing, how about `tee`, `tapAndDiscard` or `teeAndDiscard`? Or something like that?

adamgfraser:
No it is the concept. Adding general "better standard library" extension methods to every data type is not something we should be doing.

sideeffffect (approved):
I see, fair enough.

PR #1134, opened 2023-05-31
Add `forEachFlatten` function

0pg:


guizmaii (approved):

sideeffffect (approved):
Looks fine :+1:

adamgfraser:
This is somewhat anti-modular.

guizmaii:
IMO, this is neat for beginners @adamgfraser 
Even if I'm not really a beginner anymore, I was missing these kinds of methods from Cats, which has many functions like this one: `flatTraverse`, etc.
I have implemented some of them as extension functions in my project.

adamgfraser:
Yeah it is just it conflates effectually iterating over the collection with doing something with the result and there are a very large number of permutations once you start going down that road which is an issue for discoverability if you aren't already coming from another framework that provides these, which are the people who are least in need of help.

guizmaii:
> there are a very large number of permutations once you start going down that road 

Yup, that's true. That's probably why Cats has so many of these functions.
Luka Jacobovitz has a talk about them:
Screenshot 2023-06-02 at 5 52 16 pm
Screenshot 2023-06-02 at 5 52 18 pm
(From https://www.youtube.com/watch?v=0jIaeXMaH2c)

> which is an issue for discoverability

Yes and no, I'd say. Or I might not understand discoverability the way you do.
People will have to reimplement them or will reimplement them without seeing the pattern and so without seeing the possibility of extracting these behaviours into nice, abstract, correct-by-construction, reusable functions like this `forEachFlatten` one.
When I was learning FP and Cats, these functions helped me discover new things.

(I'm not advocating to add all the Cats functions into zio-prelude)

adamgfraser:
I don't think there is any need to reimplement anything.

Let's take a concrete example. I have say a `List[User]` and I have a function `getBooks` with type `User => ZIO[Any, Nothing, List[Book]]`. I do `usersIds.foreach(getBooks)` and I get back a `ZIO[Any, Nothing, List[List[Book]]]`. So I map it and do `.map(_.flatten)`.

That uses `ZIO#map`, which I have to know if I want to do anything if I am writing in that style, and `List#flatten`, which is a standard collection operation. That is the definition of composability and it scales to anything I want to do with that `List[List[_]]`. If I want to do any other transformation `XYZ` on the resulting collection I just do `.map(_.XYZ)` instead of having to remember the name of some `ForEechXYZ` variant that may or may not exist.

sideeffffect:
For the record, I don't insist on keeping this.

Issue #629, opened 2021-05-16
Some additions to ForEach


regiskuckaertz:
How about adding those?
```scala
trait ForEach[F[+_]]:
  def collect[A, B](pf: PartialFunction[A, B]): F[A] => F[B]
  def unzip[A, B]: F[(A, B)] => (F[A], F[B])
```

sideeffffect:
`unzip` could work on a mere `Covariant`, since you just need a `map`, right?
If `collect` is possible, then :+1:

adamgfraser:
We should definitely add `collect`.

The issue with `unzip` I think is as @sideeffffect says above it requires just `Covariant` but if the structure is effectual it involves evaluating the structure twice. This is fine if the structure is a list but perhaps less intuitive if it is a `ZIO`.

PR #1137, opened 2023-06-02
Add `forEachFilter`

0pg:
- `CovariantFilter` typeclass that provides `mapFilter` functionality and has a `Covariant` typeclass
- `ForEachFilter extends CovariantFilter` typeclass that provides `forEachFilter` functionality and has a `ForEach` typeclass

And the `collect` function is following by this work (https://github.com/zio/zio-prelude/issues/629)

### DIFF
--- a/benchmarks/src/main/scala/zio/prelude/ForEachBenchmark.scala
+++ b/benchmarks/src/main/scala/zio/prelude/ForEachBenchmark.scala
@@ -37,12 +37,20 @@ class ForEachBenchmarks {
     bh.consume(list.traverse_(Option(_)))
 
   @Benchmark
+  def catsTraverseFilterOption(bh: Blackhole): Unit =
+    bh.consume(list.traverseFilter(a => Option(a.some)))
+
+  @Benchmark
   def catsTraverseCIO(bh: Blackhole): Unit =
     bh.consume(list.traverse(CIO.pure(_)).unsafeRunSync())
 
   @Benchmark
   def catsTraverse_CIO(bh: Blackhole): Unit =
     bh.consume(list.traverse_(CIO.pure(_)).unsafeRunSync())
+
+  @Benchmark
+  def catsTraverseFilterCIO(bh: Blackhole): Unit =
+    bh.consume(list.traverseFilter(a => CIO.pure(a.some)).unsafeRunSync())
 
   @Benchmark
   def zioForEachOption(bh: Blackhole): Unit =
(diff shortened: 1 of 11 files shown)


adamgfraser:
This doesn't make sense. If I remove your overridden implementation for `List` I get the following:

```scala
object Example extends App {

  val as = List(1, 2, 3, 4, 5)

  val filtered = as.forEachFilter(a => if (a % 2 == 0) Id(Some(a)) else Id(None))

  println(filtered) // List()
}
```

More generally, the contract of `ForEach` is that it allows iterating over a structure and reconstructing a structure of exactly the same "shape". Filtering violates this contract since the structure has a different shape if any elements are filtered out. Furthermore, allowing elements to be filtered out implies that potentially all the elements could be filtered out, which is inconsistent with the contract of `NonEmptyForEach` and every `NonEmptyForEach` must by definition be a `ForEach`.

0pg:
Ah I understood. then `ForEachFilter` should be separated from `ForEach` 🤔

adamgfraser:
Yes it would need to be a new functional abstraction if we want to support it at all.

0pg:
`CovariantFilter` provides combined operator that `map` and `filter` conceptually.
But I think that `CovariantFilter` itself cannot be considered as following `Covariant`'s (neither `Invariant`'s) contract so I made it has a `Covariant` (like `cats` does)
What do you think? @adamgfraser

adamgfraser:
I think at this point we are adding a functional abstraction because we want to implement a particular operator instead of because we have a deep understanding of the underlying domain, which leads to a proliferation of functional abstractions and a lack of orthogonality.

This operator is conflating a bunch of different things between mapping, iterating, and filtering. We can instead use our existing `IdentityEither` functional abstraction, which describes a `none` value for a data type, and conceptualize filtering as replacing each `A` value with either `succeed` if the predicate is satisfied or the `none` value otherwise. Then it is easy to filter the results after applying any transformation.

See #1138.

0pg:
> because we have a deep understanding of the underlying domain, which leads to a proliferation of functional abstractions and a lack of orthogonality.

I was not sure about the filter functionality really needs `Flatten` and `IdentityEither` abstractions, but as far as I understand this sentence says what we actually aims to do
It seems that most of functionality can be given just by composition of already existing functional abstractions

0pg:
with the approach using `IdentityEither`, we can say that we already have `ForEachFilter` functionality but it works quite inefficient way.
I have no idea there is a way to solve this problem without a new functional abstraction

adamgfraser:
@0pg Going to close this and I will work on optimization separately.

PR #1139, opened 2023-06-04
Implement Filter

adamgfraser:


core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
```diff
@@ -240,4 +240,73 @@ trait AssociativeFlattenSyntax {
...
+      covariant: Covariant[F],
+      identityBoth: IdentityBoth[F],
+      identityEither: IdentityEither[F]
+    ): F[A] =
+      fa.flatMap(a => f(a).flatMap(b => if (b) a.succeed else identityEither.none))
+
+    /**
+     * Filters an `F[A]` value with a predicate `f`.
+     */
+    def filterNot[B](f: A => Boolean)(implicit
```
sideeffffect: Is it really worth having?

core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
```diff
@@ -240,4 +240,73 @@ trait AssociativeFlattenSyntax {
...
+      flatten: AssociativeFlatten[F],
+      covariant: Covariant[F],
+      identityBoth: IdentityBoth[F],
+      identityEither: IdentityEither[F]
+    ): F[A] =
+      filter(a => !f(a))
+
+    def withFilter(f: A => Boolean)(implicit
```
sideeffffect, line 295: Why have both `filter` and `withFilter`?

sideeffffect (dismissed):
Looks good, I just 2 questions.

core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
adamgfraser: It is definitely not fundamental. I included it for consistently with the Scala standard library but we could take it out.

core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
adamgfraser: For for comprehensions.

sideeffffect: Aha! I see, neat.
Do you think you could put that to a comment? I think I'm not the only one who wouldn't immediately think of for-comprehensions.

core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
sideeffffect: I don't mind keeping it. But I have to admit I have always wondered about the value of these `*Not` variants even in the standard library.

sideeffffect (dismissed):
Adding a comment with explanation would be great.

core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
```diff
@@ -240,4 +240,73 @@ trait AssociativeFlattenSyntax {
     def flatMap[B](f: A => F[B])(implicit flatten: AssociativeFlatten[F], covariant: Covariant[F]): F[B] =
       flatten.flatten(covariant.map(f)(fa))
   }
+
+  /**
+   * Provides infix syntax for filtering covariant types.
+   */
+  implicit class AssociativeFlattenCovariantIdentityBothIdentityEitherOps[F[+_], A](fa: F[A]) {
```
0pg, line 247: this class is composition of several typeclasses, but why do you add this class in here?
I would like to know if there are any rules for this.

core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
```diff
@@ -240,4 +240,62 @@ trait AssociativeFlattenSyntax {
...
+    ): F[B] =
+      fa.flatMap(a => pf.lift(a).getOrElse(identityEither.none))
+
+    /**
+     * Filters an `F[A]` value with a predicate `f`.
+     */
+    def filter[B](f: A => Boolean)(implicit
+      flatten: AssociativeFlatten[F],
+      covariant: Covariant[F],
+      identityBoth: IdentityBoth[F],
```
sideeffffect, line 276: Do we really need `IdentityBoth` here?

core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
adamgfraser: I took it out.

core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
adamgfraser: Yes we can relax this constraint!

core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
adamgfraser: Done!

core/shared/src/main/scala/zio/prelude/AssociativeFlatten.scala
sideeffffect: I still see it here :smile: 
Btw, can you also double check other methods, that they don't contain unnecessary implicits?

adamgfraser: We need it for `succeed.

sideeffffect (approved):
Thanks, Adam! :+1:

PR #1160, opened 2023-07-09
Add `forEachCollectSome`

0pg:
Hi, Here comes foreach filter again 😅 
Added  infix syntax `forEachCollectSome` of `collect` after `forEach`
(and there is one more for Iterable optimized, this is why the new syntax added instead of using composition of `collect` after `forEach` on demand)

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
+++ b/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
@@ -1127,7 +1127,7 @@ object ZPure {
       val builder  = bf.newBuilder(in)
 
       lazy val recurse: Option[B] => ZPure[W, S, S, R, E, Collection[B]] = { b =>
-        b.fold(builder)(builder += _)
+        b.foreach(builder += _)
         loop()
       }
 
--- a/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
+++ b/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
@@ -1115,6 +1115,30 @@ object ZPure {
     }
 
   /**
+   * Maps each element of a collection to a computation and combines them all
+   * into a single computation that passes the updated state from each
+   * computation to the next and collects only results of type `Some[B]`.
+   */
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -1114,6 +1114,30 @@ object ZPure {
...
+   */
+  def foreachCollectSome[W, S, R, E, A, B, Collection[+Element] <: Iterable[Element]](in: Collection[A])(
+    f: A => ZPure[W, S, S, R, E, Option[B]]
+  )(implicit bf: BuildFrom[Collection[A], B, Collection[B]]): ZPure[W, S, S, R, E, Collection[B]] =
+    ZPure.suspend {
+      val iterator = in.iterator
+      val builder  = bf.newBuilder(in)
+
+      lazy val recurse: Option[B] => ZPure[W, S, S, R, E, Collection[B]] = { b =>
+        b.fold(builder)(builder += _)
```
sideeffffect: Isn't `.forEach` clearer here?

sideeffffect (commented):
Looks good to me.

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
0pg: makes sense. I changed those
[`6809853` (#1160)](https://github.com/zio/zio-prelude/pull/1160/commits/680985372941272030ffeedd23af57302e6f5b9f)

sideeffffect (approved):

adamgfraser:
We already have `collectM`.

0pg:
yeah but can't it only be used with the same effect?
ex) List[A] can only get `PartialFunction[A, List[B]]` not `PartialFunction[A, ZIO[R, E, B]]`

sideeffffect:
Hello @0pg , are you still interested in getting this merged? Could you please resolve the git conflict?

0pg:
I'll close this PR because I think current `collectM` already covers this PR's purpose 😄

sideeffffect:
I see, thanks @0pg  :+1:

PR #1175, opened 2023-08-10
Give some love to the project

guizmaii:
Remove Scala 2.11 support as ZIO doesn't support it anymore


sideeffffect (dismissed):
Looks good, thanks

core/shared/src/main/scala/zio/prelude/Equal.scala
```diff
@@ -862,8 +862,8 @@ object Equal extends EqualVersionSpecific {
    * `Hash` and `Ord` (and thus also `Equal`) instance for `Unit` values.
    * Since there is only one `Unit` value all equality comparisons will always be true.
    */
-  implicit val UnitHashOrd: Hash[Unit] with Ord[Unit] =
-    HashOrd.make(_.hashCode, (_, _) => Ordering.Equals, (_, _) => true)
+  implicit lazy val UnitHashOrd: Hash[Unit] with Ord[Unit] =
+    HashOrd.make(Hash.default.hash, (_, _) => Ordering.Equals, (_, _) => true)
```
guizmaii, line 866: `().hashCode` generates a warning: "dubious usage of method hashCode with unit value"
I'm not sure what it means.
According to https://www.scala-lang.org/api/2.13.11/scala/Unit.html,
> There is only one value of type Unit, (), and it is not represented by any object in the underlying runtime system
Which would mean that calling `.hasCode` on `()` doesn't have much sens as there's not instance of `Unit` at runtime

I'm not sure why this code was compiling and isn't anymore
I'm also not sure about my fix. I know it removes the warning but that's all I know for now.

Not that I made this instance `lazy` because `Hash.default` is itself `lazy`
I'm sure how much this instance is used TBH

core/shared/src/main/scala/zio/prelude/Equal.scala
```diff
@@ -863,7 +863,7 @@ object Equal extends EqualVersionSpecific {
    * Since there is only one `Unit` value all equality comparisons will always be true.
    */
   implicit val UnitHashOrd: Hash[Unit] with Ord[Unit] =
-    HashOrd.make(_.hashCode, (_, _) => Ordering.Equals, (_, _) => true)
+    HashOrd.make(_ => 0, (_, _) => Ordering.Equals, (_, _) => true)
```
guizmaii, line 866: `().hashCode` generates a warning: "dubious usage of method hashCode with unit value"
I'm not sure what it means.
According to https://www.scala-lang.org/api/2.13.11/scala/Unit.html,
> There is only one value of type Unit, (), and it is not represented by any object in the underlying runtime system
Which would mean that calling `.hasCode` on `()` doesn't have much sens as there's not instance of `Unit` at runtime

I'm not sure why this code was compiling and isn't anymore

Why did I decided to replace it with `_ => 0`?
Well, I've found this Akka piece of code: https://github.com/thadeurc/sagat/blob/master/code/akka-1.0-rc3-with-amqp/akka/akka-actor/src/main/scala/akka/util/HashCode.scala

So, in a worksheet, I tried to execute this code:
```scala
import java.lang.reflect.{Array => JArray}
import java.lang.{Float => JFloat, Double => JDouble}

object HashCode {
  val SEED = 23

  def hash(seed: Int, any: Any): Int = any match {
    case value: Boolean => hash(seed, value)
    case value: Char => hash(seed, value)
    case value: Short => hash(seed, value)
    case value: Int => hash(seed, value)
    case value: Long => hash(seed, value)
    case value: Float => hash(seed, value)
    case value: Double => hash(seed, value)
    case value: Byte => hash(seed, value)
    case value: AnyRef =>
      var result = seed
      if (value eq null) result = hash(result, 0)
      else if (!isArray(value)) result = hash(result, value.hashCode())
      else for (id <- 0 until JArray.getLength(value)) result = hash(result, JArray.get(value, id)) // is an array
      result
  }
  def hash(seed: Int, value: Boolean): Int = firstTerm(seed) + (if (value) 1 else 0)
  def hash(seed: Int, value: Char): Int = firstTerm(seed) + value.asInstanceOf[Int]
  def hash(seed: Int, value: Int): Int = firstTerm(seed) + value
  def hash(seed: Int, value: Long): Int = firstTerm(seed)  + (value ^ (value >>> 32) ).asInstanceOf[Int]
  def hash(seed: Int, value: Float): Int = hash(seed, JFloat.floatToIntBits(value))
  def hash(seed: Int, value: Double): Int = hash(seed, JDouble.doubleToLongBits(value))

  private def firstTerm(seed: Int): Int = PRIME * seed
  private def isArray(anyRef: AnyRef): Boolean = anyRef.getClass.isArray
  private val PRIME = 37
}

HashCode.hash(HashCode.SEED, "Hello") // returns `69610501`
HashCode.hash(HashCode.SEED, ())      // returns `851`
().hashCode()                         // returns `0`
```

So I thought `0` would be the correct hashcode value for `Unit`

core/shared/src/main/scala-3/zio/prelude/Liftables.scala
```diff
@@ -37,10 +37,10 @@ trait Liftables {
...
         case '{ Assertion.Regex.literal(${Expr(str)}) }                                                => Some(Assertion.Regex.literal(str))
-        case '{ Assertion.Regex.anyCharOf(${Expr(first)}, ${Expr(second)}, ${Varargs(rest)}: _*) }     => Some(Assertion.Regex.anyCharOf(first, second, Varargs(rest).valueOrError*))
-        case '{ Assertion.Regex.anyRegexOf(${Expr(first)}, ${Expr(second)}, ${Varargs(rest)}: _*) }    => Some(Assertion.Regex.anyRegexOf(first, second, Varargs(rest).valueOrError*))
-        case '{ Assertion.Regex.notAnyCharOf(${Expr(first)}, ${Expr(second)}, ${Varargs(rest)}: _*) }  => Some(Assertion.Regex.notAnyCharOf(first, second, Varargs(rest).valueOrError*))
-        case '{ Assertion.Regex.notAnyRegexOf(${Expr(first)}, ${Expr(second)}, ${Varargs(rest)}: _*) } => Some(Assertion.Regex.notAnyRegexOf(first, second, Varargs(rest).valueOrError*))
+        case '{ Assertion.Regex.anyCharOf(${Expr(first)}, ${Expr(second)}, ${Varargs(rest)}: _*) }     => Some(Assertion.Regex.anyCharOf(first, second, Varargs(rest).valueOrAbort*))
```
guizmaii, line 40: Was generating the following scalac warning:
```scala
[warn] -- Deprecation Warning: /Users/jules/conduktor/workspace/zio-prelude/core/shared/src/main/scala-3/zio/prelude/Liftables.scala:40:152 
[warn] 40 |        case '{ Assertion.Regex.anyCharOf(${Expr(first)}, ${Expr(second)}, ${Varargs(rest)}: _*) }     => Some(Assertion.Regex.anyCharOf(first, second, Varargs(rest).valueOrError*))
[warn]    |                                                                                                                                                        ^
[warn]    |method valueOrError in trait Quotes is deprecated since 3.1.0: Use valueOrAbort
```

See https://github.com/lampepfl/dotty/blob/main/library/src/scala/quoted/Quotes.scala

core/shared/src/main/scala-3/zio/prelude/Liftables.scala
```diff
@@ -130,7 +130,7 @@ trait Liftables {
       case _: Short  => summon[Numeric[Short]].asInstanceOf[Numeric[Any]]
       case _: Byte   => summon[Numeric[Byte]].asInstanceOf[Numeric[Any]]
       case _: Char   => summon[Numeric[Char]].asInstanceOf[Numeric[Any]]
-      case other     => report.throwError(s"NO NUMERIC FOR $other")
+      case other     => report.errorAndAbort(s"NO NUMERIC FOR $other")
```
guizmaii, line 133: Was generating the following scalac warning:
```scala
[warn] -- Deprecation Warning: /Users/jules/conduktor/workspace/zio-prelude/core/shared/src/main/scala-3/zio/prelude/Liftables.scala:133:31 
[warn] 133 |      case other     => report.throwError(s"NO NUMERIC FOR $other")
[warn]     |                        ^^^^^^^^^^^^^^^^^
[warn]     |method throwError in trait reportModule is deprecated since 3.1.0: Use errorAndAbort
```

See https://github.com/lampepfl/dotty/blob/main/library/src/scala/quoted/Quotes.scala

project/BuildHelper.scala
```diff
@@ -166,14 +150,12 @@ object BuildHelper {
 
   def crossPlatformSources(scalaVer: String, platform: String, conf: String, baseDir: File) = {
     val versions = CrossVersion.partialVersion(scalaVer) match {
-      case Some((2, 11)) =>
-        List("2.11+", "2.11-2.12")
       case Some((2, 12)) =>
-        List("2.11+", "2.12+", "2.11-2.12", "2.12-2.13")
+        List("2.12", "2.12+", "2.12-2.13")
```
guizmaii, line 157: `2.12` is still used for https://github.com/zio/zio-prelude/pull/1175/files#diff-be30bee277855f18960ddd438b1594a798cbaf405b6e0ab9d4d7eccf3d965638

This class was in `2.11-2.12`. I transferred it to this `2.12` directory.
It may not be the best decision. 
I decided to do this so I'm sure not to change the behaviour.

guizmaii:
@sideeffffect 

> Looks good, thanks

You might want to have a second look as I made some changes and added some comments to explain some of my changes 🙂

project/BuildHelper.scala
```diff
@@ -201,28 +184,22 @@ object BuildHelper {
...
   def stdSettings(prjName: String) = Seq(
     name                                   := s"$prjName",
-    crossScalaVersions                     := Seq(Scala211, Scala212, Scala213, Scala3),
+    crossScalaVersions                     := Seq(Scala212, Scala213, Scala3),
     ThisBuild / scalaVersion               := Scala213,
     scalacOptions ++= stdOptions ++ extraOptions(scalaVersion.value, optimize = !isSnapshot.value),
     libraryDependencies ++= {
-      if (scalaVersion.value == Scala3)
-        Seq(
-          "com.github.ghik" % s"silencer-lib_$Scala213" % SilencerVersion % Provided
```
guizmaii, line 210: I removed Silencer because, from its doc:
> NOTE: Scala 2.13.2 and 2.12.13 introduced https://github.com/scala/scala/pull/8373. This means that unless you're still cross compiling for Scala 2.11, this plugin is obsolete, and you should use [@nowarn](https://www.scala-lang.org/api/current/scala/annotation/nowarn.html).

See https://github.com/ghik/silencer

core/shared/src/main/scala/zio/prelude/Equal.scala
sideeffffect: We definitely wan to behave like `hashCode`.

sideeffffect:
Still needs formatting

project/BuildHelper.scala
```diff
@@ -201,28 +184,22 @@ object BuildHelper {
...
+          compilerPlugin("org.typelevel" %% "kind-projector" % "0.13.2" cross CrossVersion.full)
         )
+      else Seq.empty
     },
     semanticdbEnabled                      := scalaVersion.value != Scala3, // enable SemanticDB
     semanticdbOptions += "-P:semanticdb:synthetics:on",
     semanticdbVersion                      := scalafixSemanticdb.revision,  // use Scalafix compatible version
     ThisBuild / scalafixScalaBinaryVersion := CrossVersion.binaryScalaVersion(scalaVersion.value),
     ThisBuild / scalafixDependencies ++= List(
-      "com.github.liancheng" %% "organize-imports" % "0.6.0",
```
guizmaii, line 224: `organize-imports` has been incorporated in scalafix 0.11.0

guizmaii:
@sideeffffect 
> We definitely wan to behave like hashCode.

You were too fast 🙂
I was putting the `_.hashCode` back.
Scalafix is slow...

Based on the discussion I had here: https://discord.com/channels/632150470000902164/632150470000902166/1139171420312719400, in the end, I just added a `nowarn` on top of the code

adamgfraser (approved):

PR #1176, opened 2023-08-10
Add 2 new assertions on newtypes: `startsWithIgnoreCase` and `endsWithIgnoreCase`

guizmaii:


macros/shared/src/main/scala/zio/prelude/Assertion.scala
```diff
@@ -163,20 +167,35 @@ object Assertion {
     }
   }
 
-  private[prelude] case class EndsWith(suffix: String) extends Assertion[String] {
+  private[prelude] final case class EndsWith(suffix: String) extends Assertion[String] {
     def apply(a: String, negated: Boolean): Either[AssertionError, Unit] = {
       val result = a.endsWith(suffix)
       if (!negated) {
         if (result) Right(())
-        else Left(AssertionError.Failure(s"startsWith($suffix)"))
```
guizmaii, line 171: The error message was incorrect 🙂

macros/shared/src/main/scala/zio/prelude/Assertion.scala
```diff
@@ -122,7 +126,7 @@ object Assertion {
       } else (!left || !right).apply(a, negated = false)
   }
 
-  private[prelude] case class Or[A](left: Assertion[A], right: Assertion[A]) extends Assertion[A] {
+  private[prelude] final case class Or[A](left: Assertion[A], right: Assertion[A]) extends Assertion[A] {
```
sideeffffect, line 129: There's not much point to declaring `final` AFAIK. Especially when this is private to `prelude`.

sideeffffect (approved):
Looks fine to me. Any comments @adamgfraser ?

macros/shared/src/main/scala/zio/prelude/Assertion.scala
guizmaii: I know that `final` on a `class` enables some JVM/Hotspot optimisations.
No idea about case classes. It's a reflex. 

Also, it prevents people from writing classes that would inherit from these case classes, which is likely invalid code.
See https://nrinaudo.github.io/scala-best-practices/tricky_behaviours/final_case_classes.html
(Very unlikely as it's private code indeed but we never know)

sideeffffect: AFAIK Hotspot is smart enough to speculate that the case classes aren't subclassed and do the optimizations anyway. And given that they indeed aren't in practice, it won't ever have to deoptimize.

But I don't really mind :)

adamgfraser (approved):

PR #1184, opened 2023-08-28
Expose publicly the `Newtype::wrap` && `Newtype::wrapAll` methods

guizmaii:


adamgfraser:
These methods should not be exposed as they allow creating an instance of the new type from any value. The implementor of the new type should be in control of how, if at all, they expose additional constructors for the new type.

PR #1193, opened 2023-09-16
Fix CovariantIdentityBoth#forEach Implementation

### LINKED ISSUE #1194: Broken forEach on CovariantIdentityBoth
Hey all,

using `forEach` on a bunch of data types seems to give the wrong result:

```scala
println(List(List(1, 2), List(3)).forEach(identity))
```

gives `List(List(1, 2, 3, 3), List(1, 2, 3, 3))`, while it should give `List(List(1, 3), List(2, 3))` or something like that.

The problem seems to be in `trait CovariantIdentityBoth` in "coherent.scala":
```scala
def forEach[A, B, Collection[+Element] <: Iterable[Element]](in: Collection[A])(f: A => F[B])(implicit
    bf: BuildFrom[Collection[A], B, Collection[B]]
  ): F[Collection[B]] =
    in.foldLeft(bf.newBuilder(in).succeed)((bs, a) => bs.zipWith(f(a))(_ += _)).map(_.result())
```

Here, a mutable builder is used to collect the results. `_ += _` is passed to `zipWith` which then passes it to `map`. But functions given to `map` on `Covariant` can be called multiple times, as is the case for `List`. Also, both resulting sublists are built from the same shared builder, since `+=` on a builder returns the same mutable builder, as opposed to say `:+` on Vector, which returns a new one.

adamgfraser:


.github/workflows/site.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     if: ${{ github.event_name == 'pull_request' }}
     steps:
     - name: Git Checkout
-      uses: actions/checkout@v4.0.0
+      uses: actions/checkout@v3.3.0
```
sideeffffect, line 21: Is this intentional?

sideeffffect (approved):
Looks good, but the changes to site.yaml seem accidental.

.github/workflows/site.yml
guizmaii: Probably because of zio-sbt which imposes the version of actions/checkout

adamgfraser: Yes.

khajavi: This issue can be fixed when having the latest pre-release version of zio-sbt. I'll make a release as soon as possible.

PR #1211, opened 2023-10-08
Move Filter To ForEach

adamgfraser:


sideeffffect (dismissed):
We seem to have lost `withFilter`. Otherwise the code looks fine. But what was the motivation, if you don't mind me asking?

adamgfraser:
It allows the effect type to be different than the collection type.

sideeffffect (approved):
Cool

core/shared/src/main/scala/zio/prelude/ForEach.scala
```diff
@@ -488,6 +570,8 @@ trait ForEachSyntax {
       F.sum(self)
     def toChunk(implicit F: ForEach[F]): Chunk[A]                                                               =
       F.toChunk(self)
+    def withFilter(f: A => Boolean)(implicit F: ForEach[F], I: IdentityEither[F], B: IdentityBoth[F]): F[A]     =
```
sideeffffect, line 573: Maybe we can add the comment
```
// Provides support for filtering `F[A]` values in for comphrensions.
```

Issue #1183, opened 2023-08-28
[Scala 2.13] Subtype of Newtype cause cyclic subtyping error 


guizmaii:
Same issue as in https://github.com/zio/zio-prelude/issues/817

Reproducer:
```scala
object type {
  type UUIDv1 = UUIDv1.Type
  object UUIDv1 extends Subtype[UUID]
  
  type UserId = UserId.Type
  object UserId extends Subtype[UUIDv1]
}
```

Error:
```scala
[error] example.scala:48:8: cyclic aliasing or subtyping involving type Type
[error]   type UserId = UserId.Type
```

Cc @kitlangton

guizmaii:
See also: https://github.com/zio/zio-prelude/pull/949#issuecomment-1201735998

cheikh2shift:
/attempt #1183

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clnj2g30q0002ml0fa0440m4j/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

guptasamarth61:
/attempt #1183

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clnj2g30q0002ml0fa0440m4j/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

Tan28bha:
/attempt #1183

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clnj2g30q0002ml0fa0440m4j/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

Tan28bha:
[open source.txt](https://github.com/zio/zio-prelude/files/13254211/open.source.txt)
 /claim #1183

abebeos:
@guizmaii , could you please provide a self-contained reproducer that works independently of zio, right from the cli via a usual `scalac` command?

abebeos:
@ team: this joke of a bounty-system does not allow me to cancel my attempt. Please do so from your side)

> The issue is to fix the implementation in ZIO Prelude so that users do not have to use a private compiler flag as a workaround.
from https://github.com/zio/zio-prelude/pull/1228#issuecomment-1824661085

To whomever attempts this:

[...]

PR #1224, opened 2023-11-04
done with the issue

Tan28bha:
/claim #1183


adamgfraser:
@Tan28bha This does not resolve the issue. It merely adds a file with a reproducer that is not even a Scala file in a text but a text file.

PR #1228, opened 2023-11-23
void

ghost:


adamgfraser:
The issue is to fix the implementation in ZIO Prelude so that users do not have to use a private compiler flag as a workaround.

abebeos:
@adamgfraser , what is wrong with you? Even if you do not accept this for the bounty, do your users a favor and mention the workaround until you have a fix.

Additionally, you've surely noticed that fixing one issue introduces regressions, again and again, so mentioning a workaround is simply a must do.

abebeos:
#1183 is a non-issue, see https://github.com/zio/zio-prelude/issues/1183#issuecomment-1824797238

som-snytt:
Probably the question was, what in tarnation is wrong with you. My question is, do bounties not work on closed tickets? I had a question elsewhere whether for planning purposes a ticket can stay closed, even if a PR would be accepted against it, in the unlikely event someone felt it worthwhile. Those unloved tickets are most worthy of a bounty.

abebeos:
> My question is, do bounties not work on closed tickets?

Algoras (quite flawed) bounty process depends on open issues + a mandatory PR.

abebeos:
> I had a question elsewhere whether for planning purposes a ticket can stay closed, even if a PR would be accepted against it, in the unlikely event someone felt it worthwhile. Those unloved tickets are most worthy of a bounty.

@som-snytt Tickets should reflect reality, thus, if an issue exists, it should stay open (or moved to 2nd tracker, e.g. Dedicated feature tracker).

You can always use "tips", e.g. Sending an amount to a dev that did some work. This can be done independent of bounty systems (paypal, skrill, crypto, ...), even without a PR, e.g. for "digging" type of work.

PR #1233, opened 2023-12-09
Resolved "cyclic aliasing or subtyping" 

### LINKED ISSUE #1183: [Scala 2.13] Subtype of Newtype cause cyclic subtyping error

Krishnakalani111:
/claims #1183 
Fixes #1183 

This commit addresses the "cyclic aliasing or subtyping" error that was occurring due to the `UserId` type being a subtype of `UUIDv1`, and `UUIDv1` being a subtype of `UUID`. This created a cyclic relationship that the Scala compiler could not resolve.

To resolve this issue, I've used the `Newtype` and `Subtype` classes from the ZIO Prelude package to create new types and subtypes in a type-safe way. This has allowed us to avoid the cyclic relationship that was causing the error.

### DIFF
--- a/examples/shared/src/main/scala-2.12+/examples/YoDawg.scala
+++ b/examples/shared/src/main/scala-2.12+/examples/YoDawg.scala
@@ -18,6 +18,15 @@ object YoDawg {
 
 }
 
+
+object type {
+ object UUIDv1 extends Newtype[UUID]
+ type UUIDv1 = UUIDv1.Type
+
+ object UserId extends Subtype[UUIDv1]
+ type UserId = UserId.Type
+}
+
 trait GreatTypeclass[-A] {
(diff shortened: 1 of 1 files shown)


adamgfraser:
This doesn't solve the issue. The original reproducer should work as written. A user should be able to define a subtype of a subtype.

PR #1236, opened 2024-01-09
Update Signature of ForEach#collectM

adamgfraser:
I think it is more useful if the effect can be used to determine whether to return a value.


sideeffffect (approved):
If practice shows that this is better, then why not :+1:

PR #1252, opened 2024-02-01
Shortcut `foreach` on empty collections

kyri-petrou:
Micro-optimization to avoid allocations when the collection entered in `foreach` is empty


sideeffffect (approved):
Looks good, thanks

PR #1279, opened 2024-04-13
Modernize/Update/Align CI

guizmaii:


ghostdogpr (approved):

sideeffffect (commented):
Thanks, this looks fine. I think Renovate can update more things than Dependabot, but if that's what other ZIO projects are using, then I don't mind. 

But please keep publishing with Java 8. We have _nothing_ to gain by updating to 11, only shrink the pool of users.

guizmaii:
> But please keep publishing with Java 8. We have nothing to gain by updating to 11, only shrink the pool of users.

The entire ZIO ecosystem has abandoned Java 8 support. Minimal Java version supported by ZIO is now 11.
See for example: https://github.com/zio/zio/pull/8434

PR #1321, opened 2024-05-21
CommutativeIdentity for Unit

SimY4:
Adding missing instance for Unit type.


SimY4:
Actually made this stronger and introduced Idempotency with Inverse into the mix.

sideeffffect (approved):
Great, how could we have missed that! :smile:

PR #1349, opened 2024-08-05
fix grammar: «a […] takes on […]» → «a […] take on […]»

das-g:
«take» is a noun here, not a verb to be conjugated


sideeffffect (approved):
Thank you and congratulations on your fist contribution :clap:

PR #1353, opened 2024-08-12
Review `ZValidation` code: Remove some non-necessary allocations

guizmaii:


core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -75,8 +75,8 @@ sealed trait ZValidation[+W, +E, +A] { self =>
    */
   final def flatMap[W1 >: W, E1 >: E, B](f: A => ZValidation[W1, E1, B]): ZValidation[W1, E1, B] =
     self match {
-      case Failure(w, e) => Failure(w, e)
-      case Success(w, a) =>
+      case failure @ Failure(_, _) => failure
```
kyri-petrou: From a performance perspective, I think this is probably better as it avoids extracting the case class params:

```scala
case failure: Failure[E1, B] => failure
```
Depending on the scalac flags this might raise a warning (I think if the `unchecked` flag is used) , in which case you can do this:

```scala
case failure: Failure[E1 @unchecked, B @unchecked] => failure

// Or alias the annotation to reduce boilerplate:
import scala.{ unchecked => uc }
```

sideeffffect, line 78:
```suggestion
      case failure: Failure => failure
```

sideeffffect: Indeed this.

guizmaii:
@kyri-petrou Thanks for your review. I updated the PR

core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -75,8 +75,8 @@ sealed trait ZValidation[+W, +E, +A] { self =>
    */
   final def flatMap[W1 >: W, E1 >: E, B](f: A => ZValidation[W1, E1, B]): ZValidation[W1, E1, B] =
     self match {
-      case Failure(w, e) => Failure(w, e)
-      case Success(w, a) =>
+      case failure: Failure[_, _] => failure
+      case Success(w, a)          =>
```
kyri-petrou, line 79: Very minor (feel free to ignore) but I generally prefer to put the `Success` cases first (happy path).

kyri-petrou (approved):
Just a super minor comment I missed before. Feel free to ignore, as I think there're probably lots of other places in the codebase to improve this and might be better to do it in a separate PR

core/shared/src/main/scala/zio/prelude/ZValidation.scala
guizmaii: Yeah I hesitated to change but in the end I preferred to minimize the changes

guizmaii:
Thanks @kyri-petrou @sideeffffect 🙏

PR #1356, opened 2024-08-13
`ZValidation#orElse` and `ZValidation#orElseLog` parameter must be lazy

guizmaii:
Do we want to implement tests for this change?

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -212,14 +212,14 @@ sealed trait ZValidation[+W, +E, +A] { self =>
       case Success(w, a) => (w, Right(a))
     }
 
-  final def orElse[W1 >: W, E1, A1 >: A](that: ZValidation[W1, E1, A1]): ZValidation[W1, E1, A1] =
+  final def orElse[W1 >: W, E1, A1 >: A](that: => ZValidation[W1, E1, A1]): ZValidation[W1, E1, A1] =
     self match {
       case Failure(log, _)     => that.mapLogAll(log ++ _)
       case Success(log, value) => Success(log, value)
     }
 
   final def orElseLog[W1 >: W, E1, A1 >: A](
-    that: ZValidation[W1, E1, A1]
+    that: => ZValidation[W1, E1, A1]
   )(implicit ev: E <:< W1): ZValidation[W1, E1, A1] =
     self match {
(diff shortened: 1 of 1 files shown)


kyri-petrou (approved):
This change isn't binary compatible. I guess we're OK with that while we're in RC versions?

ghostdogpr:
> This change isn't binary compatible. I guess we're OK with that while we're in RC versions?

What happened to making a 1.0 by the way 🤣

guizmaii:
@ghostdogpr @kyri-petrou 
I don't know. 
Are we ok? You tell me, mates.
zio-prelude feels quite "low-level" in the sense that lots of libs use it. 
Wouldn't we create a big issue by making a new non-bin-compatible version? 🤔
Or maybe the impact wouldn't be so important as it's only localized in these 2 methods?

(I really don't know much about bin-compat)

sideeffffect (approved):

ghostdogpr:
> @ghostdogpr @kyri-petrou I don't know. Are we ok? You tell me, mates. zio-prelude feels quite "low-level" in the sense that lots of libs use it. Wouldn't we create a big issue by making a new non-bin-compatible version? 🤔 Or maybe the impact wouldn't be so important as it's only localized in these 2 methods?
> 
> (I really don't know much about bin-compat)

As long as the library is in RC mode, there are no guarantees. But I think cutting a 1.0 would be nice to give those guarantees and maybe allow for a more widespread usage in libraries.

guizmaii:
> cutting a 1.0 would be nice

Last time I talked about this with Adam, he was telling me that him and @jdegoes wanted to rework the typeclass hierarchy before to do that. John, any news about that? What was the idea/plan?

ghostdogpr:
He's okay, I asked him back in April.
Screenshot 2024-08-14 at 10 29 01 AM

If there is a rework of typeclasses, it could become a 2.0 later. But with Adam gone, I don't expect it to happen anytime soon?

guizmaii:
@ghostdogpr 
> But with Adam gone, I don't expect it to happen anytime soon?

Agreed

Do we want to have a review of the code before releasing 1.0 tho? 
Because we'll be stuck then.
For example, I want to change `ZValidation::zipPar` and all the variants because, AFAIK, they don't "par" anything. 
They're just `zip` implementations.

guizmaii:
FYI, I started a ticket to list all the things we need to fix/change before `v1.0.0` release: https://github.com/zio/zio-prelude/issues/1359

Please help me listing all these things 🙏

PR #1417, opened 2024-11-14
Simplify `AssociativeFlattenCovariantOps#flatMap` code

guizmaii:


sideeffffect (approved):
Here we're in the definition of an extension method. Here, isn't it better to use bare instances and there methods (instead of other extension methods)?

But if you feel strongly about it, feel free to merge.

guizmaii:
Hey @sideeffffect,

I don't feel strongly about this change
I tend just to trust the Scala compiler to do the right thing (which I might be wrong about 😅)

@ghostdogpr @kyri-petrou Any opinion about this? 🤔

PR #1418, opened 2024-11-14
Make `Optional` an `IterableOnce`

guizmaii:
So that we can:
```
def get(a: A): Optional[B] = ...
val list: List[A] = ...
val result: List[B] = list.flatMap(get)
```
TODO:
- [x] Fix 2.12 compilation

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/data/Optional.scala
+++ b/core/shared/src/main/scala/zio/prelude/data/Optional.scala
@@ -1,7 +1,7 @@
 package zio.prelude.data
 
 import zio.Chunk
-import scala.collection.IterableOnce
+import zio.prelude.IterableOnceCompat
 
 import scala.language.implicitConversions
 
@@ -14,7 +14,7 @@ import scala.language.implicitConversions
  * The only difference between this type and [[scala.Option]] is that there is an implicit
  * conversion defined from `A`` to `Optional[A]`, and from `Option[A]`` to `Optional[A]`.
  */
-sealed trait Optional[+A] extends IterableOnce[A] { self =>
+sealed trait Optional[+A] extends IterableOnceCompat[A] { self =>
   val isEmpty: Boolean
   val isDefined: Boolean
   val nonEmpty: Boolean
--- a/core/shared/src/main/scala/zio/prelude/data/Optional.scala
+++ b/core/shared/src/main/scala/zio/prelude/data/Optional.scala
@@ -148,7 +148,7 @@ sealed trait Optional[+A] extends IterableOnceCompat[A] { self =>
     }
 
   /**
-   * Copied from [[Option.knownSize]]
+   * Copied from `Option.knownSize`
    */
   override final def knownSize: Int = if (isEmpty) 0 else 1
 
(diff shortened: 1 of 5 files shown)


ghostdogpr (dismissed):

core/shared/src/main/scala/zio/prelude/data/Optional.scala
```diff
@@ -13,7 +14,7 @@ import scala.language.implicitConversions
  * The only difference between this type and [[scala.Option]] is that there is an implicit
  * conversion defined from `A`` to `Optional[A]`, and from `Option[A]`` to `Optional[A]`.
  */
-sealed trait Optional[+A] { self =>
+sealed trait Optional[+A] extends IterableOnce[A] { self =>
```
sideeffffect: Why not `Iterable`?

sideeffffect (commented):
I thought Option(al) as Iterable is controversial https://www.wartremover.org/doc/warts.html#option2iterable

But if you really know what you're doing and want it, then I'm not adamantly opposed.

core/shared/src/main/scala/zio/prelude/data/Optional.scala
ghostdogpr: The goal was to make a `Chunk[Optional[A]]` behave the same way as a `Chunk[Option[A]]` where we can just use `flatten` or `flatMap` to get rid of the empty values. And those two methods require an `IterableOnce`.

ghostdogpr:
Seems `IterableOnce` was added in 2.13, in 2.12 the signature of `flatMap` was:
```scala
final override def flatMap[B, That](f: A => GenTraversableOnce[B])(implicit bf: CanBuildFrom[List[A], B, That]): That
```

core/shared/src/main/scala/zio/prelude/data/Optional.scala
sideeffffect: Fair enough, but the question still stands. Wouldn't it be better to extend `Iterable` since we're changing things anyway? Maybe there are other use cases which won't do with mere `IterableOnce`. Or do you think it will be more complicated?

guizmaii: > Why not Iterable?

Because `Option` doesn't extend `Iterable` but `IterableOnce` 🤷‍♂️
I'm just copying `Option` code 🙂

guizmaii:
@ghostdogpr
> Seems IterableOnce was added in 2.13, in 2.12 the signature of flatMap was:

Well, my solution is just to make it compile with Scala 2.12. 
The `.flatMap(optional)` doesn't compile with Scala 2.12 but IMO that's good enough. 
If someone complains, it'll be time to evaluate if the effort to implement it for Scala 2.12 is worth it. 
I don't think anyone will complain

guizmaii:
CI is broken because of https://github.com/actions/setup-java/issues/706

ghostdogpr (approved):
I think this is good enough, immediate benefits and minimal maintenance efforts.

core/shared/src/main/scala/zio/prelude/data/Optional.scala
sideeffffect: Ah, I see. Than that should be good enough for us too :)

sideeffffect:
Nice, thanks guys :clap:

PR #1426, opened 2024-11-26
Give some love to the `Validation` code

guizmaii:
- Create only 1 instance of `fail(())`
- Use `NonEmptyChunk.single` in `ZValidation.fail`

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/ZValidation.scala
+++ b/core/shared/src/main/scala/zio/prelude/ZValidation.scala
@@ -374,6 +374,7 @@ sealed trait ZValidation[+W, +E, +A] { self =>
 }
 
 object ZValidation extends LowPriorityValidationImplicits {
+  private val unitFailure: Validation[Unit, Nothing] = fail(())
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/zio/prelude/ZValidation.scala
```diff
@@ -374,6 +374,7 @@ sealed trait ZValidation[+W, +E, +A] { self =>
 }
 
 object ZValidation extends LowPriorityValidationImplicits {
+  private val unitFailure: Validation[Unit, Nothing] = fail(())
```
sideeffffect, line 377: probably doesn't make sense to make it `lazy`, right?

sideeffffect (approved):
Nice, thank you!

core/shared/src/main/scala/zio/prelude/ZValidation.scala
guizmaii: It doesn't indeed. It's had some synchronisation overhead

PR #1434, opened 2024-12-22
Fix scala steward CI

guizmaii:
Copied from zio-kafka

### DIFF
--- a/.github/workflows/scala-steward.yml
+++ b/.github/workflows/scala-steward.yml
@@ -1,10 +1,14 @@
 name: Scala Steward
 
-# This workflow will launch everyday at 00:00
+# This workflow will launch every day at 00:00
 on:
   schedule:
     - cron: '0 0 * * *'
   workflow_dispatch: {}
+
+permissions:
+  contents: write
+  pull-requests: write
 
 jobs:
   scala-steward:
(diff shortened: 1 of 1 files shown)


.github/workflows/scala-steward.yml
```diff
@@ -1,19 +1,25 @@
...
+  pull-requests: write
+
 jobs:
   scala-steward:
     timeout-minutes: 45
     runs-on: ubuntu-latest
     name: Scala Steward
     steps:
+      - name: Setup sbt
+        uses: sbt/setup-sbt@v1
```
kyri-petrou, line 20: This is new to me 😮

kyri-petrou (approved):

.github/workflows/scala-steward.yml
guizmaii: It's quite new. It's has been implemented as the more recent Ubuntu-based Github runners don't have sbt installed by default anymore

sideeffffect (commented):
This is fine, but more over we need to fix/make work again Mergify.
It used to work well for a long time, but then something changed and the PRs aren't getting automatically merged.

https://github.com/zio/zio-prelude/blob/series/2.x/.mergify.yml

PR #1446, opened 2025-01-08
Fix CI / Update README

sideeffffect:
fixes https://github.com/zio/zio-prelude/actions/runs/12662807767/job/35288328009#step:6:11


guizmaii:
Not a fan of this fix. I think we need to add the `sbt/setup-sbt` action to the CI steps

See https://github.com/zio/zio-prelude/pull/1447

PR #1453, opened 2025-01-15
IdentityFlatten#when and IdentityFlatten#unless.

SimY4:
Small addition to IdentityFlatten syntax allowing conditionally avoiding effect execution.

The signature is consistent with `ZPure#when` and `ZPure#unless`


sideeffffect (approved):
Nice, thanks!

PR #1454, opened 2025-01-15
identity both for vector.

SimY4:
List and Chunk has IdentityBoth instances, but Vector does not for some reason. 

Adding missing IdentityBoth instance to Vector.

FYI, I think there may be more cases were AssociativeBoth instances can be complemented with IdentityBoth: 

One example that comes to mind is ZStream: it only has AssociativeBoth


sideeffffect (commented):
Great, thanks, would you mind adjusting https://github.com/zio/zio-prelude/blob/series/2.x/docs/functional-abstractions/abstraction-diagrams.md#associativeboth as well?

sideeffffect (approved):
Lovely, thank you so much :pray:

PR #1480, opened 2025-02-17
remove variance requirement for Covariant/Contravariant

mberndt123:
Hey there,

I'm proposing to remove some variance annotations on the `Covariant` and `Contravariant` type classes. The downside to having these variance annotations is that you can't provide typeclass instances for types that for whatever reason can't be made covariant. This can be the case for types in a library that you don't control or for Java types.
I recently ran into this problem when I tried to implement a `Covariant` instance for the `TemporalQuery` type in `java.time`. Since I cannot discern any benefit of these annotations, I propose to drop them altogether.
Note that this doesn't preclude having a `F[+_]` type parameter on some _methods_ that really do require this thing to be covariant. But it doesn't need to be on the typeclass itself.

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/AssociativeBoth.scala
+++ b/core/shared/src/main/scala/zio/prelude/AssociativeBoth.scala
@@ -48,11 +48,11 @@ object AssociativeBoth extends AssociativeBothLowPriority {
   def apply[F[_]](implicit associativeBoth: AssociativeBoth[F]): AssociativeBoth[F] =
     associativeBoth
 
-  def compose[F[+_]: AssociativeBoth, G[+_]: AssociativeBoth](implicit
+  def compose[F[_]: AssociativeBoth, G[_]: AssociativeBoth](implicit
     f: Covariant[F],
     g: Covariant[G]
-  ): AssociativeBoth[({ type lambda[+A] = F[G[A]] })#lambda] with Covariant[({ type lambda[+A] = F[G[A]] })#lambda] =
-    new AssociativeBoth[({ type lambda[+A] = F[G[A]] })#lambda] with Covariant[({ type lambda[+A] = F[G[A]] })#lambda] {
+  ): AssociativeBoth[({ type lambda[A] = F[G[A]] })#lambda] with Covariant[({ type lambda[A] = F[G[A]] })#lambda] =
+    new AssociativeBoth[({ type lambda[A] = F[G[A]] })#lambda] with Covariant[({ type lambda[A] = F[G[A]] })#lambda] {
       def map[A, B](fn: A => B)                      = f.map(g.map(fn))
       def both[A, B](fa: => F[G[A]], fb: => F[G[B]]) = fa.zipWith(fb)(_ zip _)
     }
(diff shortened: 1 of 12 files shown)


guizmaii:
Seeing all the additional type annotations that it seems you need to add to make the code compile, I don't think this is a good idea but I don't know enough about this code nor variance to be able to give any better feedback

@jdegoes @kyri-petrou @ghostdogpr @hearnadam @sideeffffect Your opinion?

mberndt123:
Hi @guizmaii,

Thank you for the review, and thank you for bringing this up!
I would like to direct your attention to this commit:
https://github.com/zio/zio-prelude/pull/1480/commits/13ad70203e1249001678140435225ff59eca7732
This is the first commit in this PR, and it is the core of what is being proposed: removing the variance constraints from the `Covariant` and `Contravariant` typeclasses. That commit compiles without errors, and note that it doesn't contain *any* additional type annotations, so additional annotations aren't strictly *needed* for this change to work.

What makes explicit type parameters necessary in some situations is this change to the `succeed` extension method:
https://github.com/zio/zio-prelude/pull/1480/files#diff-7f400b854bb90730036393d7c3fe23654d06af7a2a6a40fcff0f43c884864381L70

The question is whether that particular extension method should be parameterized with `F[+_]` or `F[_]`. If we leave it as `F[+_]`, then type inference works just as it did before. The downside is that it wouldn't work with invariant type parameters. We could also add a separate extension method for invariant types, `.succeedInvariant` or something.

mberndt123:
Oh and some of those explicit type parameters are actually only needed on 2.12. I hope we can phase that zombie out soon.

sideeffffect (commented):
The problem is that I'm not sure I can foresee all the consequences this can have. I just know that the whole ZIO (and thus ZIO Prelude) approach is to lean into variance (co- or contra-) as much as possible. One of the benefits then is a very nice type inference.

It would be nice if @jdegoes could weigh in, if this is a step in a good direction or not.

I'm happy that you're a user and grateful for the effort you put into it. But I'm also a bit worried that this could mess up a lot... Maybe. The problem is that I don't know. At least yet.

In the meantime, could you perhaps force this on your end via the `@uncheckedVariance` annotation?
https://www.scala-lang.org/api/3.x/scala/annotation/unchecked/uncheckedVariance.html

mberndt123:
@sideeffffect,

I wouldn't know how this could possibly break anything important. All it does is allow strictly more types as parameters for `Covariant` and `Contravariant`.

Again, the core of the change is really this commit:
https://github.com/zio/zio-prelude/pull/1480/commits/13ad70203e1249001678140435225ff59eca7732
And the only change required other than the variance change itself is changing this type declaration…
```
type lambda[+A] = F[G[A]]
```
to this:
```
type lambda[A] = F[G[A]]
```
But this is a type declaration inside `Covariant` itself, and it relies on `F[_]` being covariant. But the `F[_]` type variable is scoped inside `Covariant`, so you can't use it outside of that trait and hence it's not a compatibility issue.

sideeffffect:
This is message from @jdegoes 
image

So, for now at least, I'm closing the PR.

@mberndt123 I think it would be useful to hear a feedback from you on how it works on your end when you utilize `@uncheckedVariance` and/or wrapper class(es) as John has suggested.

PR #1579, opened 2025-11-25
Update Scala to `2.13.18` && zio-sbt to `0.4.4`

guizmaii:


sideeffffect:
```scala
[error] /home/runner/work/zio-prelude/zio-prelude/core/shared/src/main/scala-2/zio/prelude/newtypes/package.scala:110:18: method assertion has an inferred structural type: zio.prelude.QuotedAssertion[Int]{def magic: Int}
[error]   members that can be accessed with a reflective call: def magic: Int
[error]     override def assertion = assert(Assertion.greaterThanOrEqualTo(0))
[error]                  ^
[error] /home/runner/work/zio-prelude/zio-prelude/core/shared/src/main/scala/zio/prelude/AssociativeBoth.scala:1532:7: Implicit resolves to enclosing class AssociativeBothCovariantOps; the enrichment wraps value fa
[error]       fa *> forever
[error]       ^
[error] /home/runner/work/zio-prelude/zio-prelude/core/shared/src/main/scala/zio/prelude/AssociativeEither.scala:293:7: Implicit resolves to enclosing class AssociativeEitherCovariantOps; the enrichment wraps value fa
[error]       fa orElse eventually
[error]       ^
[error] /home/runner/work/zio-prelude/zio-prelude/core/shared/src/main/scala/zio/prelude/Hash.scala:803:9: Calls to parameterless method ## will be easy to mistake for calls to def ##(implicit hash: zio.prelude.Hash[A]): Int, which has a single implicit parameter list.
[error]     def ##(implicit hash: Hash[A]): Int =
[error]         ^
```

guizmaii:
@sideeffffect Any idea how to fix these issue?

PR #1580, opened 2025-11-26
Add ZValidation.validateAll_

pancho-bo:
It seems that some instances of ForEach (Map, Iterable,...) overrides `forEach_` method to avoid construction of the result collection. It would be nice to have access to these optimized implementations via `validateAll_` method.


sideeffffect (approved):

sideeffffect:
* https://github.com/zio/zio-prelude/releases/tag/v1.0.0-RC43

sideeffffect:
OK, now for real :D
* https://github.com/zio/zio-prelude/releases/tag/v1.0.0-RC44

PR #1605, opened 2026-01-07
Split `Modify` into `Inspect` / `Update` and change ZPure to an abstract class

kyri-petrou:
After discussion with @ghostdogpr (arguably the person that uses ZPure the most in production world-wide), usage of `modify` directly on a ZPure is highly unlikely. We can avoid tuple allocations etc by separating Modify into `Inspect` and `Update`, while also exposing a method to retrieve the state and map it without having to do `get.map` (see `getWith`).

Other changes (because I was too lazy to create multiple PRs, happy to split it if there are objections):
- Change ZPure into an abstract class to reduce the bytecode in extending classes
- Change all methods on ZPure to be final (some where not)

### DIFF
--- a/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
+++ b/core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
@@ -34,7 +34,7 @@ import scala.util.Try
  * context, state, failure, and logging.
  */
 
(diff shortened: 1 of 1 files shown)


ghostdogpr (approved):

sideeffffect (approved):

sideeffffect:
* https://github.com/zio/zio-prelude/releases/tag/v1.0.0-RC45

PR #1607, opened 2026-01-12
Add FMap as a primitive to ZPure

pancho-bo:
The main idea is to reduce `ZPure.Succeed` allocations.

The stack implementation is a little bit lazy, probably can be done more efficient. For now it just keeps code changes minimal. Using Array of primitive bytes for tags should have little overhead. I have renamed `Stack` to `TagStack` because interface has changed; in particular `TagStack.push` now need a `tag` parameter to work properly. Also added `TagStack.peek` to get the numeric `tag` from the stack, we can't have `TagStack.pop` return a `Tuple`, because that will allocate a `Tuple` and will defeat the purpose of this change. This also requires to call `TagStack.peek` and `TagStack.pop` in one call (although it can still short-cut on `tag` = 0 which means that stack is empty).

It gives 8% reduction in total memory allocations in our `ZPure` heavy computation benchmark. I have also performed a few runs of `zio.prelude.ZPureFullBenchmark`, that benchmark only has one `.map` call so numbers are small. But it can be clearly seen that `ZPure.Succeed` allocation is gone.

I guess benefit dependents on the amount of `.map` and functions that implemented with `.map` (like modify) in the code.

Having tags in the stack probably can be used in some other way than for implementing `.map` to optimize more cases. I think there are should still be some way for improvements.

Original Alloc:
Screenshot 2026-01-12 at 7 21 59 PM
With FMap Alloc:
Screenshot 2026-01-12 at 7 23 41 PM

Original CPU:
Screenshot 2026-01-12 at 7 25 42 PM

With FMap CPU:
Screenshot 2026-01-12 at 7 25 36 PM

Original CPU Stack methods:
Screenshot 2026-01-12 at 7 31 25 PM

With FMap CPU TagStack methods:
Screenshot 2026-01-12 at 7 31 52 PM


sideeffffect:
@ghostdogpr @kyri-petrou if you would kindly make a review of this PR? :pray:

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -774,6 +774,10 @@ object ZPure {
   private val succeedNone: ZPure[Nothing, Any, Nothing, Any, Nothing, Option[Nothing]] = Succeed(None)
   private val succeedUnitFn                                                            = (_: Any) => succeedUnit
 
+  private final val TagFlatMap: Byte = 6
+  private final val TagMap: Byte = 7
+  private final val TagNull: Byte = 0
```
kyri-petrou: AFAICT we're never pushing this tag, so I think it can be removed?

It'll also remove the need for all the `if (nextTag == TagNull)` checks which should simplify the code a fair bit. I understand will require an extra `stack.pop()` at the very end of the ZPure computation, but I think that's probably better than having to push a flatmap tag each time (see my other comment)

core/shared/src/main/scala/zio/prelude/fx/TagStack.scala
```diff
@@ -3,11 +3,12 @@ package zio.prelude.fx
...
  */
-private final class Stack[A <: AnyRef] { self =>
-  import Stack._
+private final class TagStack[A <: AnyRef] { self =>
+  import TagStack._
 
-  private[this] var array  = new Array[AnyRef](ArrSize)
+  private[this] var array  = new Array[AnyRef](ArrSize + 1)
   private[this] var packed = 0
+  array(ArrSize) = new Array[Byte](ArrSize)
```
kyri-petrou: If we only have 2 tags, can we make this into an `Array[Boolean]` instead (and change the tags to be `TagFlatMap = false`, `TagMap = true`)? Since the array will be initialized with `false` as the initial value, we only need to modify it when we push `tag=true` when creating the array

core/shared/src/main/scala/zio/prelude/fx/TagStack.scala
```diff
@@ -58,8 +63,24 @@ private final class Stack[A <: AnyRef] { self =>
       a.asInstanceOf[A]
     }
   }
+
+  def peek: Byte = {
```
kyri-petrou: I think we should name this something a bit more descriptive. If we go with my recommendation of using booleans instead of bytes, it can be renamed to something like `isNextAMap` (I'm terrible at naming, but you get the point)

core/shared/src/main/scala/zio/prelude/fx/Stack.scala
```diff
@@ -21,17 +22,21 @@ private final class Stack[A <: AnyRef] { self =>
   /**
    * Pushes an item onto the stack.
    */
-  def push(a: A): Unit = {
+  def push(tag: Byte, a: A): Unit = {
     val packed0 = packed
     val used    = packed0 & 0xf
     if (used == ArrSize) {
```
kyri-petrou, line 28: Put the array in a local val to avoid fetching it twice from the heap:

```scala
val array = this.array
```

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
```diff
@@ -1216,15 +1224,34 @@ object ZPure {
...
           case succeed0: Succeed[Any] =>
             a = succeed0.value
-            val nextInstr = stack.pop()
-            if (nextInstr eq null) curZPure = null else curZPure = nextInstr(a)
+            var nextTag = stack.peek
+            while (nextTag == TagMap) {
+              val nextInstr = stack.pop()
+              a = nextInstr.asInstanceOf[Any => Any](a)
+              nextTag = stack.peek
+            }
```
kyri-petrou: If we go with booleans, we can simplify this whole thing:

```scala
while (stack.peek)
  a = stack.pop().asInstanceOf[Any => Any](a)
```

core/shared/src/main/scala/zio/prelude/fx/ZPure.scala
kyri-petrou: Also, remove the type definition so that the final val can be inlined

kyri-petrou (commented):
Just some thoughts to potentially simplify the code. I'm aware that using byte tags instead of booleans is more extendable, but I'm wondering whether we're adding too much complexity for something that can be simplified.

Let me know what you think

pancho-bo:
> Just some thoughts to potentially simplify the code. I'm aware that using byte tags instead of booleans is more extendable, but I'm wondering whether we're adding too much complexity for something that can be simplified.
> 
> Let me know what you think

Thank you for the review. I thought that the Byte tags can also be useful to mark other cases, for example Fold; but, probably, there will be not that much to get from having separate case for the Fold anyway. If there is only two case, then it will be better with booleans. Actually, it is then even possible to pack Array[Boolean] into Int, to reduce the overhead of having tags even more.

core/shared/src/main/scala/zio/prelude/fx/TagStack.scala
pancho-bo: Renamed it `tagged`. Also was considering `topTagged`, but I think single word is better.

core/shared/src/main/scala/zio/prelude/fx/Stack.scala
pancho-bo: Saved it to array0, to keep it consistent with other methods

pancho-bo:
New throughput runs: 

With Stack.tags variable field [flamegraph](https://github.com/user-attachments/files/24700342/flame-cpu-forward.html):
```
[info] Benchmark                                   (size)   Mode  Cnt        Score       Error  Units
[info] ZPureFullBenchmark.fallibleBenchmark             1  thrpt   30  5380041.614 ± 29587.637  ops/s
[info] ZPureFullBenchmark.fallibleBenchmark:async       1  thrpt               NaN                ---
[info] ZPureFullBenchmark.fallibleBenchmark          1000  thrpt   30    13286.358 ±   119.093  ops/s
[info] ZPureFullBenchmark.fallibleBenchmark:async    1000  thrpt               NaN                ---
```



Without  Stack.tags variable, just access `array(ArrSize)(0)` every time [flamegraph](https://github.com/user-attachments/files/24700364/flame-cpu-forward.html):
```
[info] ZPureFullBenchmark.fallibleBenchmark             1  thrpt   30  5144364.044 ± 26975.306  ops/s
[info] ZPureFullBenchmark.fallibleBenchmark:async       1  thrpt               NaN                ---
[info] ZPureFullBenchmark.fallibleBenchmark          1000  thrpt   30    11396.424 ±    59.792  ops/s
[info] ZPureFullBenchmark.fallibleBenchmark:async    1000  thrpt               NaN                ---
```

Without FMap [flamegraph](https://github.com/user-attachments/files/24700357/flame-cpu-forward.html):

```
[info] Benchmark                                   (size)   Mode  Cnt        Score       Error  Units
[info] ZPureFullBenchmark.fallibleBenchmark             1  thrpt   30  5430676.780 ± 46725.473  ops/s
[info] ZPureFullBenchmark.fallibleBenchmark:async       1  thrpt               NaN                ---
[info] ZPureFullBenchmark.fallibleBenchmark          1000  thrpt   30    12149.685 ±    98.542  ops/s
[info] ZPureFullBenchmark.fallibleBenchmark:async    1000  thrpt               NaN                ---
```

pancho-bo:
I have renamed `TagStack` back to `Stack`, because GitHub diff can't recognize this file as renamed anymore.

kyri-petrou:
@pancho-bo I experimented a little bit with benchmarks and different approaches and it seems that the most performant approach would be to push `FMap` directly into the stack and then pattern match on it.

Do you mind taking a look at https://github.com/kyri-petrou/zio-prelude/tree/map-primitive-2 and let me know what you think?

pancho-bo:
That would indeed be a much simpler implementation. I think that FMap on the stack is the best version.

kyri-petrou:
> That would indeed be a much simpler implementation. I think that FMap on the stack is the best version.

@pancho-bo Ok nice :) Do you mind merging my branch into your PR? I tried pushing to your branch but I don't have the right permissions for it

pancho-bo:
> Ok nice :) Do you mind merging my branch into your PR? I tried pushing to your branch but I don't have the right permissions for it

Sure. Done.

kyri-petrou (approved):

kyri-petrou:
@sideeffffect happy to merge :)

ghostdogpr (approved):

sideeffffect:
* https://github.com/zio/zio-prelude/releases/tag/v1.0.0-RC46

PR #1626, opened 2026-02-25
CI: Drop JDK 11, add JDK 25

guizmaii:
## Summary
- Remove JDK 11 from the `testJvms` matrix and replace it with JDK 25
- Update all hardcoded JDK 11 references to JDK 17 in the `publishLocal`, `publish`, and `release-docs` jobs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
         uses: actions/setup-java@v5.2.0
         with:
           distribution: temurin
-          java-version: 11 # ensure it works on 11 because we publish for 11
+          java-version: 17
           check-latest: true
       - name: Setup NodeJs
         uses: actions/setup-node@v6
@@ -133,7 +133,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: ['11', '17', '21']
+        java: ['17', '21', '25']
         platform: ['JVM']
     steps:
       - name: Checkout current branch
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -73,7 +73,7 @@ jobs:
         uses: actions/setup-java@v5.2.0
         with:
           distribution: temurin
-          java-version: 11 # ensure it works on 11 because we publish for 11
```
sideeffffect, line 76: Keep 11 here, as the comment says, it's important to know we can publish on 11

.github/workflows/ci.yml
```diff
@@ -294,7 +294,7 @@ jobs:
         uses: actions/setup-java@v5.2.0
         with:
           distribution: temurin
-          java-version: '11'
```
sideeffffect, line 297: Publish should stay on 11

.github/workflows/ci.yml
```diff
@@ -262,7 +262,7 @@ jobs:
         uses: actions/setup-java@v5.2.0
         with:
           distribution: temurin
-          java-version: 11 # publish for 11
```
sideeffffect, line 265: Publish should stay on 11

sideeffffect (commented):
It's fine to up the version where we run the tests etc.
But unless we have a good reason, we can keep publishing on 11, AFAIK there's nothing pushing us to move beyond.

PR #1644, opened 2026-04-24
fix reversed if condition in divisibleBy

ssdSSF:
The if condition in the negated path of `DivisibleBy` needs to be reversed. We should not raise `AssertionError` if `result` is false.

For example, 5 should be valid for`!divisible(2)`, so the `result` is `false`, `!negated` is `false`, then we should go to `Right()` instead of `AssertionError`.


sideeffffect (changes requested):
Thank you so much for your contribution!
Can you add a test to this.

